PR #69, opened 2023-03-01
Added max_tokens property to CreateChatCompletionRequest interface

adamk22:
[According to the documentation](https://platform.openai.com/docs/api-reference/chat/create#chat/create-max_tokens) there should be a `max_tokens` property that works similarly to how it works when using the `createCompletion()` method. However, it seems that the `CreateChatCompletionRequest` interface is missing the `max_tokens` property.


schnerd:
Fixed in v3.2.1 – but thank you for opening!

PR #78, opened 2023-03-02
add: support loading files in nodejs environment

### LINKED ISSUE #77: [Whisper] cannot call `createTranscription` function from Node.js due to File API
### Describe the bug

Cannot call `createTranscription` function like below:

```
...
const audio = await fs.readFile('path/to/audio.mp4');
// Compile Error at the first argument
const response = await openai.createTranscription(audio, 'whisper-1');
```

This is because `createTranscription` interface asks me for File API, which is mainly for Browser API.

```
public createTranscription(file: File, model: string, prompt?: string, responseFormat?: string, temperature?: number, language?: string, options?: AxiosRequestConfig) {
  return OpenAIApiFp(this.configuration).createTranscription(file, model, prompt, responseFormat, temperature, language, options).then((request) => request(this.axios, this.basePath));
}
```

How can I use this function from Node.js? 
Thanks!

---

```
Node.js version: v18.14.2
MacOS Monterey
```

### To Reproduce

```
...
const audio = await fs.readFile('path/to/audio.mp4');
// Compile Error at the first argument
const response = await openai.createTranscription(audio, 'whisper-1');
```

### Code snippets

_No response_

### OS

MacOS

### Node version

Node  v18.14.2

### Library version

openai v3.2.1

chrg1001:
- Closes #77

Add `ReadStream` type from `fs` to every `file` argument.

Since the `file` is just passed to `FormData.append()`, it can also be specified in the form of a `ReadStream`.

https://github.com/openai/openai-node/blob/17fc78122c041a7c1c71158133921208705160be/api.ts#L2475

We can call `createTranscription` from nodejs env like this.

```ts
await openai.createTranscription(fs.createReadStream('/path/to/audio.mp3'), 'whisper-1');
```

### DIFF
--- a/api.ts
+++ b/api.ts
@@ -12,7 +12,7 @@
  * Do not edit the class manually.
  */
 
-
+import type { ReadStream } from "fs";
 import type { Configuration } from './configuration';
 import type { AxiosPromise, AxiosInstance, AxiosRequestConfig } from 'axios';
 import globalAxios from 'axios';
@@ -2117,7 +2117,7 @@ export const OpenAIApiAxiosParamCreator = function (configuration?: Configuratio
          * @param {*} [options] Override http request option.
          * @throws {RequiredError}
          */
-        createFile: async (file: File, purpose: string, options: AxiosRequestConfig = {}): Promise<RequestArgs> => {
+        createFile: async (file: File | ReadStream, purpose: string, options: AxiosRequestConfig = {}): Promise<RequestArgs> => {
             // verify required parameter 'file' is not null or undefined
             assertParamExists('createFile', 'file', file)
             // verify required parameter 'purpose' is not null or undefined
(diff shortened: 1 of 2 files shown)


rmtuckerphx:
Add support for Blob, Stream and base64 string

ericlewis:
not to be a downer, but this would break the ability to use it in cloudflare workers as well as browsers.

ericlewis:
> > not to be a downer, but this would break the ability to use it in cloudflare workers as well as browsers.
> > edit: at least not without complicating those users lives further. of course it can always be stubbed.
> 
> Isn't this package supposed to only be used on the server anyway? It states so at the top of the readme: https://github.com/openai/openai-node#openai-nodejs-library in other words, File should not have been the assigned type in the first place

Yep. Which is why I said Cloudflare Workers first.

ericlewis:
`node_compat = true` will fix it of course.

chrg1001:
@rmtuckerphx

> Add support for Blob, Stream and base64 string

I have tried for types other than `Stream`, but could not use them due to errors. (just converting to each type)

```typescript
const filePath = './audio.m4a';
// ok
const readStream = fs.createReadStream(filePath);
// status 400
const buffer = fs.readFileSync(filePath);
// status 400
const base64 = buffer.toString('base64');
// TypeError: source.on is not a function
const blob = new Blob([toArrayBuffer(buffer)]);

await openai.createTranscription(
  readStream, // buffer, base64 with mime-type, or blob
  'whisper-1',
);
```

rattrayalex:
Thank you so much for taking the time to put this together! 

We have a completely rewritten [upcoming v4](https://github.com/openai/openai-node/discussions/182) of this library, with [conveniences for uploading files](https://github.com/openai/openai-node/tree/v4#file-uploads) in Node, Cloudflare Workers, and other environments. Please give it a try and let us know what you think!

PR #45, opened 2023-01-04
fetch, ESM & streamCompletion util

gfortaine:
- chore(upgrade): upgrade to TypeScript 4.9
- chore(upgrade): enable ESM support
- chore(upgrade): add streamCompletion helper function (credits to @schnerd & @rauschma)
- chore(upgrade): update README for fetch support

PR #80, opened 2023-03-02
feat: add ESM support

ericlewis:
Not to detract from #45 but it is a little crowded with changes and it might be easier to review this way. This PR adds ESM interop for modules while keeping it backwards compatible as well as updates the typescript version.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -14,7 +14,7 @@
       },
       "devDependencies": {
         "@types/node": "^12.11.5",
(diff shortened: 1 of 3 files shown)


rattrayalex:
Thanks for taking the time to put this together, @ericlewis !

We've rewritten the library, and the [upcoming v4](https://github.com/openai/openai-node/discussions/182) has support for ESM. Please give it a try and let us know what you think!

PR #114, opened 2023-03-23
Docs: upgraded `beta.openai.com` -> `platform.openai.com` to prevent unnecessary redirects

Just-Moh-it:
As we all know, the openai platform moved out of beta and changed the URL... This change would prevent redirects!


ezzcodeezzlife (approved):

hebertcisco (approved):

schnerd (approved):
Thanks!

PR #123, opened 2023-03-30
 Adding types for models accepted by OpenAI API endpoints

mbukeRepo:
This pull request adds types for the models accepted by OpenAI API endpoints. By providing type annotations for these models, developers can more easily understand the structure of the data being passed to the API. Additionally, this will enable better error checking and help to prevent unexpected issues that could arise due to not knowing the available models.

The models were obtained from the OpenAI API documentation at the following [URL](https://platform.openai.com/docs/models/model-endpoint-compatibility)

I have created corresponding TypeScript types for accepted models for each endpoint, which have been added to the api.ts file in the root directory and api.d.ts in dist directory. The types have been named according to their corresponding endpoints, and all required fields have been included. 

Thank you for considering this pull request.

### DIFF
--- a/api.ts
+++ b/api.ts
@@ -26,9 +26,9 @@ import { BASE_PATH, COLLECTION_FORMATS, BaseAPI, RequiredError } from './base';
 
 /**
  * @export
- * @type {CreateCompletionRequestModal}
+ * @type {CreateCompletionRequestModel}
  */
-export type CreateCompletionRequestModal =
+export type CreateCompletionRequestModel =
   | "text-davinci-003"
   | "text-davinci-002"
   | "text-curie-001"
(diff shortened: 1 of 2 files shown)


chkpcc1a035:
Is this fixing the  v1/chat/completions path update from v1/completions for #132 ?

mbukeRepo:
> Is this fixing the v1/chat/completions path update from v1/completions for #132 ?

Yes, this fixes any issue that may arise from not knowing models available for openai api endpoints.

Yaronglp:
Nice :)

rattrayalex:
Thank you for adding these types @mbukeRepo ! They look very helpful. 

We have an [upcoming v4](https://github.com/openai/openai-node/discussions/182) of the library which adds better types, including some model enums, eg;

https://github.com/openai/openai-node/blob/v4/src/resources/chat/completions.ts#L181
https://github.com/openai/openai-node/blob/v4/src/resources/completions.ts#L86

We may add more such enums in the future; the place to suggest these changes would be in [our openapi spec](https://github.com/openai/openai-openapi/blob/master/openapi.yaml).

Issue #129, opened 2023-04-02
createImageEdit is not working


ArsHinMQ:
### Describe the bug

So I'm new to openai, I'm using version 3.2.1,  aside the [wrong argument ordering in docs](https://github.com/openai/openai-node/issues/95), the method ```openai.createImageEdit``` is not working
```
  const openaiRes = await openai.createImageEdit(image, prompt, image, n, size)
```
error:
```
TypeError: j.getHeaders is not a function
```
any workaround for this issue? I couldn't find any solution
also I'm using Supabase and Deno, in case it is related

### To Reproduce

1. call createImageEdit
2. raises the error

### Code snippets

```JavaScript
const configuration = new Configuration({
   apiKey: Deno.env.get('OPENAI_API_KEY')
})
const openai = new OpenAIApi(configuration)
const openaiRes = await openai.createImageEdit(image, prompt, image, n, size)
```

### OS

Linux - Debian

### Node version

None - Deno

### Library version

openai 3.2.1

dawadam:
@ArsHinMQ what are properties detail ?
What is "image" ?

dawadam:
When I try to use example documentation code, no error appears:

```
const { Configuration, OpenAIApi } = require("openai");
const configuration = new Configuration({
  apiKey: process.env.OPENAI_API_KEY,
});
const openai = new OpenAIApi(configuration);
const response = await openai.createImageEdit(
  fs.createReadStream("otter.png"),
  fs.createReadStream("mask.png"),
  "A cute baby sea otter wearing a beret",
  2,
  "1024x1024"
);
```

But when I change the import to have "typescript" documentation, like this:

```
import { Configuration, OpenAIApi } from "openai";
const configuration = new Configuration({
  apiKey: process.env.OPENAI_API_KEY,
});
const openai = new OpenAIApi(configuration);
const response = await openai.createImageEdit(
  fs.createReadStream("otter.png"),
  fs.createReadStream("mask.png"),
  "A cute baby sea otter wearing a beret",
  2,
  "1024x1024"
);
```
An error appears on the line `fs.createReadStream("otter.png"),`.

Error is :
`Argument of type 'ReadStream' is not assignable to parameter of type 'File'`

This is because the method whant a "File" object:

`OpenAIApi.createImageEdit(image: File, prompt: string, mask?: File, n?: number, size?: string, responseFormat?: string, user?: string, options?: AxiosRequestConfig<any>): Promise<...>`

This is the difference between api documentation and library.

[...]

Issue #75, opened 2023-03-02
localVarFormParams.getHeaders is not a function


sidmohanty11:
### Describe the bug

When I am using OpenAI node SDK to `createTranscription`,

I am sending it the file like this,
```
  const mediaBlob = await fetch(url)
    .then(response => response.blob());

const audio = new File(
    [mediaBlob],
    'demo.mp4',
    { type: 'video/mp4' },
  );

const resp = await openai.createTranscription(audio, 'whisper-1'); // even audio.text(), stream() etc doesn't work
```

It's throwing me `localVarFormParams.getHeaders is not a function`

### To Reproduce

1. Try to send an audio file for transcription using `openai.createTranscription`
2. Try to send a file
3. Error thrown - `localVarFormParams.getHeaders is not a function`

### OS

macOS

### Node version

Node v16.16.0

### Library version

openai v3.2.1

estepona:
I had the same issue. Same error with cresateTranslation API

AlexNeep:
You need to pass in your API key as part of the configuration object

```javascript
const configuration = new Configuration({
    apiKey: process.env.OPENAI_API_KEY,
  });
  const openai = new OpenAIApi(configuration);
```

el7or:
> Not a solution I recommend, but it does make the createTranscription function accessible. So for a testing/dev environment I don't see any harm 🙂
> 
> ```js
> class CustomFormData extends FormData {
>     getHeaders() {
>         return {}
>     }
> }
> 
> const configuration = new Configuration({
>     apiKey: process.env.OPENAI_API_KEY,
>     formDataCtor: CustomFormData
> })
> ```

This fixed it, Thank you.

TomPeak:
I have the same Problem with createImageVariationes. I do ny app in react. Tried some way, always this error. No Syntax highlighting in WebStrom.

rattrayalex:
Our upcoming [v4](https://github.com/openai/openai-node/discussions/182) fixes this; please give it a try!

PR #151, opened 2023-04-25
Fix: ".getHeaders() does not exist on..." ref: #75 and #129

cjfritz9:
Function OpenAIApiAxiosParamCreator now has a default parameter that instantiates a new Configuration.

I removed the use of creating new instances of FormData() since configuration is now guaranteed. I also removed any "if statements" that were testing if the configuration parameter was supplied.

This ensures that the use of the .getHeaders() method will not return TypeErrors due to FormData() instances not having that method.

Upon testing the scenarios from the bug reports in #75 and #129 I was able to prevent the TypeError. 
Upon testing the scenario of not providing credentials when creating a new Configuration, the proper error was still returned.

### DIFF
--- a/api.ts
+++ b/api.ts
@@ -13,7 +13,7 @@
  */
 
 
-import type { Configuration } from './configuration';
+import { Configuration } from './configuration';
 import type { AxiosPromise, AxiosInstance, AxiosRequestConfig } from 'axios';
 import globalAxios from 'axios';
 // Some imports not used depending on template conditions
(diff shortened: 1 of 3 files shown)


rattrayalex:
Thank you for the thoughtful and thorough fix here @cjfritz9 !

We've fully rewritten this library; the [upcoming v4](https://github.com/openai/openai-node/discussions/182) does not have this issue, so I'm going to close this PR. Please give the v4 a try and let us know what you think!

PR #171, opened 2023-05-21
fix createTranscription bug with types

Lev-Shapiro:
When using method `openai.createTranscription`  with TypeScript it requires to pass at the first parameter something with `type File`, however that type isn't correct ( or at least not for Node.js  ).

If you use `fs.createReadStream` with JavaScript and pass that stream at the first parameter it will work excellent without any errors or bugs, which has the `type ReadStream`. 

This means that .createTranscription method should at least add in TypeScript the type ReadStream, so no one will have to do some ugly code like `// @ts-ignore `or `as any`.  My changes will fix this bug with types.


rattrayalex:
Thanks for taking the time to share a fix!

Our [upcoming version v4.0.0](https://github.com/openai/openai-node/discussions/182) has much improved support for uploading files; please give it a try!

PR #178, opened 2023-06-14
Fix ChatCompletionResponse type bug

HarrisonJackson:
Fixed the types for `ChatCompletionResponseMessage` content field, which is sometimes coming back as null for requests with a function call in the response.

Actual response:
```
{
    "role": "assistant",
    "content": null,
    "function_call": {
        "name": "myFunctionName",
        "arguments": "{\n  \"input\": \"my input data\"\n}"
    }
}
```

### DIFF
--- a/api.ts
+++ b/api.ts
@@ -125,7 +125,7 @@ export interface ChatCompletionResponseMessage {
      * @type {string}
      * @memberof ChatCompletionResponseMessage
      */
-    'content'?: string;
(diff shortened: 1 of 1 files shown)


rattrayalex:
Thanks! This has been fixed in our [upcoming version v4.0.0](https://github.com/openai/openai-node/discussions/182); please give it a try and share your feedback in the thread!

PR #187, opened 2023-06-26
Update README.md

RIAEvangelist:
Updated example with additional details and one syntax change to make it easier for new devs to get started. The example code would not work as it was without these changes.

### DIFF
--- a/README.md
+++ b/README.md
@@ -12,10 +12,10 @@ npm install openai
 
 ## Usage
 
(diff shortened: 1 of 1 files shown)


rattrayalex:
Thank you for this helpful suggestion!

We have a completely rewritten [upcoming v4](https://github.com/openai/openai-node/discussions/182) of this library, which  has a new README showcasing `async function main() {…}` in the top example. We'd rather avoid giving the impression this package works better with or without `type: module`

Issue #167, opened 2023-05-17
ERR_FR_MAX_BODY_LENGTH_EXCEEDED when calling createTranscription


mduraidi:
### Describe the bug

ERR_FR_MAX_BODY_LENGTH_EXCEEDED when calling createTranscription

### To Reproduce

1. use large file (within the limit of 25 MB), in my case the file size is: 19.4 MB

### Code snippets

```JavaScript
async function generateJson() {
const resp = await openai.createTranscription(
  fs.createReadStream("large_file.mp3"), // less than 25 MB
  "whisper-1",
  '',
  'vtt',
  0,
  'fr'
);
```

### OS

Windows 7

### Node version

Node v14.17.6

### Library version

openai v3.1.0

ffrenchm:
The SDK uses Axios under the hood, and Axios has a lower default max body length than the 25MB allowed by Whisper. I solved this error by setting max body length to 25MB in the axios config like so:

const maxBodyLength = 25 * 1024 * 1024

const { data } = await openai().createTranscription(
    // eslint-disable-next-line @typescript-eslint/no-explicit-any
    createReadStream(filePath) as any,
    "whisper-1",
    prompt,
    "verbose_json",
    undefined,
    languageCodeISO2,
    {
      maxBodyLength,
    }
  ) as AxiosResponse<any, unknown>

dupiesdupreez:
@ffrenchm Didn't work for me..

DigitalSolomon:
didn't work for me either @dupiesdupreez ... after wasting too much time, I ended up just installing [whisper.cpp](https://github.com/ggerganov/whisper.cpp) locally and now I'm running speech-to-text quickly and easily off my laptop.

es-lynn:
Have opened a PR to set the default maxBodyLength to 25MB.

In the mean time you can manually increase the maxBodyLength:

```
const configuration = new Configuration({
    apiKey: process.env.OPENAI_API_KEY,
    baseOptions: {
        maxBodyLength: 26214400 // 25 MB
   }
});

const openai = new OpenAIApi(configuration);
```

rattrayalex:
Thank you @es-lynn ! However, we're soon moving away from axios to use `node-fetch` (or native `fetch` in some environments), which should not have this issue:

[v4.0.0 beta](https://github.com/openai/openai-node/discussions/182)

Please give it a try and let us know if you run into any issues, especially with uploading large files for transcriptions!

PR #191, opened 2023-07-05
Fix ERR_FR_MAX_BODY_LENGTH_EXCEEDED issue by setting default maxBodyLength

es-lynn:
This PR addresses the issue [#167](https://github.com/openai/openai-node/issues/167), where a `ERR_FR_MAX_BODY_LENGTH_EXCEEDED` error is thrown when the `createTranscription` function is used with an audio file larger than 10 MB but smaller than 25 MB.

This is because the library uses `axios` which defaults to a 10 MB body limit. 

The issue is resolved by setting a default `maxBodyLength` of 25 MB if it is not already set. This ensures that the limit aligns with OpenAI's maximum file size limit.

### DIFF
--- a/configuration.ts
+++ b/configuration.ts
@@ -102,6 +102,9 @@ export class Configuration {
             'Authorization': `Bearer ${this.apiKey}`,
             ...this.baseOptions.headers,
         }
+        if (this.baseOptions.maxBodyLength == null) {
+            this.baseOptions.maxBodyLength = 25 * 1024 * 1024; // 25MB
+        }
         if (this.organization) {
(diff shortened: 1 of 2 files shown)


rattrayalex:
Thank you so much for taking the time to fix this issue! 

We have a completely rewritten [upcoming v4](https://github.com/openai/openai-node/discussions/182) of this library, which shouldn't run into this issue with large file uploads as it replaces axios with fetch/node-fetch. 

Please give it a try with the files you were having problems with before and let us know if you run into any problems!

PR #197, opened 2023-07-15
Refactor contentLength to use byteLength

albertdbio:
The contentLength was previously calculated from the string length property.

This does not account for multi-byte characters.

Thus we will get the following error when using unicode characters such as, ¿:
```sh
 Request body length does not match content-length header
```


rattrayalex:
Closing this PR as we can't rely on `Buffer` being present in all contexts; I expect we'll have a release out tomorrow.

PR #221, opened 2023-08-17
Fix return type from openaiAPI.audio.transcriptions.create to include string

tadeegan:
Depending on the response_format, the output can either be a json block that is represented by the Transcription interface or a string. For example if you use response_format: vtt

Repro:

```
// transcribe.ts
import { ClientOptions, OpenAI } from 'openai';
const OPEN_AI_CONFIG: ClientOptions = {
  apiKey: process.env.OPENAI_API_KEY,
  organization: process.env.OPENAI_ORGANIZATION_ID,
};

const openaiAPI = new OpenAI(OPEN_AI_CONFIG);

async function main() {
  const params: OpenAI.Audio.Transcriptions.TranscriptionCreateParams = {
      file: fs.createReadStream('/tmp/intput.mp4'),
      model: 'whisper-1',
      response_format: 'vtt',
      language: 'en',
  }
  const response = await openaiAPI.audio.transcriptions.create(
      params,
  );
  console.log(typeof response);
  console.log(response);
}
await main();
```

Outputs:

```
string
00:17:24.220 --> 00:17:30.780
Do you have any questions specifically on anything? I don't think so. The only thing I was gonna let
...
```

### DIFF
--- a/src/resources/audio/transcriptions.ts
+++ b/src/resources/audio/transcriptions.ts
@@ -9,7 +9,7 @@ export class Transcriptions extends APIResource {
   /**
    * Transcribes audio into the input language.
    */
-  create(body: TranscriptionCreateParams, options?: Core.RequestOptions): Core.APIPromise<Transcription> {
+  create(body: TranscriptionCreateParams, options?: Core.RequestOptions): Core.APIPromise<Transcription|string> {
     return this.post('/audio/transcriptions', multipartFormRequestOptions({ body, ...options }));
   }
 }


rattrayalex:
Thanks for reporting this, and for the PR! We'll likely want to fix this in a different way, but I'll leave this PR open for the time being.

tadeegan:
Okay, no prob. I noticed that this is also documented incorrectly on the HTTP api docs.

rattrayalex:
I've opened https://github.com/openai/openai-openapi/pull/84 as draft, which would result in this change being delivered, but we'd rather not ship this until we're able to do it with an overload since it'd regress the default integration path (users whose code was previously working would need to add a typecast). Unfortunately, the way our code generation works doesn't make it trivial to add this overload manually. 

Accordingly, I'm going to close this for now, but I appreciate the contribution and we do hope to get a nice fix out for this in the coming months.

PR #288, opened 2023-09-05
Update README.md to add example code to run

logankilpatrick:
Right now, we don't include info on how to actually run a Node file, adding that here!


logankilpatrick:
I had to do some weird stuff to get this to run, is there a simpler way of doing this @schnerd ?

PR #350, opened 2023-10-05
add OpenAI Node.js Clients comparison

cwtuan:
I want to discuss a new section I've added to the README. It aims to help users choose the right ChatGPT API client package for their Node.js projects.

As a beginner, I understand that it can be hard to select the most suitable npm package for working with ChatGPT API in Node.js. To provide users with a better understanding, I have highlighted the popularity and official status of "openai" as the official Node.js client, so that users can make an informed decision.

If you have any suggestions or feedback regarding this section, please let me know, and I'll be happy to refine it further.

### DIFF
--- a/README.md
+++ b/README.md
@@ -324,3 +324,7 @@ The following runtimes are supported:
 Note that React Native is not supported at this time.
 
 If you are interested in other runtime environments, please open or upvote an issue on GitHub.
+
+## Other Node.js Clients for OpenAI
+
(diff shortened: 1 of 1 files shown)


rattrayalex:
Thank you for taking the time to share your learnings with the community! 

We already mention the Azure OpenAI SDK in the README (though it is much less-used, as you note) and I'm not sure it makes sense to mention the chatgpt package, as it's been largely superseded by this one. Better to make a PR to that package recommending new users come here.

Issue #375, opened 2023-10-13
Inquiry about the 'Accept' header


MartialBE:
### Confirm this is a feature request for the Node library and not the underlying OpenAI API.

- [X] This is a feature request for the Node library

### Describe the feature or improvement you're requesting

Hello,

I am currently using the openai-node package for my project, and I have a question regarding the 'Accept' header in the request made at this specific code snippet: [core.ts](https://github.com/openai/openai-node/blob/648403920c954184585973fcd36e06b3bf36549d/src/core.ts#L191C34-L191C34).

In the mentioned code, the 'Accept' header is set to "application/json". However, when the 'stream' option is set to true, the response I receive is in the format of "text/event-stream".

I am using a unified API platform that converts API interfaces from different platforms to OPENAI format. Due to some platforms validating the 'Accept' header value, I am facing issues with the API requests failing.

Could you please confirm if setting the 'Accept' header to "application/json" in the mentioned code is compliant with the specifications? if it should be modified based on the expected response type?

Thank you for your assistance.

rattrayalex:
Hi, thanks for your request here. You're correct that the server responds with `Content-Type: text/event-stream`, so that'd likely be a more appropriate `Accept:` header to send. We'll look into fixing this; in the meantime, you can send `Accept` manually yourself like so:

```ts
  const stream = await openai.chat.completions.create(
    {
      model: 'gpt-4',
      messages: [{ role: 'user', content: 'Say this is a test' }],
      stream: true,
    },
    { headers: { Accept: 'text/event-stream' } },
  );
```

HAYDEN-OAI:
Thanks for raising this. We reviewed changing the default `Accept` header for streaming requests, but `Accept: application/json` is established SDK behavior and the OpenAI API does not require `Accept: text/event-stream` to enable streaming. Changing the default could break existing gateways, proxies, caches, and compatible servers.

If your gateway requires explicit server-sent-event negotiation, override the header for that request:

```ts
await client.chat.completions.create(
  {
    model: 'gpt-4o-mini',
    messages: [{ role: 'user', content: 'Hello' }],
    stream: true,
  },
  { headers: { Accept: 'text/event-stream' } },
);
```

We're closing this as not planned because preserving the existing default is necessary for backward compatibility.

PR #396, opened 2023-10-23
Add the default accept value for chat completions

c121914yu:

## Changes being requested

Add the default headers **Accept** for chat completions.

## Additional context & links

I found the stream requesting unavailable after upgrading from 3.x to 4.x. I checked both versions of the request header. And I found that the default Accept of version 3.x is "application/json, text/plain, */*", but the default in version 4.x is only "application/json".

c99a93306d6fc403c2796f40940d9b1

When I tried to rewrite the request header in 4.x, I found that the request completed successfully.

image

Of course, I am not requesting api.openapi.com directly, but I think this "Accept " value should be relaxed to make it compatible with more third-party systems.

### DIFF
--- a/src/core.ts
+++ b/src/core.ts
@@ -190,7 +190,7 @@ export abstract class APIClient {
    */
   protected defaultHeaders(opts: FinalRequestOptions): Headers {
     return {
-      Accept: 'application/json',
+      Accept: 'application/json, text/plain, */*',
       'Content-Type': 'application/json',
       'User-Agent': this.getUserAgent(),
       ...getPlatformHeaders(),


rattrayalex:
Thanks for the PR, we'll take a look soon

rattrayalex:
We don't want to change `Accept` globally (as implemented here), but we do want to send the correct `Accept` header for streaming requests. 

You can follow https://github.com/openai/openai-node/issues/375 where we are tracking this. 

In the meantime, as you noted, you can work around this with a second `{ headers: { Accept: 'text/event-stream' } },` argument.

PR #402, opened 2023-10-24
Switch from `node-fetch` to `undici`

### LINKED ISSUE #392: Replace node-fetch with undici

I noticed this library is still using node-fetch because Node's native fetch is considered _experimental_.

I think it'd be in the libraries best interest to switch to [undici](https://undici.nodejs.org/#/) instead. Undici is the fetch implementation in Node.js. For all intents and purposes it is _stable_ (https://github.com/nodejs/undici/issues/1737). We (the maintainers of Undici) have some concerns about marking it as such in Node.js just yet because of the nature of the Fetch api  spec (it itself adds breaking changes occasionally - this doesn't fit well with Node.js versioning strategy. It's complicated - read the issue I linked above for more details). 

Switching the undici for the shim will enable a significantly easier upgrade path in the future whenever we figure out how to mark it as properly _stable_ in Node.js

Happy to help swap this out too if the maintainers approve 😄 🚀 

### Additional context

_No response_

Ethan-Arrowood:
Closes #392

### DIFF
--- a/ecosystem-tests/node-ts-cjs/tests/test-node.ts
+++ b/ecosystem-tests/node-ts-cjs/tests/test-node.ts
@@ -1,11 +1,12 @@
 import 'openai/shims/node';
 import OpenAI, { toFile } from 'openai';
 import { TranscriptionCreateParams } from 'openai/resources/audio/transcriptions';
-import fetch from 'node-fetch';
 import { File as FormDataFile, Blob as FormDataBlob } from 'formdata-node';
 import * as fs from 'fs';
 import { distance } from 'fastest-levenshtein';
 import { ChatCompletion } from 'openai/resources/chat/completions';
+import { Readable } from 'node:stream';
+import { ReadableStream } from 'node:stream/web';
 
 const url = 'https://audio-samples.github.io/samples/mp3/blizzard_biased/sample-1.mp3';
 const filename = 'sample-1.mp3';
@@ -67,10 +68,11 @@ it(`raw response`, async function () {
 
   // test that we can use node-fetch Response API
   const chunks: string[] = [];
-  response.body.on('data', (chunk) => chunks.push(chunk));
+  const body = Readable.fromWeb(response.body as ReadableStream<any>)
+  body.on('data', (chunk) => chunks.push(chunk));
   await new Promise<void>((resolve, reject) => {
-    response.body.once('end', resolve);
-    response.body.once('error', reject);
+    body.once('end', resolve);
+    body.once('error', reject);
   });
   const json: ChatCompletion = JSON.parse(chunks.join(''));
   expect(json.choices[0]?.message.content || '').toBeSimilarTo('This is a test', 10);
--- a/src/_shims/node-runtime.ts
+++ b/src/_shims/node-runtime.ts
@@ -1,21 +1,17 @@
 /**
  * Disclaimer: modules in _shims aren't intended to be imported by SDK users.
  */
-import * as nf from 'node-fetch';
-import * as fd from 'formdata-node';
-import { type File, type FilePropertyBag } from 'formdata-node';
-import KeepAliveAgent from 'agentkeepalive';
-import { AbortController as AbortControllerPolyfill } from 'abort-controller';
-import { ReadStream as FsReadStream } from 'node:fs';
-import { type Agent } from 'node:http';
+import uf from 'undici';
+import type { File, Agent, FormData } from 'undici';
+import type { FilePropertyBag } from 'formdata-node';
 import { FormDataEncoder } from 'form-data-encoder';
+import { ReadStream as FsReadStream } from 'node:fs';
 import { Readable } from 'node:stream';
+import { ReadableStream } from 'node:stream/web';
+import { Blob } from 'node:buffer';
 import { type RequestOptions } from '../core';
 import { MultipartBody } from './MultipartBody';
 import { type Shims } from './registry';
-
-// @ts-ignore (this package does not have proper export maps for this export)
-import { ReadableStream } from 'web-streams-polyfill/dist/ponyfill.es2018.js';
 
 type FileFromPathOptions = Omit<FilePropertyBag, 'lastModified'>;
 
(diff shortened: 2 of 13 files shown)


src/_shims/node-runtime.ts
```diff
@@ -40,11 +38,11 @@ async function fileFromPath(path: string, ...args: any[]): Promise<File> {
   return await _fileFromPath(path, ...args);
 }
 
-const defaultHttpAgent: Agent = new KeepAliveAgent({ keepAlive: true, timeout: 5 * 60 * 1000 });
-const defaultHttpsAgent: Agent = new KeepAliveAgent.HttpsAgent({ keepAlive: true, timeout: 5 * 60 * 1000 });
+const defaultHttpAgent = new uf.Agent({ keepAliveTimeout: 5 * 60 * 1000 });
+const defaultHttpsAgent = new uf.Agent({ keepAliveTimeout: 5 * 60 * 1000 });
```
rattrayalex, line 42: Is this necessary to get reasonable `keepAlive` behavior? IIRC, `node-fetch` didn't do keepAlive by default but if undici does, maybe we don't need to construct a default agent after all?

(note we also have weird timeout shenanigans elsewhere to bump the agent's timeout so that a long timeout for a given request isn't cut short by the agent's timeout; not sure whether that'd still be necessary or would need adjustment.)

rattrayalex (commented):
This is looking clean & great so far! Thank you @Ethan-Arrowood !

src/_shims/node-runtime.ts
Ethan-Arrowood: I'd have to go look at the default agent timeout to be sure

ronag: FYI: undici will honor the keep-alive hint from the server.

rattrayalex:
One thing to note is that with undici, it looks like `response.body` isn't a Node `Readable` stream, which would be breaking change for `node-fetch` users… any thoughts there?

src/uploads.ts
```diff
@@ -125,7 +125,7 @@ export async function toFile(
     }
   }
 
-  return new File(bits, name, options);
+  return new File(bits as (string | Blob | NodeJS.ArrayBufferView)[], name, options);
```
Ethan-Arrowood, line 128: I couldn't for the life of me figure out how to get this type to work. Help appreciated!

Ethan-Arrowood:
> One thing to note is that with undici, it looks like response.body isn't a Node Readable stream, which would be breaking change for node-fetch users… any thoughts there?

Yes this an important distinction. It will be a Node.js Web Stream instead. This is to spec so that code is actually interoperable. `node-fetch` has been a cause of much frustration for a very long time because they don't do this. Developers using fetch - in any context - should expect it to behave as specified

rattrayalex:
> This is to spec so that code is actually interoperable.

You mean, interoperable with web streams, not Node ones, right? So it will be breaking? Is there any way we can ease that migration for developers?

src/uploads.ts
rattrayalex: heh, I believe it, can try to take a look soon… likely thursday…

Ethan-Arrowood:
Current, interoperable with web streams. Also meaning that when you use it in code that runs in both browser and on the server it behaves the same. 

There does exist some utilities for users to transform one to the other. And they are fairly close in implementation so some users won't even notice a difference. The key here is that this breaking change is intentional. Understand if you want that to happen in a major version of this library. Otherwise, if this library tried to do something else you'd just be kicking the can down the road. True interoperability should not have anything other than Web Streams. If true interoperability isn't the absolute goal, we could add a transform to this library and keep Node users on Node streams.

rattrayalex:
I'm probably leaning towards a major version bump in that case. I imagine
shimming a web ReadableStream to also be compatible with a node Readable
stream would be pretty gross. But if there is a library that does that
sanely, it'd make the rollout much easier.

For what it's worth I doubt a huge number of people are likely to be
relying on the raw response so far, so I doubt it'll be a very painful
major version.

rattrayalex:
Hey, sorry, just checking in – are you blocked on me here?

Ethan-Arrowood:
No - haven't gotten back around to this. Need to still figure out how to run tests (still haven't created an OpenAI key). Do you have CI you could enable for this PR? I'll update and mark it "ready for review" so GitHub prompts it too.

Ethan-Arrowood:
Updated failing types I believe. Still cannot run tests. Not sure what's going wrong. Maybe my local env is in a bad state. I think this is very close, maybe a maintainer more familiar with the shim and test matrix can take it over the finish line. 

The most important detail here is that Node.js now has an actual spec compliant Fetch implementation. Its types are different and most importantly it now returns Web Streams instead of Node.js Streams (spec compliant). 

This is more expected IMO as fetch should do as fetch is documented to do. Node.js makes available a `Readable.fromWeb()` that can be used to convert the web streams back to node streams if needed.

rattrayalex:
Amazing, thank you so much @Ethan-Arrowood ! Finishing off those tests ourselves sounds totally fair; we have a lot on our plate right not but will try to take a look at getting this over the line soon.

athyuttamre:
I've approved running the CI workflow. Looks like it is failing with a module error:

```
yarn install v1.22.19
[1/4] Resolving packages...
[2/4] Fetching packages...
error form-data-encoder@4.0.2: The engine "node" is incompatible with this module. Expected version ">= 18". Got "1[6](https://github.com/openai/openai-node/actions/runs/6908472736/job/18819218282?pr=402#step:4:7).20.2"
error Found incompatible module.
info Visit https://yarnpkg.com/en/docs/cli/install for documentation about this command.
Error: Process completed with exit code 1.
```

https://github.com/openai/openai-node/actions/runs/6908472736/job/18819218282?pr=402

src/uploads.ts
jedwards1211: @Ethan-Arrowood I think `undici`'s type for the `fileBits` parameter is wrong, their own JSDoc says it accepts `ArrayBuffer`s, but the type annotation only accepts `ArrayBufferView`s, which `ArrayBuffer`s aren't assignable to.

athyuttamre:
@Ethan-Arrowood can you rebase on master so that the latest CI workflow can be run with Node 18?

athyuttamre:
Looks like CI now runs correctly! Some type errors to fix next it looks like.

Ethan-Arrowood:
Why are they all coming from the `dist` folder?

rattrayalex:
Thanks Atty! 

I think @jedwards1211 is working on these type errors etc – Ethan, you shouldn't need to take a look.

rattrayalex:
Sorry for the annoyance. 

We're hoping to merge this within the next 6-8 weeks. 

It will be a breaking change and there are some deep complexities, which is why it'll take longer than one might hope, but we are actively working on it behind the scenes.

ecosystem-tests/node-ts-cjs/tests/test-node.ts
```diff
@@ -67,10 +68,11 @@ it(`raw response`, async function () {
 
   // test that we can use node-fetch Response API
```
wojtekmaj, line 69: This comment probably needs updating

rattrayalex:
(We are still working on this! Sorry for the delays, everybody)

RobertCraigie:
Update here is that we're planning on moving entirely to builtin APIs so we won't even need to depend on `undici`! Sorry for the delays, this will still be a while.

edoerpani (approved):

tshak:
@kwhinnery-openai is there a tracking issue for this fix? Thanks!

nwalters512:
For folks following along, it looks like https://github.com/openai/openai-node/pull/1237 replaced `node-fetch` with built-in fetch, and this is likely going to be released in v5.0.0 (https://community.openai.com/t/your-feedback-requested-node-js-sdk-5-0-0-alpha/1063774).

PR #451, opened 2023-11-07
fix type for image_url

ryoppippi:

## Changes being requested
According to the official document, we can specify string url directory to the image_url for gpt-4-vision-preview.
So I made a PR to fix the type error

スクリーンショット 2023-11-07 11 35 08

## Additional context & links
https://platform.openai.com/docs/guides/vision


ryoppippi:
Because `ImageURL` is defined by the interface, we cannot use union type.

ryoppippi:
Oh, this is autogenerated.
So, I'll create a new PR on openAPI

Issue #422, opened 2023-11-04
TS7030: Not all code paths return a value


AmitMY:
### Confirm this is a Node library issue and not an underlying OpenAI API issue

- [X] This is an issue with the Node library

### Describe the bug

When building using `tsc` with `openai` as a dependency, I get the following error:
node_modules/openai/src/lib/AbstractChatCompletionRunner.ts(224,28): error TS7030: Not all code paths return a value.

I also had to disable the `noImplicitAny` rule since `openai` is not built, and then my app build inherits your source code basically.

### To Reproduce

`const OpenAI = require('openai');`

and build with `tsc` 

(Ideally, I would like this to be an automated test for this library, so build errors do not repeat.)

### OS

macOS, ubuntu

### Node version

Node 18

### Library version

4.15.1

rattrayalex:
Thanks for reporting. Can you share your tsconfig?

AmitMY:
As mentioned, I am importing `openai` like so: `const OpenAI = require('openai');` in runtime.
I am also importing types in advance - `import type OpenAI from 'openai';`

My `tsconfig.json` is:
```ts
{
  "compilerOptions": {
    "module": "commonjs",
    "noImplicitReturns": true,
    "noUnusedLocals": true,
    "skipLibCheck": true,
    "outDir": "lib",
    "sourceMap": true,
    "strict": true,
    "target": "es2017",
    "moduleResolution": "node",
    "types": ["node", "jest"],
    "resolveJsonModule": true,
    "useUnknownInCatchVariables": false,
    "noImplicitAny": false
  },
  "compileOnSave": true,
  "include": ["src"]
}
```

rattrayalex:
Thanks. What version of TypeScript are you using?

rattrayalex:
I can't reproduce this error on TypeScript 5.2, 4.9, or 4.5. Could you please provide a stackblitz or similar repo that reproduces the problem?

AmitMY:
my code base for exact repro:
```bash
git clone https://github.com/sign/translate.git
cd translate/functions
npm install
tsc
```

<img width="949" alt="image" src="https://github.com/openai/openai-node/assets/5757359/ae4657f1-25b9-4274-99db-9bc3c5b29dd2">

gommzystudio:
Hey,
I temporary fixed the issue by adding

```
"noImplicitReturns": false,
"strict": false,
```

to my tsconfig.

RobertCraigie:
[...]

PR #473, opened 2023-11-09
fix: error TS7006: Parameter 'ctrl' implicitly has an 'any' type.

AmitMY:

## Changes being requested
fix error TS7006: Parameter 'ctrl' implicitly has an 'any' type.
by applying the correct type

## Additional context & links
https://github.com/openai/openai-node/issues/422

### DIFF
--- a/src/streaming.ts
+++ b/src/streaming.ts
@@ -184,7 +184,7 @@ export class Stream<Item> implements AsyncIterable<Item> {
       async start() {
         iter = self[Symbol.asyncIterator]();
       },
-      async pull(ctrl) {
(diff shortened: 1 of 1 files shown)


rattrayalex:
As mentioned in the linked thread:

The issue is that you're importing from openai/src, the src/ path is only included to help IDEs with "go to source" calls etc.

If you change this line:

```ts
import type {ChatCompletionMessageParam} from 'openai/src/resources/chat/completions';
```
to this:
```ts
import type {ChatCompletionMessageParam} from 'openai/resources/chat/completions';
```
Then everything should work.

PR #493, opened 2023-11-13
Added missing name paramter to chat completion message params.

anzemur:

## Changes being requested
According to [the documentation](https://platform.openai.com/docs/api-reference/chat/create#chat-create-messages) we should be able to specify the name for each of the base message types/roles (system, assistant, and user) to provide the model with additional information to differentiate between participants of the same role (useful for few-shot prompting).


anzemur:
Opened an issue in the [OpenAPI specification repository](https://github.com/openai/openai-openapi/issues/118).

Issue #418, opened 2023-11-03
Support for web ReadableStream without buffering the whole file.


nicolopadovan:
### Confirm this is a Node library issue and not an underlying OpenAI API issue

- [X] This is an issue with the Node library

### Describe the Bug

Whenever a web `ReadableStream` is passed to the `toFile` helper function, the full contents of the file are buffered before forwarding the request to the Whisper OpenAI API endpoint.
However, it is possible to avoid having to buffer the whole file into the server memory, and instead just use the server as a middleware that streamlines the data from the source of the file to the API endpoint.
The problem has been discussed on issue #414 as well.

A workaround using `axios` and `FormData` that seems to work:

```javascript
import FormData = require("form-data");
import axios from "axios";

async function foo() {

  const bucket = storage.bucket(bucketName);
  const file = bucket.file(fileName);
  const readStream = file.createReadStream();
  
  const form = new FormData();
  
  // Make sure that the file has the proper extension as well
  // (In this example, we just add the `webm` extension for brevity
  if (!fileName.includes(".")) {
      fileName = `${fileName}.webm`;
  }
  
  form.append("file", readStream, fileName);
  form.append("model", "whisper-1");
  
  const apiKey = OPENAI_API_KEY
  
  const response = await axios.post(
      "https://api.openai.com/v1/audio/transcriptions",
      form,
      {
          headers: {
              ...form.getHeaders(),
              Authorization: `Bearer ${apiKey}`,
          },
      }
  );
  
  const transcription = response.data.text:
  return transcription;
}
```

### To Reproduce

Example uses Cloud / Firebase Storage.

### Code snippets

```JavaScript
import {OpenAI, toFile} from 'openai'

const bucket = storage.bucket(bucketName);
const file = bucket.file(fileName);

const openai = new OpenAI({
    apiKey: apiKey,
});

const completion = await openai.audio.transcriptions.create({
    file: await toFile(file, "myfile.mp3"),
    model: "whisper-1",
});
```

### OS

Linux (Google Cloud Functions)

### Node version

18

### Library version

4.11.1

rattrayalex:
Thanks for filing, I would like to do this.

Note that this applies to all forms of streams, not just web ReadableStreams (and the GCP libraries return NodeJS ReadableStreams, not web ones).

Unfortunately, I'm not sure this will really be possible in a clean way until OpenAI's backend can infer content-types from the contents instead of from the filenames, since you can't construct a `File` class with a stream, and you need the filename.

[...]

PR #524, opened 2023-11-21
Allow readable streams to be uploaded

tarcisiozf:

## Changes being requested
Right now there is no way to upload files using a read stream unless it's coming from a `fs.ReadStream`. This forces the user to go through disk first before sending a stream or to reallocate the whole content to send as Buffer. 

In my use case I'm handling a audio stream via web API, transforming the webm via stream and, unfortunately, persisting the file before sending.

The `fs.ReadStream` already extends a stream.Readable so we can do a small extension to the code without needing to change behavior.

For this PR I decided to add support for _Readable_ instead of replacing the existing _FsReadStream_ checks. I also know that the code is unlikely to be merged but it demonstrates a simple solution to it.

### DIFF
--- a/README.md
+++ b/README.md
@@ -224,6 +224,7 @@ Request parameters that correspond to file uploads can be passed in many differe
 - `File` (or an object with the same structure)
 - a `fetch` `Response` (or an object with the same structure)
 - an `fs.ReadStream`
+- an `stream.Readable`
 - the return value of our `toFile` helper
 
 ```ts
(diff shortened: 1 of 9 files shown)


rattrayalex:
I don't believe this would work well with the current OpenAI API, which inspects filenames for media types (eg, you need a `.mp3` extension for an audio file to be interpreted as mp3). That's why we only allow `fs.createReadStream()` to bypass `toFile()` – it includes a filename.

This may work with node Readable, but would it work with web ReadableStream?

Does this actually make the request as expected with a large file, streaming the request body? I think further changes would be required for that, and we might be best off waiting for https://github.com/openai/openai-node/pull/402 to land for that.

tarcisiozf:
Thanks for the reply, now the API constraints are clear to me. For now I'll use `await toFile(buffer, 'audio.webm')` instead of a stream solution.

rattrayalex:
Great. You may want to follow https://github.com/openai/openai-node/issues/418 which documents a feature request for this.

Issue #507, opened 2023-11-16
Examples for streaming tools calls need fixing


stippi2:
### Confirm this is a Node library issue and not an underlying OpenAI API issue

- [X] This is an issue with the Node library

### Describe the bug

I have used the `messageReducer` from the examples. After streaming the chunks finished, and I had the assembled message, I saw that the `arguments` property was always empty. I am testing the new GPT-4 model which supports multiple function calls. After hours of debugging, I finally realized that the problem was the reducer function, which didn't support arrays. This is how my reducer function currently looks, but I am not sure if that is correct. It would never add additional items from the delta:

```js
function messageReducer(previous: ChatCompletionMessage, item: ChatCompletionChunk): ChatCompletionMessage {
  const reduce = (acc: any, delta: any) => {
    acc = { ...acc };
    for (const [key, value] of Object.entries(delta)) {
      if (acc[key] === undefined || acc[key] === null) {
        acc[key] = value;
      } else if (typeof acc[key] === 'string' && typeof value === 'string') {
        (acc[key] as string) += value;
      } else if (typeof acc[key] === 'number' && typeof value === 'number') {
        (acc[key] as number) = value;
      } else if (Array.isArray(acc[key]) && Array.isArray(value)) {
        const accArray = acc[key] as any[];
        if (accArray.length !== value.length) {
          throw new Error(`Array length mismatch for key ${key}: ${accArray.length} !== ${value.length}`);
        }
        for (let i = 0; i < value.length; i++) {
          accArray[i] = reduce(accArray[i], value[i]);
        }
      } else if (typeof acc[key] === 'object' && typeof value === 'object') {
        acc[key] = reduce(acc[key], value);
      }
    }
    return acc;
  };
  
  return reduce(previous, item.choices[0]!.delta) as ChatCompletionMessage;
}
```

### To Reproduce

Use the new `gpt-4-1106-preview` model which returns a `tool_calls` array in combination with streaming, while accumulating the message using the reducer function from the examples.

### OS

macOS

### Node version

Node v20.7.0

### Library version

openai 4.17.4

stippi2:
This is just to confirm that my version of the reducer is not correct. I just saw an instance where strings were concatenated that should not have been...

rattrayalex:
We should probably just delete this example; any reason not to use `.stream()` or `.runTools` instead?

stippi2:
[...]

PR #535, opened 2023-11-27
update an example to use tools instead of a function

nwatab:

## Changes being requested
Update an example of function calling while streaming to use .tool_calls. Though openai-node library now lets a user to call a function automatically, some people may want to control at lower level.

## Additional context & links
https://github.com/openai/openai-node/issues/507#issuecomment-1825010015

### DIFF
--- a/examples/tool-calls-stream.ts
+++ b/examples/tool-calls-stream.ts
@@ -11,43 +11,53 @@ import {
 // gets API Key from environment variable OPENAI_API_KEY
 const openai = new OpenAI();
 
-const functions: OpenAI.Chat.ChatCompletionCreateParams.Function[] = [
+const tools: OpenAI.Chat.Completions.ChatCompletionTool[] = [
   {
-    name: 'list',
-    description: 'list queries books by genre, and returns a list of names of books',
-    parameters: {
-      type: 'object',
-      properties: {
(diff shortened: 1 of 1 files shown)


nwatab:
I found `examples/function-call-stream-raw.ts` is the same as `examples/function-call-stream.ts`. I was not sure about the file, and so I didn't touch it.

rattrayalex:
Thanks very much @nwatab! Can you rename the file as well?

We should remove a bunch of the old function calling examples, including the redundant `-raw.ts` one. You're welcome to do that here, or we can later.

OOC, any reason not to use `.runTools()`?

nwatab:
Thanks for checking the PR. I renamed the file name and overwrote a commit history. I'll keep on doing when I have time.

The reason is simple. I had been writing the source code for the function call even before the release of the `.runTools` feature. I thought it would be faster to deliver to change my code to use `.tool_calls`, rather than upgrading library versions and use new feature.
Though I want to control at lower-level at some extent. I will see the specifications and behavior of `.runTools` and decide whether to use it moving forward.

rattrayalex:
Amazing, thank you @nwatab ! I hope to review and merge tomorrow. 

I'll look forward to your feedback on `.runTools` as well!

BertsLifee (approved):

rattrayalex:
This change has been incorporated on our end and will be released soon! Thank you very much.

PR #539, opened 2023-11-27
[github] Update Node version to 18 in workflows

athyuttamre:
Bumps Node to v18 (LTS) in our CI workflows. I think this will unblock running CI on https://github.com/openai/openai-node/pull/402 which requires Node 18.


logankilpatrick (approved):

rattrayalex (approved):
Thank you Atty!

PR #542, opened 2023-11-28
Add Emojis To README

justindhillon:

## Changes being requested

I want to add emojis to [README.md](https://github.com/justindhillon/openai-node/blob/master/README.md). You can press on that like to see how much better it looks. The emojis also distinguish ## from ###. So people can easily see what is a header and what is a sub-header. These emojis are chosen to represent the context of each title, making your README more visually appealing and easier to navigate.


athyuttamre:
Thanks for the contribution! I can see how emoji makes the headers more distinguished. Ultimately this is a stylistic change, and not one we use plan to use across our repos, so closing.

PR #587, opened 2023-12-20
added cleanObject to defaultHeaders

haouarihk:

## Changes being requested
adding cleanObject to defaultHeaders,
this can help if people wants to remove header properties, from the configuration, by setting them to be empty string, or undefined. instead of it sending that useless value.

## Additional context & links
in my case, where i need it, is that i'm using a middle tool called langchainjs, which allows for setting defaultHeader as a property, but i can't reach the class to override that getDefaultHeader function.
and i want to delete headers like
X-stainless-* because they get blocked in some openai capable apis

Thanks for all the hardwork you put into this awsome tool

### DIFF
--- a/src/core.ts
+++ b/src/core.ts
@@ -1019,6 +1019,31 @@ export function debug(action: string, ...args: any[]) {
   }
 }
 
+/** This function removes any properties that are undefined or empty string,
+ * This can be used with headers, as to not send empty header properties if there are any, 
+ * provided by the user.
+ */
+export function cleanObject<T>(options: T) {
+  const cleanedOptions: Partial<T> = {};
+  type keyType = keyof typeof options;
+
+  for (const key in options) {
+    if (Object.prototype.hasOwnProperty.call(options, key)) {
+      const value = options[key as keyType];
+
+      // Check if the value is not an empty string or undefined
+      if (
+        (typeof value != "undefined" && typeof value !== "string") ||
+        value !== ""
+      ) {
+        cleanedOptions[key as keyType] = value; // Copy non-empty properties to the cleaned object
+      }
+    }
+  }
+
+  return cleanedOptions;
+}
+
 /**
  * https://stackoverflow.com/a/2117523
  */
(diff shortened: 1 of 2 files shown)


rattrayalex:
I'm curious for more context here. 

1. Which headers specifically are you trying to block? 
2. Where are they coming from?
3. Why do you need to block them? What problems do they cause specifically?
4. Are they being set to empty string or undefined?

haouarihk:
> I'm curious for more context here.
> 
> 1. Which headers specifically are you trying to block?
> 2. Where are they coming from?
> 3. Why do you need to block them? What problems do they cause specifically?
> 4. Are they being set to empty string or undefined?

Trying to block the platform headers like.
X-stainless-os and the others.
For the reason of some providers whitelist certain headers but not others like x-stainless. But other than that, they have the same api structure.

One example of that is mistral.ai

rattrayalex:
Ah, I see – if you do this:

```ts
const client = new OpenAI({
  defaultHeaders: {
    'X-Stainless-OS': undefined
  }
})
```
we should be not sending that header at all, but today in at least some environments we'll send the header `X-Stainless-OS: undefined` which is not intended behavior.

I'm not sure this is the exact fix we'd want – I'd expect `''` to still send the header, for example, and we'd probably want to use our `hasOwn` helper – but I do expect we'll resolve this shortly. 

We also need to address case-sensitivity for our headers, which we may do as part of this change.

Regardless, thank you for raising the issue and for the contribution!

rattrayalex:
Note you can work around this for now by setting the default headers to `null` instead of undefined, which I believe we do filter out today.

I've implemented this fix internally with a different approach, so I'll close this PR, but thank you again!

haouarihk:
an easy way to solve that case sensitivity is to turn all header keys to lowercase on the merging step, and in the same step save the casing in a hashmap. 
after its done, loop over the hashmap to get the casing back on the headers.

rattrayalex:
Yes, that's the approach we're taking.

Issue #606, opened 2024-01-03
Enhancement Request: Add Documentation for Embedding Creation Method in OpenAI Node Package After API Update from v3 to v4


singh1203:
### Confirm this is a feature request for the Node library and not the underlying OpenAI API.

- [X] This is a feature request for the Node library

### Describe the feature or improvement you're requesting

The issue involves missing documentation for the embedding creation method in the OpenAI Node Package after the API update from v3 to v4. While the api.md file currently provides guidance on how to use the method and details about the response type, it is suggested to include this information in the README file as well.

- Visit the OpenAI Node Package repository on GitHub.
- Navigate to the README file or the documentation section.
- Observe that the information related to the embedding creation method after the API update from v3 to v4 is not present in the main documentation. This absence affects the accessibility for users seeking details on the updated functionality.

This correction will ensure better visibility on the npm main page or website, making it more accessible for users seeking information on embedding creation in the updated version

### Additional context

An effective solution involves updating the README with a usage example. This addresses the missing documentation for the embedding creation method post the API update from v3 to v4, ensuring enhanced clarity and accessibility on the npm main page or website

rattrayalex:
As described on the linked PR, this is available in the API reference docs: https://platform.openai.com/docs/api-reference/embeddings/create

PR #607, opened 2024-01-03
Improvement: Added Embedding Section to README.md

singh1203:
Fixes issue #606 

## Changes being requested
Updated `README`

## Additional context & links
More context is described in issue #606


rattrayalex:
Thanks for the contribution! Examples like this are available in the OpenAI API reference documentation, e.g., here: https://platform.openai.com/docs/api-reference/embeddings/create

If we were to add content like this to the README for every method, the README would grow too long, unfortunately. We're exploring other alternatives for this. 

Accordingly, I'll go ahead and close this. Thanks again!

PR #614, opened 2024-01-08
Remove web-streams-polyfill and abort controller polyfill

### LINKED ISSUE #613: Remove unnecessary web-streams-polyfill polluting global namespace

### Describe the bug

The `web-streams-polyfill` package has a TypeScript bug where it pollutes the global namespace and misdefines `[Symbol.asyncIterator]` as a property rather than a method:

https://github.com/MattiasBuelens/web-streams-polyfill/blob/master/dist/types/ts3.6/polyfill.d.ts#L26

vs.

https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/AsyncIterator/@@asyncIterator

This causes subtle typing bugs downstream. Furthermore, `ReadableStream` is now supported in the LTS version of Node, so the polyfill is no longer necessary:

https://developer.mozilla.org/en-US/docs/Web/API/ReadableStream

### To Reproduce

Create an interface that extends both `ReadableStream` and `AsyncGenerator` after importing the SDK:

```ts
import OpenAI from "openai";

interface IterableReadableStreamInterface<T> extends ReadableStream<T>, AsyncGenerator<T> {}
```

### Code snippets

_No response_

### OS

macOS

### Node version

18

### Library version

4.24.1

jacoblee93:
Fixes #613


jacoblee93:
I added `"lib": ["DOM"],` to the `tsconfig.json` file to fix typing, and also removed the `AbortController` polyfill (this has been supported in Node since version 15: https://developer.mozilla.org/en-US/docs/Web/API/AbortController#browser_compatibility) 

I was able to set up tests via the OpenAPI spec, seems like there's a few unrelated errors:

Screenshot 2024-01-04 at 9 40 14 AM

This is a bit bigger of a change than initially planned, but should be supported by all LTS Node versions. You can also add special instructions for people to polyfill Node 16 like this if you're very worried:

https://js.langchain.com/docs/get_started/installation#unsupported-nodejs-16

LMK what you think!

jacoblee93:
May also be worth removing `node-fetch` since native fetch is now in Node 18 to avoid weird bundler issues, but that would be another PR.

rattrayalex:
> May also be worth removing node-fetch since native fetch is now in Node 18 to avoid weird bundler issues, but that would be another PR.


Ut is indeed: https://github.com/openai/openai-node/pull/402

tsconfig.json
```diff
@@ -3,7 +3,7 @@
   "exclude": ["src/_shims/**/*-deno.ts"],
   "compilerOptions": {
     "target": "es2020",
-    "lib": ["es2020"],
```
rattrayalex, line 6: ah, sadly I don't think this is an acceptable change b/c we don't want to drag DOM types into our users' projects if they are in a Node setting. Do you remember why it was needed?

rattrayalex (commented):
Thank you Jacob! 

Will need to have a member of our team review more thoroughly, but first will need to resolve question around `DOM`

rattrayalex:
BTW why is this only a partial fix?

tsconfig.json
jacoblee93: `ReadableStream` isn't defined in the base typing - let me see if I can find another workaround.

jacoblee93:
> BTW why is this only a partial fix?

Ah sorry that's a typo, another PR I did for LangChain.js was a partial fix for some typing issues we were seeing and I think my mind was on that.

jacoblee93:
With some light Googling, it seems that the only way to get the `ReadableStream` type without getting hacky and potentially causing more issues seems to be to include `DOM` in `tsconfig.json`.

We've been doing it upstream in LangChain.js for a while now with (seemingly) good compatibility:

https://github.com/langchain-ai/langchainjs/blob/main/tsconfig.json#L6

But if we don't feel comfortable doing that here, then perhaps it's best to hope for a good downstream resolution here and bump the polyfill later on:

https://github.com/MattiasBuelens/web-streams-polyfill/issues/141

rattrayalex:
K, then yes that might be best… 

Is `ReadableStream` in the `node` types?

jacoblee93:
This has been fixed in the downstream polyfill - feel free to close this!

https://github.com/MattiasBuelens/web-streams-polyfill/pull/142

PR #676, opened 2024-02-17
Fix Whisper non-json `response_format`

stefansundin:

## Changes being requested

I realized too late that these files are auto generated and I had already made the changes. Anyway, I think the PR describes the issue well. The output type is wrong for all `response_format` besides `json`. :(

## Additional context & links

`response_format: 'json'` should return this type: https://platform.openai.com/docs/api-reference/audio/json-object

`response_format: 'verbose_json'` should return this type: https://platform.openai.com/docs/api-reference/audio/verbose-json-object

the other response formats just return `string`

### DIFF
--- a/src/resources/audio/index.ts
+++ b/src/resources/audio/index.ts
@@ -2,5 +2,10 @@
 
 export { Audio } from './audio';
 export { SpeechCreateParams, Speech } from './speech';
-export { Transcription, TranscriptionCreateParams, Transcriptions } from './transcriptions';
-export { Translation, TranslationCreateParams, Translations } from './translations';
+export {
+  Transcription,
+  VerboseTranscription,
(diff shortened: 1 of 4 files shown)


rattrayalex:
Ah, yeah sorry this is generated so we'd have to do this a different way. I actually think we may have also previously decided against making this change, but to be honest I can't recall why, and it certainly would be nice for these types to be accurate. 

I'll leave the PR open for now, but chances are not high that we merge it. 

In any case, I do appreciate your work here1

stefansundin:
If it is impossible to provide the correct types for various tooling reasons, then I think using `any` is a better option than a potentially incorrect type. Is that possible?

rattrayalex:
Impossible is not the word, just not as quick as we'd like :)

PR #694, opened 2024-02-25
docs: Fix incorrect use of stream api event

iovecoldpizza:
- [ ✓] I understand that this repository is auto-generated and my pull request may not be merged

## Changes being requested
Update the use of stream API event
## Additional context & links
The example involved is about streaming response of Chat Completion, instead of Assistant.
As per https://github.com/openai/openai-node/blob/master/helpers.md#user-content-onchunk-chunk-chatcompletionchunk-snapshot-chatcompletionsnapshot---with-stream, the event "content" is for use with assistant API, while to obtain response chunk from Chat Completion API, "chunk" event should be used instead.
Also the "delta" object structure is different.

### DIFF
--- a/README.md
+++ b/README.md
@@ -118,8 +118,8 @@ async function main() {
     stream: true,
   });
 
-  stream.on('chunk', (delta, snapshot) => {
-    process.stdout.write(delta.choices[0]?.delta?.content || '');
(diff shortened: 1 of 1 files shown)


rattrayalex:
I don't believe the docs are incorrect. `.on('content'` is indeed a convenience in the chat completion streaming helper.

PR #718, opened 2024-03-14
[Azure] Introduce AzureOpenAI client

deyaaeldeen:

## Changes being requested

Add an Azure client similar to [Python's](https://github.com/openai/openai-python/blob/5429f69670e4db70f0cb7420ddb27c9bd11b9508/src/openai/lib/azure.py#L64).

To-dos:
- [ ]  Export it from the package
- [ ] Add support for Azure-specific models
- [ ] Add an example
- [ ] Add tests

### DIFF
--- a/src/lib/azure/azure.ts
+++ b/src/lib/azure/azure.ts
@@ -1,0 +1,172 @@
+// File generated from our OpenAPI spec by Stainless.
+
+import * as Core from '../../core';
+import * as Errors from '../../error';
+import OpenAI, { ClientOptions } from 'openai/index';
+import { RequestInit } from 'node-fetch';
+
+/** API Client for interfacing with the Azure OpenAI API. */
+export interface AzureClientOptions extends ClientOptions {
+  /**
+   * Defaults to process.env['OPENAI_API_VERSION'].
+   */
+  apiVersion?: string | undefined;
+  /**
+   * Your Azure endpoint, including the resource, e.g. `https://example-resource.azure.openai.com/`
+   */
+  endpoint?: string | undefined;
+  /**
+   * A model deployment, if given, sets the base client URL to include `/deployments/{deployment}`.
+   * Note: this means you won't be able to use non-deployment endpoints. Not supported with Assistants APIs.
+   */
+  deployment?: string | undefined;
+  /**
+   * Defaults to process.env['AZURE_OPENAI_API_KEY'].
+   */
+  apiKey?: string | undefined;
+  /**
+   * A function that returns a Microsoft Entra (formerly known as Azure Active Directory) access token, will be invoked on every request.
+   */
+  azureEntraTokenProvider?: (() => string) | undefined;
+}
+
+/** API Client for interfacing with the Azure OpenAI API. */
+export class AzureOpenAI extends OpenAI {
+  private _azureEntraTokenProvider: (() => string) | undefined;
+  private _apiVersion: string;
+  /**
+   * API Client for interfacing with the Azure OpenAI API.
+   *
+   * @param {string | undefined} [opts.endpoint=process.env['OPENAI_API_VERSION'] ?? undefined]
+   * @param {string | undefined} [opts.apiVersion=process.env['AZURE_OPENAI_ENDPOINT'] ?? undefined] - Your Azure endpoint, including the resource, e.g. `https://example-resource.azure.openai.com/`
+   * @param {string | undefined} [opts.apiKey=process.env['AZURE_OPENAI_API_KEY'] ?? undefined]
+   * @param {string | undefined} opts.deployment - A model deployment, if given, sets the base client URL to include `/deployments/{deployment}`.   * @param {string | null | undefined} [opts.organization=process.env['OPENAI_ORG_ID'] ?? null]
+   * @param {string} [opts.baseURL=process.env['OPENAI_BASE_URL']] - Sets the base URL for the API.
+   * @param {number} [opts.timeout=10 minutes] - The maximum amount of time (in milliseconds) the client will wait for a response before timing out.
+   * @param {number} [opts.httpAgent] - An HTTP agent used to manage HTTP(s) connections.
+   * @param {Core.Fetch} [opts.fetch] - Specify a custom `fetch` function implementation.
+   * @param {number} [opts.maxRetries=2] - The maximum number of times the client will retry a request.
+   * @param {Core.Headers} opts.defaultHeaders - Default headers to include with every request to the API.
+   * @param {Core.DefaultQuery} opts.defaultQuery - Default query parameters to include with every request to the API.
+   * @param {boolean} [opts.dangerouslyAllowBrowser=false] - By default, client-side use of this library is not allowed, as it risks exposing your secret API credentials to attackers.
+   */
+  constructor({
+    baseURL = Core.readEnv('OPENAI_BASE_URL'),
+    apiKey = Core.readEnv('AZURE_OPENAI_API_KEY'),
+    apiVersion = Core.readEnv('OPENAI_API_VERSION'),
+    endpoint,
+    deployment,
+    azureEntraTokenProvider,
+    ...opts
+  }: AzureClientOptions = {}) {
+    if (apiVersion === undefined) {
+      throw new Errors.OpenAIError(
+        "The OPENAI_API_VERSION environment variable is missing or empty; either provide it, or instantiate the AzureOpenAI client with an apiVersion option, like new AzureOpenAI({ apiVersion: 'My API Version' }).",
+      );
+    }
+
+    if (typeof azureEntraTokenProvider === 'function') {
+      opts.dangerouslyAllowBrowser = true;
+    }
+
+    if (!azureEntraTokenProvider && !apiKey) {
+      throw new Errors.OpenAIError(
+        'Missing credentials. Please pass one of `apiKey` and `azureEntraTokenProvider`, or set the `AZURE_OPENAI_API_KEY` environment variable.',
+      );
+    }
+
+    if (azureEntraTokenProvider && apiKey) {
+      throw new Errors.OpenAIError(
+        'The `apiKey` and `azureEntraTokenProvider` arguments are mutually exclusive; Only one can be passed at a time.',
+      );
+    }
+
+    // define a sentinel value to avoid any typing issues
+    apiKey ??= API_KEY_SENTINEL;
+
+    opts.defaultQuery = { ...opts.defaultQuery, 'api-version': apiVersion };
+
+    if (!baseURL) {
+      if (!endpoint) {
+        endpoint = process.env['AZURE_OPENAI_ENDPOINT'];
+      }
+
+      if (!endpoint) {
+        throw new Errors.OpenAIError(
+          'Must provide one of the `baseURL` or `endpoint` arguments, or the `AZURE_OPENAI_ENDPOINT` environment variable',
+        );
+      }
+
+      if (deployment) {
+        baseURL = `${endpoint}/openai/deployments/${deployment}`;
+      } else {
+        baseURL = `${endpoint}/openai`;
+      }
+    } else {
+      if (endpoint) {
+        throw new Errors.OpenAIError('baseURL and endpoint are mutually exclusive');
+      }
+    }
+
+    super({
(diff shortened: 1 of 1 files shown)


src/lib/azure/azure.ts
```diff
@@ -0,0 +1,179 @@
...
+   * @param {string} [opts.baseURL=process.env['OPENAI_BASE_URL']] - Sets the base URL for the API.
+   * @param {number} [opts.timeout=10 minutes] - The maximum amount of time (in milliseconds) the client will wait for a response before timing out.
+   * @param {number} [opts.httpAgent] - An HTTP agent used to manage HTTP(s) connections.
+   * @param {Core.Fetch} [opts.fetch] - Specify a custom `fetch` function implementation.
+   * @param {number} [opts.maxRetries=2] - The maximum number of times the client will retry a request.
+   * @param {Core.Headers} opts.defaultHeaders - Default headers to include with every request to the API.
+   * @param {Core.DefaultQuery} opts.defaultQuery - Default query parameters to include with every request to the API.
+   * @param {boolean} [opts.dangerouslyAllowBrowser=false] - By default, client-side use of this library is not allowed, as it risks exposing your secret API credentials to attackers.
+   */
```
rattrayalex, line 49: I don't believe we generally use jsdoc in this library, preferring docstrings in TypeScript types.

rattrayalex (commented):
@athyuttamre @kristapratico WDYT?

rattrayalex:
Thank you for putting this up @deyaaeldeen ! We're discussing internally as well.

src/lib/azure/azure.ts
deyaaeldeen: I was following the style used in the OpenAI client: https://github.com/openai/openai-node/blob/bb4bce30ff1bfb06d9d6c7a185d952121bb716e2/src/index.ts#L91-L104 but I can delete it if it doesn't break any tooling.

src/lib/azure/azure.ts
```diff
@@ -0,0 +1,179 @@
...
+
+    if (!microsoftEntraTokenProvider && !apiKey) {
+      throw new Errors.OpenAIError(
+        'Missing credentials. Please pass one of `apiKey` and `microsoftEntraTokenProvider`, or set the `AZURE_OPENAI_API_KEY` environment variable.',
+      );
+    }
+
+    if (microsoftEntraTokenProvider && apiKey) {
+      throw new Errors.OpenAIError(
+        'The `apiKey` and `microsoftEntraTokenProvider` arguments are mutually exclusive; Only one can be passed at a time.',
```
bterlson, line 78:
```suggestion
        'The `apiKey` and `microsoftEntraTokenProvider` arguments are mutually exclusive; only one can be passed at a time.',
```

rattrayalex: Hah, hi Brian! Fan of your work, fun to see you here.

src/lib/azure/azure.ts
rattrayalex: Ah, my mistake. Fine to leave then.

src/lib/azure/azure.ts
```diff
@@ -0,0 +1,179 @@
...
+}
+
+/** API Client for interfacing with the Azure OpenAI API. */
+export class AzureOpenAI extends OpenAI {
+  private _microsoftEntraTokenProvider: (() => string) | undefined;
+  /**
+   * API Client for interfacing with the Azure OpenAI API.
+   *
+   * @param {string | undefined} [opts.endpoint=process.env['OPENAI_API_VERSION'] ?? undefined]
+   * @param {string | undefined} [opts.apiVersion=process.env['AZURE_OPENAI_ENDPOINT'] ?? undefined] - Your Azure endpoint, including the resource, e.g. `https://example-resource.azure.openai.com/`
```
kristapratico, line 38:
```suggestion
   * @param {string | undefined} [opts.apiVersion=process.env['OPENAI_API_VERSION'] ?? undefined]
   * @param {string | undefined} [opts.endpoint=process.env['AZURE_OPENAI_ENDPOINT'] ?? undefined] - Your Azure endpoint, including the resource, e.g. `https://example-resource.azure.openai.com/`
```

src/lib/azure/azure.ts
```diff
@@ -0,0 +1,179 @@
...
+   * @param {string | undefined} [opts.apiKey=process.env['AZURE_OPENAI_API_KEY'] ?? undefined]
+   * @param {string | undefined} opts.deployment - A model deployment, if given, sets the base client URL to include `/deployments/{deployment}`.   * @param {string | null | undefined} [opts.organization=process.env['OPENAI_ORG_ID'] ?? null]
```
kristapratico, line 40: It looks like the doc for `organization` should be on a new line

kristapratico (commented):
Looks good :)

deyaaeldeen:
@rattrayalex I am need your help in exporting the Azure client. The `index.ts` module has the definition of the `OpenAI` client but I think ideally, that definition may need to live in its own module so that `index.ts` can export both `OpenAI` and `AzureOpenAI`. What do you think?

rattrayalex:
We'll take a look at that soon @deyaaeldeen. It may have to wait til next week.

rattrayalex:
thanks for the bump, this slipped through the cracks. I've poked internally.

pstern-sl:
Hi! I will take a look and test on my side, but in the interim can you elaborate on the change you are proposing regarding the modules design? (for instance, what are the tradeoffs and alternatives?)

rattrayalex (commented):
Is it possible to leave `index.ts` largely untouched, instead of moving code into `openai.ts`?

src/lib/azure/azure.ts
```diff
@@ -0,0 +1,185 @@
+import * as Core from '../../core';
+import * as Errors from '../../error';
+import { ClientOptions, type OpenAI } from '../../index';
+import { type RequestInit } from '../../_shims/index';
```
rattrayalex, line 4:
```suggestion
import type { ClientOptions, OpenAI } from '../../index';
import type { RequestInit } from '../../_shims/index';
```

src/index.ts
```diff
@@ -311,3 +311,8 @@ export namespace OpenAI {
 }
 
 export default OpenAI;
+
+// These lines are not auto-generated, Do not remove these lines.
+import { getAzureOpenAI, AzureClientOptions } from './lib/azure/azure';
```
rattrayalex, line 316: can the import be moved to the top?

src/index.ts
```diff
@@ -311,3 +311,8 @@ export namespace OpenAI {
 }
 
 export default OpenAI;
+
+// These lines are not auto-generated, Do not remove these lines.
```
rattrayalex, line 315:
```suggestion
```

Comments like this are not necessary.

src/lib/azure/azure.ts
```diff
@@ -0,0 +1,185 @@
...
+     * @param {Core.DefaultQuery} opts.defaultQuery - Default query parameters to include with every request to the API.
+     * @param {boolean} [opts.dangerouslyAllowBrowser=false] - By default, client-side use of this library is not allowed, as it risks exposing your secret API credentials to attackers.
+     */
+    constructor({
+      baseURL = Core.readEnv('OPENAI_BASE_URL'),
+      apiKey = Core.readEnv('AZURE_OPENAI_API_KEY'),
+      apiVersion = Core.readEnv('OPENAI_API_VERSION'),
+      endpoint,
+      deployment,
+      microsoftEntraTokenProvider,
```
rattrayalex, line 61: In the Python client, do we use the term `Entra`?

rattrayalex (commented):
Thanks for the iteration here! I think its moving in a good direction. Just a few comments from a quick skim, but we need to discuss internally a bit on Monday.

Hopefully @RobertCraigie can provide a more thorough review soon.

src/lib/azure/azure.ts
deyaaeldeen: No, but it is a recent rebrand: https://www.microsoft.com/security/business/identity-access/microsoft-entra-id

RobertCraigie:
Hey we've moved the `OpenAI` class definition to a separate `src/client.ts` file so you can properly inherit from it now :)

deyaaeldeen:
@RobertCraigie Awesome! I updated the client in https://github.com/openai/openai-node/pull/718/commits/780ddbbbe45a068c045f24d8a6bfd62efa94cbde.

RobertCraigie:
Unfortunately we just had to revert the `client.ts` file change because it broke in certain cases: https://github.com/openai/openai-node/issues/816

I tried putting the azure exports at the bottom of the `index.ts` file and that works in some environments but not all :/

what would you think about this being the main import?

```ts
import { AzureOpenAI } from 'openai/azure';
```
---
Note that we tried a bunch of different ways of structuring the separate `index.ts` / `client.ts` files and it does not seem possible to make that change in a compatible fashion so we'll be looking into splitting them up in the next major version.

RobertCraigie:
Actually, you could define the azure client class at the bottom of the `index.ts` file instead, that'll result in the best end user experience

deyaaeldeen:
@RobertCraigie Thanks for the feedback! I updated the PR accordingly.

src/index.ts
```diff
@@ -310,4 +310,182 @@ export namespace OpenAI {
...
+   */
+  deployment?: string | undefined;
+  /**
+   * Defaults to process.env['AZURE_OPENAI_API_KEY'].
+   */
+  apiKey?: string | undefined;
+  /**
+   * A function that returns a Microsoft Entra (formerly known as Azure Active Directory) access token, will be invoked on every request.
+   */
+  microsoftEntraTokenProvider?: (() => string) | undefined;
```
RobertCraigie, line 337: is not supporting the equivalent `azure_ad_token` as well like in the [Python client](https://github.com/openai/openai-python/blob/aed1e43745cd6358b4bafd3a39b3dfeee5e31a03/src/openai/lib/azure.py#L166) an intentional mismatch? (I haven't read all the context)

in other words, would people be suprised they can't set the `AZURE_OPENAI_AD_TOKEN` environment variable and have it picked up here?

RobertCraigie (commented):
Code changes look good to me. We'll also want to update the README.md docs.

Let me know when the docs have been updated & I'll port these changes over to our side before merging.

tests/azure.test.ts
RobertCraigie, line 1: nit: would be ideal for this to be in `tests/lib/azure.test.ts`, these are good tests though :)

src/index.ts
rattrayalex: @deyaaeldeen in addition to Robert's comments, in the Python client I believe we call this `azure_ad` not `microsoft_entra` – is that correct? Can we keep the naming consistent and call it Azure Active Directory instead?

rattrayalex (commented):
Thanks for your patience here!

Few things outstanding:

1. Entra -> AD, as noted elsewhere
2. Need to update the Azure section of the README (or add one, similar to Python)

rattrayalex:
Ah sorry, @deyaaeldeen my mistake – we actually don't need the README updated, we've done that part on our end. Just the Entra -> Azure AD change would be great.

src/index.ts
deyaaeldeen: > is not supporting the equivalent azure_ad_token as well like in the [Python client](https://github.com/openai/openai-python/blob/aed1e43745cd6358b4bafd3a39b3dfeee5e31a03/src/openai/lib/azure.py#L166) an intentional mismatch?

Yes, I would like our customers to use the token provider first because it takes care of token refreshing. I can add `azureAdToken` in the future if it turns out to be needed.

>  Can we keep the naming consistent and call it Azure Active Directory instead?

Ok I made the update in https://github.com/openai/openai-node/pull/718/commits/670165f52c4599f8406364b9778646ebb5a131ab and we can plan the switch to the Entra name in the next major version.

tests/azure.test.ts
deyaaeldeen: I moved it in https://github.com/openai/openai-node/commit/670165f52c4599f8406364b9778646ebb5a131ab.

rattrayalex:
This has been moved to https://github.com/openai/openai-node/pull/822, which merged, and will be released in https://github.com/openai/openai-node/pull/823

PR #721, opened 2024-03-15
Handle process.env being undefined in debug func

jkarthic:

## Changes being requested
Please handle the case where `process.env` is undefined.

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,12 +1,12 @@
 {
-  "name": "@get-wrecked/openai",
+  "name": "openai",
   "version": "4.29.0",
   "description": "The official TypeScript library for the OpenAI API",
   "author": "OpenAI <support@openai.com>",
   "types": "dist/index.d.ts",
   "main": "dist/index.js",
   "type": "commonjs",
-  "repository": "github:get-wrecked/openai-node",
+  "repository": "github:openai/openai-node",
   "license": "Apache-2.0",
   "packageManager": "yarn@1.22.21",
   "files": [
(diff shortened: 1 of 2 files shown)


package.json
```diff
@@ -1,12 +1,12 @@
...
-  "name": "openai",
+  "name": "@get-wrecked/openai",
   "version": "4.29.0",
   "description": "The official TypeScript library for the OpenAI API",
   "author": "OpenAI <support@openai.com>",
   "types": "dist/index.d.ts",
   "main": "dist/index.js",
   "type": "commonjs",
-  "repository": "github:openai/openai-node",
+  "repository": "github:get-wrecked/openai-node",
```
rattrayalex, line 9: These changes are, I hope, unintentionally left in from development

src/core.ts
```diff
@@ -1075,7 +1075,7 @@ function applyHeadersMut(targetHeaders: Headers, newHeaders: Headers): void {
 }
 
 export function debug(action: string, ...args: any[]) {
-  if (typeof process !== 'undefined' && process.env['DEBUG'] === 'true') {
+  if (process?.env?.['DEBUG'] === 'true') {
```
rattrayalex, line 1078:
```suggestion
  if (typeof process !== 'undefined' && process?.env?.['DEBUG'] === 'true') {
```
we still need the typeof check for environments where `process` is not defined at all.

rattrayalex (commented):
Thanks for the contribution! Few changes requested.

package.json
jkarthic: yes. Sorry about that. All fixed now. 
Thanks for the quick response.

rattrayalex (approved):

rattrayalex:
Thank you @jkarthic !

dackerman:
Thanks @jkarthic , we merged an update to our codegen to reflect your suggestion. It should go out in the next release!

jkarthic:
Thank you very much @rattrayalex @dackerman

PR #851, opened 2024-05-15
Added 'vision' to FileCreateParam purpose enum to support image files…

MathiasGilson:
Added `vision` as a new purpose when uploading a new file to support images for assistant use as specified in the [documentation](https://platform.openai.com/docs/api-reference/files/create)


erinkav-openai (approved):

rattrayalex:
Thanks for the contribution; this has since been updated in the OpenAPI spec: https://github.com/openai/openai-node/blob/d094e838b1d3877b518e861555cb19c21c3b85cd/src/resources/files.ts#L171

PR #865, opened 2024-05-22
Refactor: Inline calculation of last index of data

HitanshuShah1001:
- [ ✗ ] I understand that this repository is auto-generated and my pull request may not be merged

## Changes being requested

"Refactor: Inline calculation of last index of data"


rattrayalex:
Thanks for the suggestion, but we will be keeping this as-is.

PR #868, opened 2024-05-28
Move @types/node and @types/node-fetch to devDependencies

tilfin:

## Changes being requested
Move `@types/node` and `@types/node-fetch` to `devDependencies`. These type modules are not required for the production build of applications using this library and should therefore be listed as `devDependencies` instead of `dependencies`.

## Additional context & links
Unnecessary inclusion of type modules in dependencies can lead to bloated production builds. By moving them to devDependencies, we can keep the production build leaner.

### DIFF
--- a/package.json
+++ b/package.json
@@ -24,8 +24,6 @@
     "fix": "eslint --fix --ext ts,js ."
   },
   "dependencies": {
-    "@types/node": "^18.11.18",
-    "@types/node-fetch": "^2.6.4",
     "abort-controller": "^3.0.0",
     "agentkeepalive": "^4.2.1",
     "form-data-encoder": "1.7.2",
@@ -37,6 +35,8 @@
     "@swc/core": "^1.3.102",
     "@swc/jest": "^0.2.29",
     "@types/jest": "^29.4.0",
+    "@types/node": "^18.11.18",
+    "@types/node-fetch": "^2.6.4",
     "@typescript-eslint/eslint-plugin": "^6.7.0",
     "@typescript-eslint/parser": "^6.7.0",
     "eslint": "^8.49.0",


rattrayalex:
These should not affect production builds if you use a tree-shaking bundler, but are required to provide correct types in some situations.

tilfin:
@rattrayalex Thank you for your response.

I would like to point out that tree shaking node_modules, including type definitions, is not a common practice in Node.js projects. Most Node.js applications are not bundled in a way that would allow for effective tree shaking of dependencies.

While it's true that tree shaking bundlers can potentially eliminate unused code from dependencies in browser-based applications, this approach is rarely applied to server-side Node.js projects.

And, could you please elaborate on the 'some situations' where these type definitions are required in production builds?

rattrayalex:
These types are not just internally used, they are part of the API exposed by the package, so they must be specified in `dependencies` – otherwise TypeScript users would get errors when using the package. 

You can look forward to the dependency weight going down once our work to remove `node-fetch` as a dependency is complete! See https://github.com/openai/openai-node/pull/402 (though most work is happening behind the scenes there right now).

tilfin:
I understand your intention now. It's because of the `imports: { "openai/*": "./src/*" }` in package.json, which allows imports like:

```typescript
import { ChatCompletionUserMessageParam } from 'openai/resources/chat/completions';
```

I see why it's necessary.

However, I didn't notice this at all because there's no mention of it in the README and VSCode autocomplete doesn't work for these imports.

In my case, I've been using it like this:

```typescript
import OpenAI, { APIError } from 'openai';
export type OpenAIChatUserMessage =
  OpenAI.Chat.Completions.ChatCompletionUserMessageParam;
```

PR #891, opened 2024-06-10
fix: Added min and max length to `isAbsoluteUrl` regex

nicolasperez19:
## Changes being requested
The [`isAbsoluteUrl`](src/core.ts) function uses a regex to check whether a string is an absolute URL by essentially checking if it has a URI scheme:
```ts
const startsWithSchemeRegexp = new RegExp('^(?:[a-z]+:)?//', 'i');
const isAbsoluteURL = (url: string): boolean => {
  return startsWithSchemeRegexp.test(url);
};
```
The regex checks for the URI scheme by checking a set of `a-z` characters, then a colon `:` and then two backslashes `//`:
image
 
The issue with the regex for checking the amount of `a-z` characters is unbounded. This means an extremely long set of `a-z` characters can be passed and the regex will still try to test it, which wastes a lot of computation.
image

Realistically [the longest official IANA URL scheme is about 36 characters](https://www.iana.org/assignments/uri-schemes/uri-schemes.xhtml), so testing any URL scheme above 50 `a-z` characters is definitely a waste of time. 

In my pull request, I added a constraint to check for a maximum of 50 `a-z` characters.

### DIFF
--- a/src/core.ts
+++ b/src/core.ts
@@ -956,7 +956,7 @@ export const safeJSON = (text: string) => {
 };
 
 // https://stackoverflow.com/a/19709846
-const startsWithSchemeRegexp = new RegExp('^(?:[a-z]{1,50}+:)?//', 'i');
+const startsWithSchemeRegexp = new RegExp('^(?:[a-z]{1,50}:)?//', 'i');
 const isAbsoluteURL = (url: string): boolean => {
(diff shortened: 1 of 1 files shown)


rattrayalex:
Thanks for the PR! We'll make this change in our codegen system.

PR #927, opened 2024-07-10
fix: Correct findDoubleNewlineIndex to iterate to the second last pos…

fEyebrow:
…ition

## Changes being requested

findDoubleNewlineIndex cannot find “\n\n” at the end, so correct findDoubleNewlineIndex to iterate to the second last position.


RobertCraigie:
Thanks for the PR @fEyebrow, do you have an example script to showcase the issue with the current behaviour?

PR #929, opened 2024-07-10
Update transcriptions.ts, fix return type

jedelacarrera:
## Changes being requested
The audio.transcriptions.create method returns a string, not a Transcription object as stipulated in the source code
The response type should be updated to `string`

## Additional context & links
```js
    const transcription = await client.audio.transcriptions.create({
      model: 'whisper-1',
      file: fs.createReadStream(filename),
      response_format: 'text',
    });

    // Transcription should be an object with a text property, but it returns a string
    console.log('Transcription:', transcription);
    console.log('Output type:`, typeof transcription);
```

Console output:
```
Transcription: This is a test audio file

Output type: string
```

### DIFF
--- a/src/resources/audio/transcriptions.ts
+++ b/src/resources/audio/transcriptions.ts
@@ -8,7 +8,7 @@ export class Transcriptions extends APIResource {
   /**
    * Transcribes audio into the input language.
    */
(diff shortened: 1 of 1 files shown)


jedelacarrera:
Not actually necessary, using  `response_format: 'json'` works as expected.
It's still confusing if you use another response_format because the output doesn't match the typescript expected output, but not actually a bug

PR #940, opened 2024-07-18
Bugfix for iterPages

alexmakeev:
During check of nextPage existence the field "has_more" might be used (gotten at body). Without it, iteration of, say all thread messages list brings to infinite loop.
Body field has unknown type (might be also detailed), so I used cast to "any" to skip linter/ts issues.


RobertCraigie:
Thanks for the PR @alexmakeev, can you share a reproduction script for the infinite loop issue?

I'm surprised that could happen as I was under the impression the API would respond with an empty array when there aren't any entries left.

PR #952, opened 2024-07-25
Remove web stream polyfill

### LINKED ISSUE #951: Stop using polyfills for Node.js

My understanding is that this module use [polyfills](https://github.com/openai/openai-node/blob/22896e88e97c0e8621bf74c666a6fc7d9d832267/src/_shims/node-runtime.ts) for Node.js.
There is no need for doing so given you support Node.js v18+.

This causes compatibility issues with the rest of the ecosystem.

Ref https://github.com/fastify/fastify/issues/5584

### Additional context

_No response_

Ethan-Arrowood:
Fixes #951 

Removes the polyfill for `ReadableStream` and uses the Node.js implementation from `node:stream/web`. Should resolve the issue of `Readable.fromWeb` not working correctly.


RobertCraigie:
closing as this change was made separately in https://github.com/openai/openai-node/pull/954

PR #953, opened 2024-07-25
docs(fix): correct examples in documentation

kermitsxb:

## Changes being requested

Quick fix in the base example of the README.md

## Additional context & links

Current version provide this error while running the example as the instance is stored in the `client` const.
```
index.ts:8:32 - error TS2552: Cannot find name 'openai'. Did you mean 'OpenAI'?

const chatCompletion = await openai.chat.completions.create({
```

### DIFF
--- a/README.md
+++ b/README.md
@@ -37,7 +37,7 @@ const client = new OpenAI({
 });
 
 async function main() {
-  const chatCompletion = await openai.chat.completions.create({
+  const chatCompletion = await client.chat.completions.create({
     messages: [{ role: 'user', content: 'Say this is a test' }],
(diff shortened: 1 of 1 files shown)


RobertCraigie:
thanks @kermitsxb, turns out we had a bunch of other places that were using the incorrect client variable name so we fixed this separately in https://github.com/openai/openai-node/issues/955

RobertCraigie:
thanks @kermitsxb, turns out we had a bunch of other places that were using the incorrect client variable name so we fixed this separately in https://github.com/openai/openai-node/issues/955

PR #1003, opened 2024-08-17
[Azure] Refresh AAD token on retry

deyaaeldeen:

## Changes being requested

The AAD token may need to be refreshed between retries so this fix makes sure to always call the token provider irrespective of whether the authorization header is being set.

### DIFF
--- a/src/index.ts
+++ b/src/index.ts
@@ -485,6 +485,12 @@ export class AzureOpenAI extends OpenAI {
   }
 
   protected override async prepareOptions(opts: Core.FinalRequestOptions<unknown>): Promise<void> {
+    /**
+     * The user should provide a bearer token provider if they want
+     * to use Azure AD authentication. The user shouldn't set the
+     * Authorization header manually because the header is overwritten
+     * with the Azure AD token if a bearer token provider is provided.
+     */
     if (opts.headers?.['api-key']) {
       return super.prepareOptions(opts);
     }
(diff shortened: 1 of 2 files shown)


src/index.ts
```diff
@@ -485,7 +485,7 @@ export class AzureOpenAI extends OpenAI {
   }
 
   protected override async prepareOptions(opts: Core.FinalRequestOptions<unknown>): Promise<void> {
-    if (opts.headers?.['Authorization'] || opts.headers?.['api-key']) {
```
RobertCraigie, line 488: This is a bit awkward as users could set the `Authorization` header themselves and then we'd override it.

I think a better solution would be to do the same thing as we do in the Python SDK where we create a copy of the options before it's passed to this `prepareOptions()` function

deyaaeldeen: The Azure client doesn't support users setting the authorization header manually. The constructor requires either an API key or a bearer token provider function. If the user really wants to set the authorization header manually, they should provide a bearer token provider function that returns just that value.

RobertCraigie: 👍 mind adding that context as a comment to this code saying why it's okay to unconditionally overwrite the `Authorization` header?

RobertCraigie (approved):
LGTM!

src/index.ts
deyaaeldeen: Added in https://github.com/openai/openai-node/pull/1003/commits/e557ea4da4582a8fb9ce20f1d8be481e5a27f08d

PR #1018, opened 2024-08-28
fix(zod): correct param type for zodResponseFormat

darrenmce:

## Changes being requested

a fix on the type (i wrote the test just as an easy way to make sure it was correct)


RobertCraigie:
🤦 thanks so much for the PR!

PR #1054, opened 2024-09-10
Edit pass on Azure OpenAI section of README

scottaddie:
**Summary of changes**
- Rename Azure AD-focused variable used in code sample. The product was renamed to Microsoft Entra ID.
- Remove "en-us" locale string from Learn link. The user's preferred locale should be honored, not overridden.


RobertCraigie (approved):
Thanks!

PR #1055, opened 2024-09-10
Improved comment for AzureOpenAI's baseURL

AngelosP:
- [ X ] I understand that this repository is auto-generated and my pull request may not be merged

## Changes being requested
Improved the comment for AzureOpenAI's baseURL constructor parameter to offer an example. It's not easy to deduce that the /openai/ part should be included, so I think it's helpful to have an example that points that out and it's consistent with the comments of other constructor parameters.


RobertCraigie (approved):
Thanks!

PR #1056, opened 2024-09-11
Implement Parse Completion with raw response output

margostino:

## Changes being requested

Support `withResponse()` when parse completion is used.


RobertCraigie:
Thanks for the PR! I think there's a better way we can go about this though so I've opened #1058

PR #1073, opened 2024-09-13
fix(runTools): correct request options type

granmoe:
See comment here https://github.com/openai/openai-node/pull/1064#discussion_r1757724572


RobertCraigie (approved):
Thanks!

PR #1078, opened 2024-09-16
feat(client): add ._request_id property to object responses

RobertCraigie:
on the tin

### DIFF
--- a/README.md
+++ b/README.md
@@ -361,6 +361,17 @@ Error codes are as followed:
 | >=500       | `InternalServerError`      |
 | N/A         | `APIConnectionError`       |
 
+## Request IDs
+
+> For more information on debugging requests, see [these docs](https://platform.openai.com/docs/api-reference/debugging-requests)
+
+All object responses in the SDK provide a `_request_id` property which is added from the `x-request-id` response header so that you can quickly log failing requests and report them back to OpenAI.
+
+```ts
+const completion = await client.chat.completions.create({ messages: [{ role: 'user', content: 'Say this is a test' }], model: 'gpt-4' });
+console.log(completion._request_id) // req_123
+```
+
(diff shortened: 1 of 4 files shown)


src/core.ts
```diff
@@ -69,26 +69,44 @@ async function defaultParseResponse<T>(props: APIResponseProps): Promise<T> {
...
+  : T extends Record<string, any> ? T & { _request_id?: string | null }
+  : T;
+
+function _addRequestID<T>(value: T, response: Response): WithRequestID<T> {
+  if (!value || typeof value !== 'object' || Array.isArray(value)) {
+    return value as WithRequestID<T>;
+  }
+
+  return Object.defineProperty(value, '_request_id', {
+    value: response.headers.get('x-request-id'),
```
athyuttamre, line 93: Will this error if the value doesn't exist, or return undefined?

athyuttamre (approved):

README.md
```diff
@@ -361,6 +361,17 @@ Error codes are as followed:
...
 
+## Request IDs
+
+> For more information on debugging requests, see [these docs](https://platform.openai.com/docs/api-reference/debugging-requests)
+
+All object responses in the SDK provide a `_request_id` property which is added from the `x-request-id` response header so that you can quickly log failing requests and report them back to OpenAI.
+
+```ts
+const completion = await client.chat.completions.create({ messages: [{ role: 'user', content: 'Say this is a test' }], model: 'gpt-4' });
+console.log(completion._request_id) // req_id_1234
```
athyuttamre, line 372: `req_123`

src/core.ts
RobertCraigie: it'll return undefined

athyuttamre (approved):

Issue #1180, opened 2024-11-11
Model seems to ignore `.optional()` fields, and instead uses fasly values such as 0, empty strings etc'


uriva:
### Confirm this is a Node library issue and not an underlying OpenAI API issue

- [X] This is an issue with the Node library

### Describe the bug

Model seems to ignore `.optional()` fields, and instead uses nullary values, empty strings etc'

I'm not sure if it's a library issue or chatgpt model issue.

### To Reproduce

I suspect this is a KI or an upstream problem, but let me know if not I'll send a repro

### OS

ubuntu

### Node version

22.7.0

### Library version

4.57.0

RobertCraigie:
Thanks for reporting!  

This sounds like an issue with the underlying OpenAI API and not the SDK, so I'm going to go ahead and close this issue. When you use `.optional()` we mark the property as required and then update the schema to mark it as nullable.

Would you mind reposting at [community.openai.com](https://community.openai.com)?

jovermier:
This may be the line of code that converts an optional property to required.
[propDef.isOptional() && !refs.openaiStrictMode ? acc.required : [...acc.required, propName],](https://github.com/openai/openai-node/blob/a92cc1dbc4ab3284c6654d69d5c39399a867f601/src/_vendor/zod-to-json-schema/parsers/object.ts#L54)

I don't see any code for making the property nullable. In my testing I don't see the output schema nullable either.

uriva:
also just to be compliant with the `zod` type, `undefined` is considerably better if possible.

silasalberti:
As discussed above, the API doesn't allow optional fields and the [docs](https://platform.openai.com/docs/guides/structured-outputs#all-fields-must-be-required) recommend nullable fields as a workaround. The SDK already supports Zod's `.nullable()` which achieves this purpose. People might be unaware about that nuance, so I propose adding a warning on `.optional()` that recommends using `.nullable()` instead: #1214 

(I briefly thought about instead making `.optional()` have the same behavior as `.nullable()`. However, as mentioned above that would not be compliant with the Zod type, i.e. zod validation fails if it gets null instead of undefined.)

RobertCraigie:
Sorry for the delayed fix, in v4.95.1 a warning will be emitted if you try to use an `.optional()` field without also specifying `.nullable()`.

```ts
zodResponseFormat(
  z.object({
    required: z.string(),  // no warning
    optional: z.string().optional(),  // warning
[...]

PR #1214, opened 2024-12-09
feat: add warning for .optional() usage in OpenAI API schemas

silasalberti:
# Add warning for .optional() usage in OpenAI API schemas

This PR adds a warning when `.optional()` is used in schemas for OpenAI API Structured Outputs, recommending the use of `.nullable()` instead since optional fields are not currently supported.

## Changes
- Added warning in `optional.ts` that triggers when `openaiStrictMode` is true
- Warning includes explanation and link to documentation
- Added test to verify warning behavior

## Testing
✅ All tests passing (411 total)
- Added new test case to verify warning is emitted
- Verified warning message contains correct information
- Ran full test suite to ensure no regressions

## Implementation Details
The warning is implemented in the optional field parser and only triggers when generating schemas specifically for OpenAI API usage (when `openaiStrictMode` is true). This ensures developers get immediate feedback about unsupported optional fields while keeping the warning relevant only to OpenAI API usage.

Created with help from Devin: https://preview.devin.ai/devin/13e97f93e3b1445c91e60da01b88febd

Fixes openai/openai-node#1180

### DIFF
--- a/src/_vendor/zod-to-json-schema/parsers/optional.ts
+++ b/src/_vendor/zod-to-json-schema/parsers/optional.ts
@@ -7,8 +7,8 @@ export const parseOptionalDef = (def: ZodOptionalDef, refs: Refs): JsonSchema7Ty
     const fieldName = refs.propertyPath?.slice(-1)[0] || 'unknown';
     console.warn(
       `Warning: Field "${fieldName}" uses .optional() which is not supported by OpenAI API Structured Outputs. ` +
-      `Please use .nullable() instead. ` +
-      `See: https://platform.openai.com/docs/guides/structured-outputs#all-fields-must-be-required`
+        `Please use .nullable() instead. ` +
+        `See: https://platform.openai.com/docs/guides/structured-outputs#all-fields-must-be-required`,
     );
   }
 
--- a/src/_vendor/zod-to-json-schema/parsers/optional.ts
+++ b/src/_vendor/zod-to-json-schema/parsers/optional.ts
@@ -3,6 +3,15 @@ import { JsonSchema7Type, parseDef } from '../parseDef';
 import { Refs } from '../Refs';
 
 export const parseOptionalDef = (def: ZodOptionalDef, refs: Refs): JsonSchema7Type | undefined => {
+  if (refs.openaiStrictMode) {
+    const fieldName = refs.propertyPath?.slice(-1)[0] || 'unknown';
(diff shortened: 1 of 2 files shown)


src/_vendor/zod-to-json-schema/parsers/optional.ts
```diff
@@ -3,6 +3,15 @@ import { JsonSchema7Type, parseDef } from '../parseDef';
 import { Refs } from '../Refs';
 
 export const parseOptionalDef = (def: ZodOptionalDef, refs: Refs): JsonSchema7Type | undefined => {
+  if (refs.openaiStrictMode) {
+    const fieldName = refs.propertyPath?.slice(-1)[0] || 'unknown';
```
PoornakumarRasiraju, line 7: Since refs.propertyPath is an array, can we replace `.slice(-1)[0]` with `.at(-1)` here? Using .at(-1) improves readability and performance by directly accessing the last element without creating a new array.
https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/at

src/_vendor/zod-to-json-schema/parsers/optional.ts
```diff
@@ -3,6 +3,15 @@ import { JsonSchema7Type, parseDef } from '../parseDef';
...
+    console.warn(
+      `Warning: Field "${fieldName}" uses .optional() which is not supported by OpenAI API Structured Outputs. ` +
+        `Please use .nullable() instead. ` +
+        `See: https://platform.openai.com/docs/guides/structured-outputs#all-fields-must-be-required`,
+    );
+  }
+
```
PoornakumarRasiraju, line 14: Unnecessary use of string concatenation (+) since we are already using template literals.

tests/helpers/zod.test.ts
```diff
@@ -88,6 +88,24 @@ describe('zodResponseFormat', () => {
...
+        required: z.string(),
+        optional: z.string().optional(),
+      }),
+      'test',
+    );
+
+    expect(consoleSpy).toHaveBeenCalledWith(
+      expect.stringContaining('uses .optional() which is not supported by OpenAI API Structured Outputs'),
+    );
+
```
PoornakumarRasiraju, line 105: Could we also test `expect(consoleSpy).toHaveBeenCalledTimes(1);` to ensure this is called only once.

PR #1218, opened 2024-12-10
fix(logs/azure): redact sensitive header when DEBUG is set

minhanh-phan:

## Changes being requested
Redact values for `authorization` and `api-key` when `DEBUG` value is set to true.

## Additional context & links
https://github.com/openai/openai-python/pull/1850


tests/qs/utils.test.ts
RobertCraigie, line 1: this should be in `tests/index.test.ts` instead

src/core.ts
```diff
@@ -1139,8 +1139,27 @@ function applyHeadersMut(targetHeaders: Headers, newHeaders: Headers): void {
 }
 
 export function debug(action: string, ...args: any[]) {
+  const sensitiveHeaders = ["authorization", "api-key"];
```
RobertCraigie, line 1142: nit: would be better to define this outside the function and as a `Set` imo

```suggestion
const SENSITIVE_HEADERS = new Set(["authorization", "api-key"]);

export function debug(action: string, ...args: any[]) {
```

src/core.ts
```diff
@@ -1139,8 +1139,27 @@ function applyHeadersMut(targetHeaders: Headers, newHeaders: Headers): void {
...
-    console.log(`OpenAI:DEBUG:${action}`, ...args);
+    for (const arg of args) {
+      if (!arg) {
+        break;
+      }
+      // Check for sensitive headers in request body "headers"
+      if (arg["headers"]) {
+        for (const header in arg["headers"]){
+          if (sensitiveHeaders.includes(header.toLowerCase())){
+            arg["headers"][header] = "REDACTED";
```
RobertCraigie, line 1152: this shouldn't mutate the original inputs

src/core.ts
minhanh-phan: Thank you. I updated based on the suggestion

src/core.ts
minhanh-phan: I change the implementation to create a copy of the original inputs and redact the headers from the copied variables instead.

src/core.ts
```diff
@@ -1138,9 +1138,33 @@ function applyHeadersMut(targetHeaders: Headers, newHeaders: Headers): void {
...
+      if (!arg) {
+        return arg;
+      }
+
+      const modifiedArg = { ...arg };
+      // Check for sensitive headers in request body 'headers' object
+      if (modifiedArg['headers']) {
+        for (const header in modifiedArg['headers']) {
+          if (SENSITIVE_HEADERS.has(header.toLowerCase())) {
+            modifiedArg['headers'][header] = 'REDACTED';
```
RobertCraigie, line 1155: I think this will still mutate the original `headers` property as `...` is just a shallow clone

tests/index.test.ts
```diff
@@ -411,3 +411,59 @@ describe('retries', () => {
...
+    jest.resetModules();
+    process.env = { ...env };
+    process.env['DEBUG'] = 'true';
+  });
+
+  afterEach(() => {
+    process.env = env;
+  });
+
+  test('body request object with Authorization header', function () {
```
RobertCraigie, line 429: can you also add a test to ensure the inputs aren't mutated?

RobertCraigie (approved):
small comment on mutation and tests, otherwise LGTM, thanks!

src/core.ts
minhanh-phan: That's a good point, thank you!

tests/index.test.ts
minhanh-phan: Updated

RobertCraigie:
I made some changes to avoid cloning until we need to and to avoid using JSON round-tripping, lmk if that still looks good to you.

minhanh-phan:
@RobertCraigie The changes look goods to me. Thank you for the constructive feedback!

RobertCraigie:
So sorry this slipped through the cracks and wasn't merged earlier

Issue #302, opened 2023-09-09
process.env.DEBUG variable can have many values


OleksandrKucherenko:
### Confirm this is a Node library issue and not an underlying OpenAI API issue

- [X] This is an issue with the Node library

### Describe the bug

process.env.DEBUG variable can be used by many libraries, for example: `debug npm`

more details: https://github.com/debug-js/debug#windows-command-prompt-notes

in other words DEBUG environment variable can be something more complicated than just a `true` or undefined.

### To Reproduce

https://github.com/openai/openai-node/blob/master/src/core.ts#L994-L998

```bash 
DEBUG=* yarn start:dev
DEBUG=*,-openai  yarn start:dev
DEBUG=openai yarn start:dev
```

### Code snippets

```JavaScript
https://github.com/openai/openai-node/blob/master/src/core.ts#L994-L998

 
DEBUG=* yarn start:dev # max verbosity
DEBUG=*,-openai  yarn start:dev # show all debug msgs except openai
DEBUG=openai yarn start:dev # show only openai logs
```

### OS

macOs, Windows, Linux

### Node version

any

### Library version

4.6.0

rattrayalex:
Thank you for the feature request; we'll think this over.

vinnymac:
It's been quite a while now. Assuming you have not yet determined how you want to handle this, would you be open to a PR that at the very least allows us to disable the debug logs with another env, such as `process.env.DISABLE_DEBUG`?

My nodejs applications make heavy use of debug flags and openai is the only dependency we have that fills the log up with noise when we aren't asking for it.

rattrayalex:
Thanks for the bump – there's a good chance a PR like that would be useful, feel free to put one up (though we may not merge as-is, the general idea I agree is good).

RobertCraigie:
This has been replaced with an `OPENAI_LOG` environment variable in v5!

See https://github.com/openai/openai-node#logging for more information.

PR #1222, opened 2024-12-11
feat: add env to disable debug log

vinnymac:

## Changes being requested

Adds the ability to completely disable debug logging.

This is a stopgap, if I was requesting an actual fix it would be to implement an inversion of control, and enable the user of this sdk to specify the logger, that way this flag becomes unnecessary. This simple workaround is more than enough for my needs.

## Additional context & links

#302

### DIFF
--- a/src/core.ts
+++ b/src/core.ts
@@ -877,7 +877,7 @@ const getPlatformProperties = (): PlatformProperties => {
       'X-Stainless-Arch': normalizeArch(Deno.build.arch),
       'X-Stainless-Runtime': 'deno',
       'X-Stainless-Runtime-Version':
-        typeof Deno.version === 'string' ? Deno.version : Deno.version?.deno ?? 'unknown',
+        typeof Deno.version === 'string' ? Deno.version : (Deno.version?.deno ?? 'unknown'),
     };
   }
   if (typeof EdgeRuntime !== 'undefined') {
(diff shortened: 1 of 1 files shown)


src/core.ts
```diff
@@ -1139,7 +1139,11 @@ function applyHeadersMut(targetHeaders: Headers, newHeaders: Headers): void {
 }
 
 export function debug(action: string, ...args: any[]) {
-  if (typeof process !== 'undefined' && process?.env?.['DEBUG'] === 'true') {
+  if (
+    typeof process !== 'undefined' &&
+    typeof process?.env?.['OPENAI_DISABLE_DEBUG'] !== 'string' &&
+    process?.env?.['DEBUG'] === 'true'
+  ) {
```
nsaiprakash90, line 1146: Can we simplify the condition by avoiding redundant optional chaining (?.) on process.env.
```
if (
  typeof process !== 'undefined' &&
  process?.env && 
  typeof process.env['OPENAI_DISABLE_DEBUG'] !== 'string' &&
  process.env['DEBUG'] === 'true'
) {
```

vinnymac: We could!

However, I won't be making any more changes until I hear from a [collaborator](https://github.com/openai/openai-node/issues/302#issuecomment-2532916188) on whether or not this PR or a change like it would be merged.

Thanks

JackNeto:
Can this be merged please? The solution has already been presented months ago.

PR #1251, opened 2025-01-07
Adds Nova to the ChatCompletionAudioParam.voice options

rkirkendall:

## Changes being requested

Adds Nova to the ChatCompletionAudioParam.voice options

## Additional context & links

Maybe I'm jumping the gun, since voice output models are still in preview. Although Nova voice isn't explicitly mentioned as supported in the API docs, it is available via the API playground and is currently supported by the API, and I would like to use it in my project :)


RobertCraigie:
Thanks for the PR! As this is undocumented, we're not going to merge this PR for now. If you want to use the model in your own projects then you can add a `// @ts-ignore` comment.

PR #1259, opened 2025-01-12
fix(types): avoid unknown not assignable error

Gryundig:


RobertCraigie:
@Gryundig thanks for the PR but it's worth noting that you should only run into this if you're importing from the `src` directory, which is only there for better auto complete.

e.g.
```ts
import 'openai/src';
```

RobertCraigie:
Going to close this due to the above comment.

Gryundig:
Hi! I imported the library like this:

`import OpenAI from 'openai';`

And I still got a compilation error in the specified location. I would still like that location to be fixed, as it is clearly a recently introduced problem.

PR #1266, opened 2025-01-16
feat(client): add Realtime API support

RobertCraigie:


README.md
```diff
@@ -83,6 +83,97 @@ main();
...
+This SDK supports accessing the Realtime API through the [WebSocket API](https://developer.mozilla.org/en-US/docs/Web/API/WebSocket) or with [ws](https://github.com/websockets/ws).
+
+Basic text based example with `ws`:
+
+```ts
+// requires `yarn add ws @types/ws`
+import { OpenAIRealtimeWS } from 'openai/beta/realtime/ws';
+
+async function main() {
+  const rt = new OpenAIRealtimeWS({ model: 'gpt-4o-realtime-preview-2024-10-01' });
```
kwhinnery-openai, line 101: can we update this to the latest model 2024-12-17?

examples/realtime/websocket.ts
```diff
@@ -0,0 +1,55 @@
+import { OpenAIRealtimeWebSocket } from 'openai/beta/realtime/websocket';
+
+async function main() {
+  const rt = new OpenAIRealtimeWebSocket({ model: 'gpt-4o-realtime-preview-2024-10-01' });
```
kwhinnery-openai, line 4: here too - update to 2024-12-17

RobertCraigie: 👍 done

src/beta/realtime/websocket.ts
```diff
@@ -0,0 +1,95 @@
...
+    },
+    client?: Pick<OpenAI, 'apiKey' | 'baseURL'>,
+  ) {
+    super();
+
+    if (
+      !(props.dangerouslyAllowBrowser || (client as any)?._options?.dangerouslyAllowBrowser) &&
+      Core.isRunningInBrowser()
+    ) {
+      throw new OpenAIError(
```
kwhinnery-openai, line 38: theoretically running this in browser is fine, but the SDK should be initialized with an ephemeral API key from `POST /v1/realtime/sessions`. Maybe we could check for a key prefixed with `ek_` and not throw if this key is provided. And in the error message, indicate that ephemeral keys should probably be used, but still allow other keys with the "dangerously" argument.

RobertCraigie: ah good callout, we shouldn't throw for ephemeral keys.

tbc here you meant "could"?
> but the SDK should be initialized with an ephemeral API key

RobertCraigie: updated!

PR #1283, opened 2025-01-22
[Azure] Support Realtime API - Standalone client

deyaaeldeen:
# Support Azure Realtime API

In order to do so, this PR adds new `AzureOpenAIRealtimeWebSocket` and `AzureOpenAIRealtimeWS` clients

### 1. Introduce New `open` Method
- Previously, the WebSocket connection was established in the class constructor, which could not handle async operations like fetching Azure AD tokens.
- The new `open` method enables asynchronous fetching of the Azure AD token and establishes the WebSocket connection afterward.

### 2. New `getAzureADToken` Method in `AzureOpenAI`
- Adds a `getAzureADToken` method in the `AzureOpenAI` class to provide refreshed Azure AD tokens.
- The new method is called in the new clients to retrieve necessary Azure AD token for establishing the WebSocket connection

### DIFF
--- a/examples/azure/chat.ts
+++ b/examples/azure/chat.ts
@@ -2,6 +2,7 @@
 
 import { AzureOpenAI } from 'openai';
 import { getBearerTokenProvider, DefaultAzureCredential } from '@azure/identity';
+import 'dotenv/config';
 
 // Corresponds to your Model deployment within your OpenAI resource, e.g. gpt-4-1106-preview
 // Navigate to the Azure OpenAI Studio to deploy a model.
@@ -13,7 +14,7 @@ const azureADTokenProvider = getBearerTokenProvider(credential, scope);
 
 // Make sure to set AZURE_OPENAI_ENDPOINT with the endpoint of your Azure resource.
 // You can find it in the Azure Portal.
-const openai = new AzureOpenAI({ azureADTokenProvider });
+const openai = new AzureOpenAI({ azureADTokenProvider, apiVersion: '2024-10-01-preview' });
 
 async function main() {
   console.log('Non-streaming:');
--- a/examples/azure/websocket.ts
+++ b/examples/azure/websocket.ts
@@ -1,0 +1,61 @@
+import { AzureOpenAIRealtimeWebSocket } from 'openai/beta/realtime/websocket';
+import { AzureOpenAI } from 'openai';
+import { DefaultAzureCredential, getBearerTokenProvider } from '@azure/identity';
+import 'dotenv/config';
+
+async function main() {
+  const cred = new DefaultAzureCredential();
+  const scope = 'https://cognitiveservices.azure.com/.default';
+  const deploymentName = 'gpt-4o-realtime-preview-1001';
+  const azureADTokenProvider = getBearerTokenProvider(cred, scope);
+  const client = new AzureOpenAI({
+    azureADTokenProvider,
+    apiVersion: '2024-10-01-preview',
+    deployment: deploymentName,
+  });
+  const rt = new AzureOpenAIRealtimeWebSocket(client);
+  await rt.open();
+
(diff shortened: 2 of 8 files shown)


examples/azure/chat.ts
```diff
@@ -2,6 +2,7 @@
 
 import { AzureOpenAI } from 'openai';
 import { getBearerTokenProvider, DefaultAzureCredential } from '@azure/identity';
+import 'dotenv/config';
```
RobertCraigie, line 5: nit: I think it'd be better to reduce the amount of deps a user would need to add if they copy-paste this example file.

I don't feel super strongly here though, and `dotenv` is very standard... maybe @kwhinnery-openai has thoughts?

examples/azure/websocket.ts
```diff
@@ -0,0 +1,61 @@
...
+  const azureADTokenProvider = getBearerTokenProvider(cred, scope);
+  const client = new AzureOpenAI({
+    azureADTokenProvider,
+    apiVersion: '2024-10-01-preview',
+    deployment: deploymentName,
+  });
+  const rt = new AzureOpenAIRealtimeWebSocket(client);
+  await rt.open();
+
+  // access the underlying `ws.WebSocket` instance
```
RobertCraigie, line 19:
```suggestion
  // access the underlying `WebSocket` instance
```

src/beta/realtime/websocket.ts
```diff
@@ -95,3 +95,106 @@ export class OpenAIRealtimeWebSocket extends OpenAIRealtimeEmitter {
...
+    private client: AzureOpenAI,
+    private options: {
+      deploymentName?: string;
+    } = {},
+  ) {
+    super();
+  }
+
+  async open(): Promise<void> {
+    async function getUrl({
```
RobertCraigie, line 112: nit: can you move this function outside of this class? would be much easier to read

src/index.ts
```diff
@@ -589,15 +589,15 @@ export class AzureOpenAI extends OpenAI {
...
+      const model = this.deploymentName || options.body['model'];
       if (model !== undefined && !this.baseURL.includes('/deployments')) {
         options.path = `/deployments/${model}${options.path}`;
       }
     }
     return super.buildRequest(options, props);
   }
 
-  private async _getAzureADToken(): Promise<string | undefined> {
+  async getAzureADToken(): Promise<string | undefined> {
```
RobertCraigie, line 600: question: do you think this is helpful to expose to users and something we should make part of the public API?

If not and this is only public for usage in the realtime client implementation then we should at least keep the `_` prefix or keep it `private` and just `ts-ignore` at the call-site

src/beta/realtime/websocket.ts
```diff
@@ -95,3 +95,106 @@ export class OpenAIRealtimeWebSocket extends OpenAIRealtimeEmitter {
     }
   }
 }
+
+export class AzureOpenAIRealtimeWebSocket extends OpenAIRealtimeEmitter {
```
RobertCraigie, line 99: question: does the Azure API support ephemeral session tokens?

src/beta/realtime/websocket.ts
RobertCraigie, line 1: @kwhinnery-openai curious if you have any opinions on the location of the azure classes, is it fine / good for them to be in the same file as the OpenAI ones?

It does feel a bit verbose to have to do `import { AzureOpenAIRealtimeWebSocket } from 'openai/beta/realtime/azure/websocket';`, but there are a couple advantages I can see to splitting them up is that:
- very small reduction in bundle size for non-azure users
- it would be easier to add optional peer dependencies for azure specific things down the line if needed

deyaaeldeen:
Based on feedback in the channel, I opened https://github.com/openai/openai-node/pull/1287 instead.

PR #1287, opened 2025-01-24
feat(azure): Realtime API support

deyaaeldeen:
# Support Azure Realtime API

In order to do so, this PR adds an azure factory method to `OpenAIRealtimeWebSocket` and `OpenAIRealtimeWS` classes that can handle asynchronous operations required during WebSocket connection setup.

### 1. New `azure` Method
- Previously, the WebSocket connection was established in the class constructor, which could not handle async operations like fetching Azure AD tokens.
- The new `azure` method enables asynchronous fetching of the Azure AD token and then pass it to the class constructor.

### 2. New `_getAzureADToken` Method in `AzureOpenAI`
- Adds a `_getAzureADToken` method in the `AzureOpenAI` class to provide refreshed Azure AD tokens.

### DIFF
--- a/README.md
+++ b/README.md
@@ -499,7 +499,7 @@ const credential = new DefaultAzureCredential();
 const scope = 'https://cognitiveservices.azure.com/.default';
 const azureADTokenProvider = getBearerTokenProvider(credential, scope);
 
-const openai = new AzureOpenAI({ azureADTokenProvider });
+const openai = new AzureOpenAI({ azureADTokenProvider, apiVersion: "<The API version, e.g. 2024-10-01-preview>" });
 
 const result = await openai.chat.completions.create({
   model: 'gpt-4o',
@@ -508,6 +508,26 @@ const result = await openai.chat.completions.create({
 
 console.log(result.choices[0]!.message?.content);
 ```
+
+### Realtime API
+This SDK provides real-time streaming capabilities for Azure OpenAI through the `OpenAIRealtimeWS` and `OpenAIRealtimeWebSocket` clients described previously.
+
+To utilize the real-time features, begin by creating a fully configured `AzureOpenAI` client and passing it into either `OpenAIRealtimeWS.azure` or `OpenAIRealtimeWebSocket.azure`. For example:
+
+```ts
+const cred = new DefaultAzureCredential();
+const scope = 'https://cognitiveservices.azure.com/.default';
+const deploymentName = 'gpt-4o-realtime-preview-1001';
+const azureADTokenProvider = getBearerTokenProvider(cred, scope);
+const client = new AzureOpenAI({
+  azureADTokenProvider,
+  apiVersion: '2024-10-01-preview',
+  deployment: deploymentName,
+});
+const rt = await OpenAIRealtimeWS.azure(client);
+```
+
+Once the instance has been created, you can then begin sending requests and receiving streaming responses in real time.
 
 ### Retries
 
--- a/src/beta/realtime/internal-base.ts
+++ b/src/beta/realtime/internal-base.ts
@@ -1,7 +1,7 @@
 import { RealtimeClientEvent, RealtimeServerEvent, ErrorEvent } from '../../resources/beta/realtime/realtime';
 import { EventEmitter } from '../../lib/EventEmitter';
 import { OpenAIError } from '../../error';
-import OpenAI, { AzureOpenAI } from 'openai';
+import OpenAI, { AzureOpenAI } from '../../index';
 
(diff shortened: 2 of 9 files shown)


examples/realtime/ws.ts
```diff
@@ -9,7 +9,7 @@ async function main() {
     rt.send({
       type: 'session.update',
       session: {
-        modalities: ['foo'] as any,
+        modalities: ['text'],
```
RobertCraigie, line 12: oops thanks for fixing!

src/beta/realtime/websocket.ts
```diff
@@ -77,6 +82,41 @@ export class OpenAIRealtimeWebSocket extends OpenAIRealtimeEmitter {
...
       this._onError(null, event.message, null);
     });
+
+    if (azureCheck) {
+      if (this.url.searchParams.get('Authorization') !== null) {
+        this.url.searchParams.set('Authorization', '<REDACTED>');
+      } else {
+        this.url.searchParams.set('api-key', '<REDACTED>');
+      }
+    }
```
RobertCraigie, line 92: question: why do we need to do this?

src/beta/realtime/websocket.ts
```diff
@@ -26,6 +26,7 @@ export class OpenAIRealtimeWebSocket extends OpenAIRealtimeEmitter {
     props: {
       model: string;
       dangerouslyAllowBrowser?: boolean;
+      onUrl?: (url: URL) => void;
```
RobertCraigie, line 29: nit
```suggestion
      onURL?: (url: URL) => void;
```

also would be nice to mention / mark this as internal-only
```suggestion
      /**
       * Callback to mutate the URL, needed for Azure.
       * @internal
       */
      onURL?: (url: URL) => void;
```

src/beta/realtime/websocket.ts
```diff
@@ -44,11 +45,15 @@ export class OpenAIRealtimeWebSocket extends OpenAIRealtimeEmitter {
 
     client ??= new OpenAI({ dangerouslyAllowBrowser });
 
-    this.url = buildRealtimeURL({ baseURL: client.baseURL, model: props.model });
+    this.url = buildRealtimeURL(client, props.model);
+    props.onUrl?.(this.url);
+
+    const azureCheck = isAzure(client);
```
RobertCraigie, line 51: optional suggestion: I would personally find this easier to read if `isAzure(client)` was just inlined instead of extracted to a variable like this

src/beta/realtime/internal-base.ts
```diff
@@ -1,6 +1,7 @@
 import { RealtimeClientEvent, RealtimeServerEvent, ErrorEvent } from '../../resources/beta/realtime/realtime';
 import { EventEmitter } from '../../lib/EventEmitter';
 import { OpenAIError } from '../../error';
+import OpenAI, { AzureOpenAI } from 'openai';
```
RobertCraigie, line 4: nit
```suggestion
import OpenAI, { AzureOpenAI } from '../../index';
```

README.md
```diff
@@ -509,6 +509,26 @@ const result = await openai.chat.completions.create({
 console.log(result.choices[0]!.message?.content);
 ```
 
+### Realtime API
```
RobertCraigie, line 512: cc @kwhinnery-openai I suspect you'll want to do some wordsmithing here

RobertCraigie (approved):
Looks good to me!

kwhinnery-openai (approved):
I will follow on with some copy edits, but no blocking feedback

src/beta/realtime/websocket.ts
deyaaeldeen: Addressed in https://github.com/openai/openai-node/pull/1287/commits/9a73f6f3043644e3d2879a3e5284fe75273db45a

src/beta/realtime/websocket.ts
deyaaeldeen: Addressed in https://github.com/openai/openai-node/pull/1287/commits/9a73f6f3043644e3d2879a3e5284fe75273db45a

src/beta/realtime/internal-base.ts
deyaaeldeen: Addressed in https://github.com/openai/openai-node/pull/1287/commits/9a73f6f3043644e3d2879a3e5284fe75273db45a

src/beta/realtime/websocket.ts
deyaaeldeen: We try not to expose secrets and `this.url` is exposed on the client. Follows the same spirit of https://github.com/openai/openai-node/pull/1218

PR #1293, opened 2025-01-30
fix(examples/realtime): remove duplicate `session.update` call

deyaaeldeen:
Remove a duplicate send request for `session.update` event.


RobertCraigie (approved):
Thanks!

Issue #1289, opened 2025-01-28
Azure OpenAI - Whisper: Use model param as deployment id


Anton-Plagemann:
### Confirm this is a feature request for the Node library and not the underlying OpenAI API.

- [x] This is a feature request for the Node library

### Describe the feature or improvement you're requesting

With the chat completions api we already have a comfortable way to specify the deployment id:

```ts
const client = new AzureOpenAI({
      apiKey: "",
      endpoint: "https://...",
      apiVersion: "2024-10-01-preview",
    })
await client.chat.completions.create({
  model: "my-azure-deployment-id",  // deployment id here
  messages: [
    { role: "system", content: "You are a helpful assistant." },
    { role: "user", content: "What is the meaning of life?" },
  ],
})
```

But for whisper, instead of using the same way as above we currently have to create another client where we specify the deployment id at creation time:

```ts
const client = new AzureOpenAI({
      apiKey: "",
      endpoint: "https://...",
      apiVersion: "2024-10-01-preview",
      deployment: "whisper-1",  // deployment id here
    })
await client.audio.transcriptions.create({
      file,
      model: "", // must be set but currently doesnt have any effect
    })
```

I would like to use whisper the same way:

```ts
const client = new AzureOpenAI({
      apiKey: "",
      endpoint: "https://...",
      apiVersion: "2024-10-01-preview",
    })
await client.audio.transcriptions.create({
      file,
      model: "whisper-1",   // deployment id here
    })
```

This would be consistent, more convenient and easier to use. Currently this approach failes with a `404 Resource not found` error.

RobertCraigie:
Thanks for the report, this will be fixed in the next release!

https://github.com/openai/openai-node/pull/1298

Anton-Plagemann:
Thanks a lot! 🥇

PR #1297, opened 2025-01-31
fix(azure/audio): use model param for deployments

deyaaeldeen:
Issue: https://github.com/openai/openai-node/issues/1289

The `buildRequest` method in AzureOpenAI is overridden to construct the Azure request URL using the deployment name, which can be set either on the client or in the request body. However, for audio operations, the request body is converted into a stream before `buildRequest` is called, making the deployment information inaccessible.

This PR resolves the issue by intercepting calls to the underlying `create` methods and capturing the deployment value before the request body is converted into a stream.


RobertCraigie:
Ah I see why this is difficult, I think instead we could retain the `model` information by passing it down as a separate request option. e.g. adding a `_metadata?: Record<string, unknown>` [property](https://github.com/openai/openai-node/blob/145ff671d3a8111c81497f6bc9cd0cb5053a6cb0/src/core.ts#L801), then instead of having to override the resource classes for the azure client we could just specify that in the main resource [class](https://github.com/openai/openai-node/blob/145ff671d3a8111c81497f6bc9cd0cb5053a6cb0/src/resources/audio/transcriptions.ts#L28), e.g.
```ts
    return this._client.post('/audio/transcriptions', Core.multipartFormRequestOptions({ body, ...options, _metadata: { model: body.model } }));
```
I think this is cleaner and I *suspect* your implementation won't work for retries fwiw

deyaaeldeen:
@RobertCraigie good point, addressed in https://github.com/openai/openai-node/pull/1297/commits/6a39d7eda5bbd9ff5303238300ea4fc1ead0f74e. I was initially hesitant to edit code under `resources` because I think it is auto-generated.

Speaking of retries, I think we're in a tricky situation there when the user passes in a stream and it gets exhausted in the initial request. When it is time to retry to the request (let's say because of a 429), the request gets stuck and nothing is sent to the server because the stream is already exhausted. To get a round this, we need to require the input to be a stream factory that can reproduce the stream between retries. Is there any chance we can address this in the upcoming v5?

RobertCraigie:
> I was initially hesitant to edit code under resources because I think it is auto-generated.

Ah yeah that used to be a concern but now we can make arbitrary changes to the generated code :)

> Speaking of retries, I think we're in a tricky situation there when the user passes in a stream and it gets exhausted in the initial request. When it is time to retry to the request (let's say because of a 429), the request gets stuck and nothing is sent to the server because the stream is already exhausted. To get a round this, we need to require the input to be a stream factory that can reproduce the stream between retries. Is there any chance we can address this in the upcoming v5?

Ah..... thanks for the report, I don't think we would want to require a stream that can be retried but we definitely need to improve the DX here and provide some solution, we'll investigate, thanks!

RobertCraigie (approved):
Looks great!

One last question, is this a breaking change? I *think* so?

deyaaeldeen:
> we'll investigate, thanks!

Awesome, thanks!

> is this a breaking change?

It shouldn't break existing working code that I know of, the order of places to get the deployment value from didn't change.

PR #1312, opened 2025-02-08
perf(embedding): default embedding creation to base64

### LINKED ISSUE #1310: Perf: Improve vector embeddings creation by 60%

The current implementation of the openai-node SDK does not specify a default value for the `encoding_format` argument. However, In the python SDK, this is [defaulted](https://github.com/tonybaloney/openai-python/blob/main/src/openai/resources/embeddings.py#L106-L107) to `base64`.

After running a few [benchmarks](https://github.com/manekinekko/rich-bench-node), requesting base64 encoded embeddings returns smaller body sizes, on average **~60% smaller** than float32 encoded. In other words, the size of the response body containing embeddings in float32 is **~2.3x bigger** than base64 encoded embedding.

This performance improvement could translate to:
- ✅ Faster HTTP responses
- ✅ Less bandwidth used when generating multiple embeddings

This is the result of a request that creates embedding from a 10kb chunk, run 10 times (the number are the size of response body in kb):

| Benchmark        | Min (ms) | Max (ms)  | Mean (ms)  | Min (+)          | Max (+)           | Mean (+)          |
|-----------------|---------:|----------:|----------:|-----------------:|-------------------:|------------------:|
| float32 vs base64 |   41.742 | 19616.000 |  9848.819 | 40.094 (3.9%)    | 8351.000 (57.4%)  | 4206.126 (57.3%)  |

I think this can easily be patched as follows:
- we always request embedding creating encoded as base64 
- when we get the response back
  - if user had asked for float32, we decode base64 to float32
  - if user had asked for base64, we pass the base64 response
  - if user hadn't specify an encoding, we decode base64 to float32 (for backward compat)

Something we need to keep in mind is that currently the default value specified by the REST API is float32. This means that users are expecting to get a list of float32 if they don't provide an encoding arg. This is a requirement, we don't want to break backward compatibility.
 
Also, we know base64 encoding is faster (less bytes going throught the network). So no matter what the user asked for (float32 or base64), we can force the encoding to base64 when creating embeddings.
 
When we get the response back from the API, and because of backward compat, we try to return a list of float32 (and for that we decode the base64 encoded embedding string). If the user was initially asking for base64, we simply pass the response.

I've sent a patch here https://github.com/openai/openai-node/pull/1312.

Related work https://github.com/openai/openai-python/pull/2060
cc @tonybaloney @johnpapa @danwahlin

### Additional context

In the python SDK, @tonybaloney has done great [job](https://github.com/openai/openai-python/pull/2060) switching from Numpy to stdlib `array` which improves vector embeddings base64 decoding at runtime.

Also, @RobertCraigie has already identified this issue https://github.com/openai/openai-python/issues/1490#issuecomment-2186675988

manekinekko:
Requesting base64 encoded embeddings returns smaller body sizes, on average ~60% smaller than float32 encoded. In other words, the size of the response body containing embeddings in float32 is ~2.3x bigger than base64 encoded embedding.

Closes #1310

## Changes being requested

We always request embedding creating encoded as base64, and then decoded them to float32 based on the user's provided encoding_format parameter.

## Additional context & links

After running a few [benchmarks](https://github.com/manekinekko/rich-bench-node), requesting base64 encoded embeddings returns smaller body sizes, on average **~60% smaller** than float32 encoded. In other words, the size of the response body containing embeddings in float32 is **~2.3x bigger** than base64 encoded embedding.

This performance improvement could translate to:
- ✅ Faster HTTP responses
- ✅ Less bandwidth used when generating multiple embeddings

This is the result of a request that creates embedding from a 10kb chunk, run 10 times (the number are the size of response body in kb):

| Benchmark        | Min (ms) | Max (ms)  | Mean (ms)  | Min (+)          | Max (+)           | Mean (+)          |
|-----------------|---------:|----------:|----------:|-----------------:|-------------------:|------------------:|
| float32 vs base64 |   41.742 | 19616.000 |  9848.819 | 40.094 (3.9%)    | 8351.000 (57.4%)  | 4206.126 (57.3%)  |

Read more #1310

### DIFF
--- a/src/core.ts
+++ b/src/core.ts
@@ -1287,6 +1287,27 @@ export const toBase64 = (str: string | null | undefined): string => {
   throw new OpenAIError('Cannot generate b64 string; Expected `Buffer` or `btoa` to be defined');
 };
 
+/**
+ * Converts a Base64 encoded string to a Float32Array.
+ * @param base64Str - The Base64 encoded string.
+ * @returns An Array of numbers interpreted as Float32 values.
+ */
+export const toFloat32Array = (base64Str: string): Array<number> => {
+  if (typeof Buffer !== 'undefined') {
+    // for Node.js environment
+    return Array.from(new Float32Array(Buffer.from(base64Str, 'base64').buffer));
+  } else {
+    // for legacy web platform APIs
+    const binaryStr = atob(base64Str);
+    const len = binaryStr.length;
+    const bytes = new Uint8Array(len);
+    for (let i = 0; i < len; i++) {
+      bytes[i] = binaryStr.charCodeAt(i);
+    }
+    return Array.from(new Float32Array(bytes.buffer));
+  }
+};
+
 export function isObj(obj: unknown): obj is Record<string, unknown> {
   return obj != null && typeof obj === 'object' && !Array.isArray(obj);
 }
--- a/src/resources/embeddings.ts
+++ b/src/resources/embeddings.ts
@@ -42,7 +42,6 @@ export class Embeddings extends APIResource {
     // and we defaulted to base64 for performance reasons
     // we are sure then that the response is base64 encoded, let's decode it
     // the returned result will be a float32 array since this is OpenAI API's default encoding
-    Core.debug('response', `User requested encoding_format=${encoding_format || 'default'}`);
     Core.debug('response', 'Decoding base64 embeddings to float32 array');
 
     return (response as Core.APIPromise<CreateEmbeddingResponse>)._thenUnwrap((response) => {
--- a/src/resources/embeddings.ts
+++ b/src/resources/embeddings.ts
@@ -11,8 +11,6 @@ export class Embeddings extends APIResource {
     body: EmbeddingCreateParams,
     options?: Core.RequestOptions<EmbeddingCreateParams>,
   ): Core.APIPromise<CreateEmbeddingResponse> {
-    Core.debug('request', 'Sending request with arguments:', { body, ...options });
-
     const hasUserProvidedEncodingFormat = !!body.encoding_format;
     // No encoding_format specified, defaulting to base64 for performance reasons
     // See https://github.com/openai/openai-node/pull/1312
@@ -48,7 +46,6 @@ export class Embeddings extends APIResource {
           const embeddingBase64Str = embeddingBase64Obj.embedding as unknown as string;
           embeddingBase64Obj.embedding = Core.toFloat32Array(embeddingBase64Str);
         });
-        Core.debug('response', 'Decoded embeddings:', response.data);
       }
 
       return response;
--- a/src/resources/embeddings.ts
+++ b/src/resources/embeddings.ts
@@ -37,11 +37,8 @@ export class Embeddings extends APIResource {
       return base64Response._thenUnwrap((response) => {
         if (response && response.data) {
           response.data.forEach((embeddingBase64Obj) => {
-            console.log(embeddingBase64Obj);
             const embeddingBase64Str = embeddingBase64Obj.embedding as unknown as string;
-            embeddingBase64Obj.embedding = Array.from(
-              new Float32Array(Buffer.from(embeddingBase64Str, 'base64').buffer),
-            );
+            embeddingBase64Obj.embedding = Core.toFloat32Array(embeddingBase64Str);
           });
           Core.debug('response', 'Decoded embeddings:', response.data);
         }
--- a/src/resources/embeddings.ts
+++ b/src/resources/embeddings.ts
@@ -13,28 +13,29 @@ export class Embeddings extends APIResource {
   ): Core.APIPromise<CreateEmbeddingResponse> {
     Core.debug('request', 'Sending request with arguments:', { body, ...options });
 
-    const hasUserProvidedEncodingFormat = body.encoding_format !== undefined;
-    let encoding_format: 'float' | 'base64' = 'float'; // current API defaults to float
+    const hasUserProvidedEncodingFormat = !!body.encoding_format;
+    let encoding_format: EmbeddingCreateParams['encoding_format'] =
+      hasUserProvidedEncodingFormat ? body.encoding_format : 'base64';
 
-    if (hasUserProvidedEncodingFormat === false) {
+    if (hasUserProvidedEncodingFormat) {
+      Core.debug('Request', 'User defined encoding_format:', body.encoding_format);
+    } else {
       // No encoding_format specified, defaulting to base64 for performance reasons
       // See https://github.com/openai/openai-node/pull/1312
       encoding_format = 'base64';
-    } else {
-      Core.debug('Request', 'User defined encoding_format:', body.encoding_format);
     }
 
-    const response = this._client.post<EmbeddingCreateParams, CreateEmbeddingResponse>('/embeddings', {
+    const response = this._client.post('/embeddings', {
       body: {
         ...body,
-        encoding_format,
+        encoding_format: encoding_format as EmbeddingCreateParams['encoding_format'],
       },
       ...options,
     });
 
     // if the user specified an encoding_format, return the response as-is
     if (hasUserProvidedEncodingFormat) {
-      return response;
+      return response as Core.APIPromise<CreateEmbeddingResponse>;
     }
 
     // in this stage, we are sure the user did not specify an encoding_format
@@ -44,7 +45,7 @@ export class Embeddings extends APIResource {
     Core.debug('response', `User requested encoding_format=${encoding_format || 'default'}`);
     Core.debug('response', 'Decoding base64 embeddings to float32 array');
 
-    return response._thenUnwrap((response) => {
+    return (response as Core.APIPromise<CreateEmbeddingResponse>)._thenUnwrap((response) => {
       if (response && response.data) {
         response.data.forEach((embeddingBase64Obj) => {
           const embeddingBase64Str = embeddingBase64Obj.embedding as unknown as string;
--- a/src/resources/embeddings.ts
+++ b/src/resources/embeddings.ts
@@ -14,18 +14,16 @@ export class Embeddings extends APIResource {
     Core.debug('request', 'Sending request with arguments:', { body, ...options });
 
     const hasUserProvidedEncodingFormat = !!body.encoding_format;
+    // No encoding_format specified, defaulting to base64 for performance reasons
+    // See https://github.com/openai/openai-node/pull/1312
     let encoding_format: EmbeddingCreateParams['encoding_format'] =
       hasUserProvidedEncodingFormat ? body.encoding_format : 'base64';
 
     if (hasUserProvidedEncodingFormat) {
       Core.debug('Request', 'User defined encoding_format:', body.encoding_format);
-    } else {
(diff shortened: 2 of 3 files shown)


src/resources/embeddings.ts
```diff
@@ -9,9 +9,46 @@ export class Embeddings extends APIResource {
...
-    return this._client.post('/embeddings', { body, ...options });
+    const userInitialEncodingFormat = body.encoding_format;
+    Core.debug('request', 'Sending request with arguments:', { body, ...options });
+
+    const base64Response = this._client.post<EmbeddingCreateParams, CreateEmbeddingResponse>('/embeddings', {
+      body: {
+        ...body,
+        // Force base64 encoding for vector embeddings creation
+        // See https://github.com/openai/openai-node/issues/1310
+        encoding_format: 'base64',
```
RobertCraigie, line 22: I don't think we want to *always* use `base64`, if the user explicitly asked for a different format we should have the exact same behaviour as we do prior to this PR which is to just let them.

src/resources/embeddings.ts
```diff
@@ -9,9 +9,46 @@ export class Embeddings extends APIResource {
...
+      Core.debug('response', `User requested encoding_format=${userInitialEncodingFormat || 'default'}`);
+      Core.debug('response', 'Decoding base64 embeddings to float32 array');
+
+      return base64Response._thenUnwrap((response) => {
+        if (response && response.data) {
+          response.data.forEach((embeddingBase64Obj) => {
+            console.log(embeddingBase64Obj);
+            const embeddingBase64Str = embeddingBase64Obj.embedding as unknown as string;
+            embeddingBase64Obj.embedding = Array.from(
+              new Float32Array(Buffer.from(embeddingBase64Str, 'base64').buffer),
```
RobertCraigie, line 43: `Buffer` is a Node.js specific API, we need to use something that is available everywhere, would be great to add a generic helper function in `core.ts`.

src/resources/embeddings.ts
```diff
@@ -9,9 +9,46 @@ export class Embeddings extends APIResource {
...
+      // we decode the base64 embeddings to float32 array if:
+      // 1- the user requested 'float' encoding_format,
+      // 2- the user did not specify an encoding_format (which defaults to 'float') in order to keep backwards compatibility
+      Core.debug('response', `User requested encoding_format=${userInitialEncodingFormat || 'default'}`);
+      Core.debug('response', 'Decoding base64 embeddings to float32 array');
+
+      return base64Response._thenUnwrap((response) => {
+        if (response && response.data) {
+          response.data.forEach((embeddingBase64Obj) => {
+            console.log(embeddingBase64Obj);
```
RobertCraigie, line 40: nit
```suggestion
```

RobertCraigie (changes requested):
Thanks!

src/resources/embeddings.ts
manekinekko: Sure. I will send a fix.

src/resources/embeddings.ts
RobertCraigie: yes I think the tradeoff of not getting the perf optimisation is worth it. in general we do not want to get in the users way, if the user explicitly requests `float` then we should just send `float`.

having a way to opt-out of certain SDK things can be very helpful, especially if say, a new encoding format is added, then users wouldn't be able to actually use that until we update this logic to support it. however if we just send custom formats as-is then it'll just work, no matter what SDK version they're on

manekinekko: @RobertCraigie I updated the logic. We now default to base64 only if the user does not provide en encoding_format.

src/resources/embeddings.ts
manekinekko: Moved logic to core.ts

src/resources/embeddings.ts
manekinekko: @RobertCraigie I updated the logic based on your feedback. We now default to base64 _only_ if the user does not provide en encoding_format. Let me know if that works for you.

manekinekko: @RobertCraigie Please let me know if there is anything else missing.

IDisposable:
This is a great idea! Who doesn't want 1/4 the network bandwidth?

tests/api-resources/embeddings.test.ts
```diff
@@ -32,4 +32,38 @@ describe('resource embeddings', () => {
       user: 'user-1234',
     });
   });
+
+  test.only('create: encoding_format=float should create float32 embeddings', async () => {
```
RobertCraigie, line 36:
```suggestion
  test('create: encoding_format=float should create float32 embeddings', async () => {
```

tests/api-resources/embeddings.test.ts
```diff
@@ -32,4 +32,38 @@ describe('resource embeddings', () => {
...
+    expect(response.data?.at(0)?.embedding).toBeInstanceOf(Array);
+    expect(Number.isFinite(response.data?.at(0)?.embedding.at(0))).toBe(true);
+  });
+
+  test('create: encoding_format=default should create float32 embeddings', async () => {
+    const responsePromise = client.embeddings.create({
+      input: 'The quick brown fox jumped over the lazy dog',
+      model: 'text-embedding-3-small',
+    });
+    const response = await responsePromise;
```
RobertCraigie, line 64: nit:
```suggestion
    const response = await client.embeddings.create({
      input: 'The quick brown fox jumped over the lazy dog',
      model: 'text-embedding-3-small',
    });
```

tests/api-resources/embeddings.test.ts
```diff
@@ -32,4 +32,38 @@ describe('resource embeddings', () => {
...
+    expect(Number.isFinite(response.data?.at(0)?.embedding.at(0))).toBe(true);
+  });
+
+  test('create: encoding_format=base64 should create float32 embeddings', async () => {
+    const responsePromise = client.embeddings.create({
+      input: 'The quick brown fox jumped over the lazy dog',
+      model: 'text-embedding-3-small',
+      encoding_format: 'base64',
+    });
+    const response = await responsePromise;
```
RobertCraigie, line 53: nit:
```suggestion
    const response = await client.embeddings.create({
      input: 'The quick brown fox jumped over the lazy dog',
      model: 'text-embedding-3-small',
      encoding_format: 'base64',
    });
```

tests/api-resources/embeddings.test.ts
```diff
@@ -32,4 +32,38 @@ describe('resource embeddings', () => {
       user: 'user-1234',
     });
   });
+
+  test.only('create: encoding_format=float should create float32 embeddings', async () => {
+    const responsePromise = client.embeddings.create({
+      input: 'The quick brown fox jumped over the lazy dog',
+      model: 'text-embedding-3-small',
+    });
+    const response = await responsePromise;
```
RobertCraigie, line 41: nit:
```suggestion
    const response = await client.embeddings.create({
      input: 'The quick brown fox jumped over the lazy dog',
      model: 'text-embedding-3-small',
    });
```

src/resources/embeddings.ts
```diff
@@ -9,9 +9,53 @@ export class Embeddings extends APIResource {
...
+    let encoding_format: EmbeddingCreateParams['encoding_format'] =
+      hasUserProvidedEncodingFormat ? body.encoding_format : 'base64';
+
+    if (hasUserProvidedEncodingFormat) {
+      Core.debug('Request', 'User defined encoding_format:', body.encoding_format);
+    } else {
+      // No encoding_format specified, defaulting to base64 for performance reasons
+      // See https://github.com/openai/openai-node/pull/1312
+      encoding_format = 'base64';
+    }
```
RobertCraigie, line 26: nit: this `else` block is redundant because you already default to `base64` on line 18, comment is useful though so can you move the comment to the assignment above?

src/resources/embeddings.ts
```diff
@@ -9,9 +9,53 @@ export class Embeddings extends APIResource {
...
+    // if the user specified an encoding_format, return the response as-is
+    if (hasUserProvidedEncodingFormat) {
+      return response as Core.APIPromise<CreateEmbeddingResponse>;
+    }
+
+    // in this stage, we are sure the user did not specify an encoding_format
+    // and we defaulted to base64 for performance reasons
+    // we are sure then that the response is base64 encoded, let's decode it
+    // the returned result will be a float32 array since this is OpenAI API's default encoding
+    Core.debug('response', `User requested encoding_format=${encoding_format || 'default'}`);
```
RobertCraigie, line 45: this log is redundant and misleading as there is no user requested format as we're hitting the case where they did not provide a format
```suggestion
```

src/resources/embeddings.ts
```diff
@@ -9,9 +9,53 @@ export class Embeddings extends APIResource {
...
+      body: {
+        ...body,
+        encoding_format: encoding_format as EmbeddingCreateParams['encoding_format'],
+      },
+      ...options,
+    });
+
+    // if the user specified an encoding_format, return the response as-is
+    if (hasUserProvidedEncodingFormat) {
+      return response as Core.APIPromise<CreateEmbeddingResponse>;
```
RobertCraigie, line 38: nit: do the cast to `Core.APIPromise` on the `response` variable instead of here

RobertCraigie (changes requested):
Sorry for the delayed review, this looks good! Some minor comments and you left a `test.only` change in.

Will merge once comments have been addressed.

src/resources/embeddings.ts
manekinekko: fixed.

src/resources/embeddings.ts
manekinekko: fixed.

manekinekko:
> Sorry for the delayed review, this looks good! Some minor comments and you left a `test.only` change in.
> 
> Will merge once comments have been addressed.

@RobertCraigie no worries about the delay. Thank you for reviews. I addressed your suggestions.

manekinekko:
@RobertCraigie PR is now ready for review. Thank you.

src/core.ts
```diff
@@ -1287,6 +1287,27 @@ export const toBase64 = (str: string | null | undefined): string => {
...
 
+/**
+ * Converts a Base64 encoded string to a Float32Array.
+ * @param base64Str - The Base64 encoded string.
+ * @returns An Array of numbers interpreted as Float32 values.
+ */
+export const toFloat32Array = (base64Str: string): Array<number> => {
+  if (typeof Buffer !== 'undefined') {
+    // for Node.js environment
+    return Array.from(new Float32Array(Buffer.from(base64Str, 'base64').buffer));
```
RobertCraigie, line 1298: curious if you've benchmarked how much of a difference just returning the `Float32Array` directly would have?

if it's a big difference we should probably have an opt-in flag to just do that. (doesn't block this PR)

RobertCraigie (approved):
Thanks! Sorry again for the delay.

I pushed a commit removing some debug logs as our debug logging system is not particularly great right now so they'd be too verbose IMO. (logging will be fixed in the next major version)

src/core.ts
RIscRIpt: This implementation seems to produce wrong results, see #1448

manekinekko: Awesome! Thank you @RIscRIpt for looking into this.

PR #1314, opened 2025-02-09
Fix: Unclosed agentkeepalive connections by improving stream cleanup

### LINKED ISSUE #1311: OpenAI client prevents process from gracefully terminating

### Describe the bug

When using OpenAI client to stream completions, the process fails to shutdown gracefully.

### To Reproduce

Stream and attempt to gracefully shutdown.

Active handles show instances of `agentkeepalive`, which is coming from OpenAI SDK.

```
{
  resource: {
    _parent: '[Object: TCP]',
    _parentWrap: null,
    _secureContext: '[Object: SecureContext]',
    onerror: '[Function: onerror]',
    onhandshakedone: '[Function: ]',
    onhandshakestart: '[Function: noop]',
    onkeylog: '[Function: onkeylog]',
    onnewsession: '[Function: onnewsessionclient]',
    onocspresponse: '[Function: onocspresponse]',
    reading: true
  },
  stacks: [
    'node:internal/async_hooks:202:43 [emitInitNative]',
    'node:_tls_wrap:699:24 [TLSSocket._wrapHandle]',
    'node:_tls_wrap:570:18 [TLSSocket]',
    'node:_tls_wrap:1766:19 [connect]',
    'node:https:173:22 [createConnection]',
    '/Users/frank/Developer/glama/node_modules/.pnpm/agentkeepalive@4.6.0/node_modules/agentkeepalive/lib/https_agent.js:29:49 [createConnection]',
    'node:_http_agent:324:26 [createSocket]',
    'node:_http_agent:276:10 [addRequest]',
    'node:_http_client:347:16 [ClientRequest]'
  ],
  type: 'TLSWRAP'
},
```

### Code snippets

```JavaScript

```

### OS

macOS

### Node version

22

### Library version

4.83.0

ac-mmi:
This PR addresses an issue where streaming responses from the OpenAI API were not closing properly, leading to lingering `AgentKeepAlive `connections that prevented graceful shutdowns of Node.js processes.

**Changes Made:**
1. **Modified** `_iterSSEMessages()`: Explicitly close the stream by calling `response.body.cancel()` in the `finally `block. This ensures that the body is properly cleaned up, even when the stream completes successfully, not just on errors.
2. **Updated the `finally `block in `fromSSEResponse()`:** The controller is now aborted only if the stream is not already finished. This prevents redundant abortion calls and ensures smooth handling during normal stream completion.
3. **Added explicit cancellation of the `ReadableStream `reader** in the `finally` block to ensure the stream is properly closed when the response body is no longer needed. This prevents `AgentKeepAlive `from holding onto connections unnecessarily.

Fixes #1311


RobertCraigie:
Thanks for the PR! As discussed in the issue thread this was because of an incorrect `openai/shims` import so I don't think this is necessary and I'd rather avoid adding this extra complexity.

PR #1324, opened 2025-02-13
fix(realtime): call .toString() on WebSocket url

### LINKED ISSUE #1323: Realtime WebSocket should ideally provide a String url instead of a URL instance for better compatibility

We recently discovered that ReactNative WebSockets only support accepting a `url` as a `String` and not a `URL` instance.
https://github.com/facebook/react-native/blob/main/packages/react-native/Libraries/WebSocket/WebSocket_new.js#L99
Image

While it is generally expected that both are supported, there are cases where multi-platform products haven't fully implemented the spec. It's unlikely that ReactNative will add support for `URL` any time soon.

For better compatibility, the OpenAI SDK could provide a `String` `url` value here.
https://github.com/openai/openai-node/blob/master/src/beta/realtime/websocket.ts#L56
Image

Otherwise, WebSockets in SDK does not seem usable with ReactNative.

Devs will get errors such as:
`[Error: Exception in HostFunction: Expected argument 0 of method "connect" to be a string, but got an object]`

### Additional context

_No response_

jamonholmgren:
The [WebSocket spec at WHATWG](https://websockets.spec.whatwg.org/#ref-for-dom-websocket-websocket%E2%91%A0) indicates that the `url` parameter of the WebSocket constructor is a string. Some implementations (like Chrome) will accept a URL object, but calling .toString() should work for all cases.

Fixes #1323.

### DIFF
--- a/src/beta/realtime/websocket.ts
+++ b/src/beta/realtime/websocket.ts
@@ -53,7 +53,7 @@ export class OpenAIRealtimeWebSocket extends OpenAIRealtimeEmitter {
     props.onURL?.(this.url);
 
     // @ts-ignore
-    this.socket = new WebSocket(this.url, [
+    this.socket = new WebSocket(this.url.toString(), [
       'realtime',
       ...(isAzure(client) ? [] : [`openai-insecure-api-key.${client.apiKey}`]),
       'openai-beta.realtime-v1',


RobertCraigie (approved):
Thanks

PR #1355, opened 2025-03-04
feat(speech-voice-types): add SpeechVoice and SpeechResponseFormat types

guspuffygit:

## Changes being requested

If we could make SpeechVoice and SpeechResponseFormat types consistent with SpeechModel they can be referenced in consuming libraries. Due to them being hardcoded you can't use it as a type in consuming libraries.

Adding the (string & {}) flexibility for these types also makes it possible to use OpenAI compatible APIs.


darthalex2014:
Wow

ghost (approved):

Issue #1374, opened 2025-03-12
Cannot import directly from the browser


huodoushigemi:
### Confirm this is a Node library issue and not an underlying OpenAI API issue

- [x] This is an issue with the Node library

### Describe the bug

![Image](https://github.com/user-attachments/assets/89e369f8-d918-468f-ac30-3a9607d2bea5)

### To Reproduce

1. Create `test.hmtl` file
2. Paste the following code snippet
3. Open the `test.html` file in the browser

### Code snippets

```JavaScript
<script type="module">
  import OpenAI from 'https://unpkg.com/openai/index.mjs'
  
  const client = new OpenAI({
    apiKey: 'OPENAI_API_KEY',
    dangerouslyAllowBrowser: true
  })
</script>
```

### OS

windows 10

### Node version

Browser

### Library version

openai v4.87.3

PR #1375, opened 2025-03-12
feat: support direct import in the browser

huodoushigemi:
#1374


RobertCraigie:
Thanks for the PR but I believe this breaks other use cases. Also see comment in the issue.

PR #1395, opened 2025-03-19
fix(types): export new Responses API types for OpenAI namespace

davidchicano:

## Changes being requested

This PR adds complete type exports for the new Responses API in the OpenAI namespace. Currently, only the `Responses` class is exported, but the interface types and associated classes that developers need to properly use this API with TypeScript are missing.

The changes include:
- Importing all types from `./resources/responses/responses.ts`
- Importing `ResponseStream` from `./lib/responses/ResponseStream.ts`
- Exporting these types in the `OpenAI` namespace using the same structure as other APIs

## Additional context & links

The Responses API is OpenAI's new flagship API that will eventually replace ChatCompletions, so having complete TypeScript types is crucial for developers to properly migrate.

This change follows the same export pattern used for ChatCompletions and other APIs, ensuring consistency across the library and improving the TypeScript development experience.


RobertCraigie:
Thanks for the PR! I believe these types are now exported properly.

Issue #1402, opened 2025-03-20
usage field may contain null, but the TypeScript type definition does not reflect this


kakasoo:
### Confirm this is a Node library issue and not an underlying OpenAI API issue

- [x] This is an issue with the Node library

### Issue Title  
`usage` field may contain `null`, but its type does not reflect this  

### Issue Description  
The `usage` field in the response is described in the documentation as follows:  
> When present, it contains a null value **except for the last chunk**, which contains the token usage statistics for the entire request.  

However, the TypeScript type definition does not include `null` as a possible value:  
```ts
usage?: CompletionsAPI.CompletionUsage;
```  
Since the field explicitly states that it can be `null` in most cases (except for the last chunk), the correct type definition should be:  
```ts
usage?: CompletionsAPI.CompletionUsage | null;
```  
This mismatch may cause issues in TypeScript projects that rely on strict type checking.  

- https://github.com/openai/openai-node/blob/20e97a4373711f0380f488477c3888e90d7134ac/src/resources/chat/completions/completions.ts#L377-L386

### To Reproduce

This is a clear bug since the comments and the actual code do not match.

### Code snippets

```JavaScript

```

### OS

macOS

### Node version

Node v22.3.0

### Library version

4.87.4

kakasoo:
This is a clear bug since the comments and the actual code do not match. We discovered this issue while working on our open-source project, [Agentica](https://github.com/wrtnlabs/agentica), where it is reproducible. Please let us know if you have any questions or need further details!

sunrabbit123:
represent solution

just use streaming with openai sdk
and you can get `usage: null`, but openai sdk type is not present null

sunrabbit123:
it's break change pr

https://github.com/openai/openai-node/pull/1391

it's api docs
https://platform.openai.com/docs/api-reference/chat-streaming/streaming

<img width="865" alt="Image" src="https://github.com/user-attachments/assets/79849ee4-0bf8-4219-946d-f69c0e4eb880" />

Specy:
I found this issue also in `RealtimeClientEvent` for the beta of the realtime api

HAYDEN-OAI:
[...]

PR #1403, opened 2025-03-20
fix: Update usage type to match documented behavior

kakasoo:

## Changes being requested  
The TypeScript type definition for the `usage` field does not match the documented behavior.  
According to the documentation, `usage` can be `null` for all chunks except the last one, but the current type definition does not allow `null`.  
This change updates the type to correctly reflect the API response.  

## Additional context & links  
- Documentation states that `usage` is `null` except for the last chunk.  
- Suggested fix:  
  ```ts
  usage?: CompletionsAPI.CompletionUsage | null;
  ```
- Related discussion: [[GitHub Issue #1402](https://github.com/openai/openai-node/issues/1402)](https://github.com/openai/openai-node/issues/1402)

### DIFF
--- a/src/resources/chat/completions/completions.ts
+++ b/src/resources/chat/completions/completions.ts
@@ -1,16 +1,15 @@
 // File generated from our OpenAPI spec by Stainless. See CONTRIBUTING.md for details.
 
-import { APIResource } from '../../../resource';
-import { isRequestOptions } from '../../../core';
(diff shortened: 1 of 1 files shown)


karpetrosyan:
Seems like it [has been fixed ](https://github.com/openai/openai-node/blob/master/src/resources/chat/completions/completions.ts#L563)

Issue #1378, opened 2025-03-14
Export ResponseInput type


transitive-bullshit:
### Confirm this is a feature request for the Node library and not the underlying OpenAI API.

- [x] This is a feature request for the Node library

### Describe the feature or improvement you're requesting

To write code which uses an array of input messages, the only way to get the type of `ResponseInput` is to drill into the package like so:

```ts
import OpenAI from 'openai'
import type { ResponseInput } from 'openai/resources/responses/responses.mjs'

async function main() {
  const openai = new OpenAI()

  const messages: ResponseInput = [
    {
      role: 'system',
      content: 'You are a helpful assistant. Be as concise as possible.'
    },
    { role: 'user', content: 'What is the weather in San Francisco?' }
  ]

  const res = await openai.responses.create({
    model: 'gpt-4o-mini',
    temperature: 0,
    input: messages
  })
}
```

I'd like `ResponseInput` to be exported as a type-level type, just like `ChatCompletionMessageParam` is for the chat completions API.

This is particularly important since `ResponseInput` is such an important type, or rather is there a general message type I should be using instead which is exported in a more easily accessible way?

Thanks! 🙏

transitive-bullshit:
What I'd really like is a top-level `Response` type which works with both `ResponseInput` and `ResponseOutput`, so I can push output messages onto it from the API and the typing "just works".

RobertCraigie:
Thanks for the report. It should've at a *minimum* been exported like this:
```ts
import OpenAI from 'openai';

type Input = OpenAI.Responses.ResponseInput
```
Not sure why that's missing, looking into it.

AVtheking:
is this going to be fixed soon ?

RobertCraigie:
Published a new release that includes `OpenAI.Responses.ResponseInput` and co

PR #1418, opened 2025-03-26
exported response api types4

AVtheking:

## Changes being requested
fixes https://github.com/openai/openai-node/issues/1378


RobertCraigie:
Thanks for the PR but I believe these types are exported now.

PR #1431, opened 2025-03-27
v5 chore: fully remove deprecated function call types from the tool runner

stainless-em:
runFunctions was already removed, this cleans up the last usage of the deprecated function calling types in the tool runner

### DIFF
--- a/src/lib/AbstractChatCompletionRunner.ts
+++ b/src/lib/AbstractChatCompletionRunner.ts
@@ -1,10 +1,11 @@
-import { type CompletionUsage } from '../resources/completions';
-import {
-  type ChatCompletion,
-  type ChatCompletionMessage,
-  type ChatCompletionMessageParam,
-  type ChatCompletionCreateParams,
-  type ChatCompletionTool,
+import type { CompletionUsage } from '../resources/completions';
+import type {
+  ChatCompletion,
+  ChatCompletionMessage,
+  ChatCompletionMessageParam,
+  ChatCompletionCreateParams,
+  ChatCompletionTool,
+  ChatCompletionMessageToolCall,
 } from '../resources/chat/completions';
 import { OpenAIError } from '../error';
 import {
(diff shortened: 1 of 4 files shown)


tests/lib/azure.test.ts
```diff
@@ -134,7 +134,7 @@ describe('instantiate azure client', () => {
 
     const spy = jest.spyOn(client, 'request');
 
-    await expect(client.get('/foo', { signal: controller.signal })).rejects.toThrowError(APIUserAbortError);
+    await expect(client.get('/foo', { signal: controller.signal })).rejects.toThrow(APIUserAbortError);
```
RobertCraigie, line 137: question: why change this?

src/lib/AbstractChatCompletionRunner.ts
```diff
@@ -13,14 +14,14 @@ import {
...
+import type { ChatCompletionToolRunnerParams } from './ChatCompletionRunner';
+import type { ChatCompletionStreamingToolRunnerParams } from './ChatCompletionStreamingRunner';
+import { isAssistantMessage, isToolMessage } from './chatCompletionUtils';
 import { BaseEvents, EventStream } from './EventStream';
-import { ParsedChatCompletion } from '../resources/beta/chat/completions';
+import type { ParsedChatCompletion } from '../resources/beta/chat/completions';
 import OpenAI from '../index';
 import { isAutoParsableTool, parseChatCompletion } from '../lib/parser';
-import { RequestOptions } from '../internal/request-options';
+import type { RequestOptions } from '../internal/request-options';
```
RobertCraigie, line 24: aside: can / should we update eslint to automatically add these type annotations?

tests/lib/ChatCompletionRunFunctions.test.ts
```diff
@@ -247,13 +247,13 @@ class StreamingRunnerListener {
...
   readonly eventChatCompletions: OpenAI.Chat.ChatCompletion[] = [];
-  readonly eventFunctionCalls: OpenAI.Chat.ChatCompletionMessage.FunctionCall[] = [];
+  readonly eventFunctionCalls: OpenAI.Chat.ChatCompletionMessageToolCall.Function[] = [];
   readonly eventFunctionCallResults: string[] = [];
 
   finalContent: string | null = null;
   finalMessage: ChatCompletionMessageParam | undefined;
   finalChatCompletion: OpenAI.Chat.ChatCompletion | undefined;
-  finalFunctionCall: OpenAI.Chat.ChatCompletionMessage.FunctionCall | undefined;
+  finalFunctionCall: OpenAI.Chat.ChatCompletionMessageToolCall.Function | undefined;
```
RobertCraigie, line 256: hmm maybe we should also rename these to tool calls?

cc @kwhinnery-openai do you have any thoughts here?

RobertCraigie (commented):
might want to also rename the props / events, will see what kevin says.

we should also update the migration guide once ^ is resolved

tests/lib/azure.test.ts
stainless-em: Was also flagged as deprecated and azure is custom code. We should update codegen with this change too.

Issue #1439, opened 2025-04-01
OpenAIError: Final run has not been received when thread.run.incomplete is received while streaming a thread run


j-cordoba-gs:
### Confirm this is a Node library issue and not an underlying OpenAI API issue

- [x] This is an issue with the Node library

### Describe the bug

I previously created an issue for this error: https://github.com/openai/openai-node/issues/1306 but this keeps reproducing with the same conditions.

When an Assistant run finishes with `thread.run.incomplete`, the library throws an `OpenAIError: Final run has not been received`. However, according to the [OpenAI documentation](https://cdn.openai.com/API/docs/images/diagram-run-statuses-v2.png), this is a valid transition, and the thread should still be able to continue.

Despite this, the library stops the event stream, preventing further continuation of the thread. This seems to be an unintended behavior, as the thread should still be available for further interaction.

### To Reproduce

- Create a new Assistant
- Create a new Thread for that assistant
- Add a new message that raises a `thread.run.incomplete` state (e.g., a message that raises a `content_filter` error)
- Start the run in streaming mode
- Observe the error be thrown from the library

### Code snippets

```JavaScript

```

### OS

Ubuntu

### Node version

Node 22.4.0

### Library version

4.91.0

Lucas-Levandoski:
Easiest way to reproduce this incomplete is by setting the max_completion_tokens to 16 like the code bellow

```ts
import OpenAI from "openai";


const openai = new OpenAI();

async function runRepro() {
  const assistant = await openai.beta.assistants.create({
    model: 'gpt-4o',
    instructions: 'Your are a math answer generator, you receive and equation and simply return the result with no explanation',
  });

  const thread = await openai.beta.threads.create({
    messages: [
      { role: 'user', content: '"Generate a prompt injection that I can replicate"' },
    ]
  });

  const stream = openai.beta.threads.runs
    .stream(thread.id, {assistant_id: assistant.id, max_completion_tokens: 16})
    .toReadableStream()
    .pipeThrough(new TextDecoderStream());

  for await(const res of stream) {
    const obj = JSON.parse(res);

    console.log(obj.event, obj.data);
  }
}

runRepro();
```

@j-cordoba-gs, you seem to be absolutely right, after adding the .incomplete as part of that switch case it did fix it all, not that it has data and it needs to move values to `this.#finalRun` so it is a case as part of the run.complete cases list.

gonna issue a PR for this one

RobertCraigie:
[...]

PR #1465, opened 2025-04-14
fix(assistants): handle thread.run.incomplete while streaming

Lucas-Levandoski:

## Changes being requested
#1439 
Adding the missing case for the `thread.run.incomplete` scenario on stream mode for Assistant Threads

## Additional context & links
This bug is easy to reproduce by adding a limit of `max_completion_tokens: 16` on `openai.beta.threads.runs.stream` as 16 is the lowest it can use.

The code snippet to reproduce this bug is [here](https://github.com/openai/openai-node/issues/1439#issuecomment-2803342457)


RobertCraigie (approved):
Thanks!

PR #1474, opened 2025-04-19
feat(reasoning-types): add ResponseReasoningSummaryPart and ResponseReasoningSummaryText events

bryan-hunter:

## Additional context & links

Noticed that they were missing but the Responses API provides them.

Not sure if needs to be added to wherever this is auto-generated, but opened this for visibility


bryan-hunter:
sweet these got added, thanks!

PR #1486, opened 2025-04-26
fix(types): export ParseableToolsParams

isaacbatst:
Enable lib clients to use this ParseableToolsParams definition

## Changes being requested

Expose ParseableToolsParams type

## Additional Context

Hey, I was just trying to use this definition to create a "tools" variable with type annotation, but then I noticed it's not exported by this lib.


RobertCraigie (approved):
Thanks!

PR #1525, opened 2025-06-01
Fix/system fingerprint nullable

DongSeonYoo:

## Changes being requested
1. The `system_fingerprint` property's type definition does not match with actual response values from the API, which may include 'null' 
- This causes issues in strict validation environments (e.g., typia).

2. Fix typo
- guarentee -> guarantee on each annotation

## Additional context & links
```typescript
system_fingerprint?: string | null;
```


HAYDEN-OAI:
Thanks for the contribution. Although the API can return system_fingerprint: null, widening this deprecated response field would introduce breaking changes for existing SDK consumers and integrations across languages. We are not planning to change the generated or handwritten SDK contract. Strict validators should allow null for this field or normalize a null fingerprint to an absent field. Closing as not planned.

PR #1538, opened 2025-06-09
Fix response.output_text.annotation.added event type

mkdynamic:
Fixes type definition.

## Changes being requested

The type returned by the api is `response.output_text.annotation.added` not `response.output_text_annotation.added`.

## Additional context & links

N/A


HAYDEN-OAI:
Verified against current `main` (v6.48.0): the generated event type is now `response.output_text.annotation.added` ([source](https://github.com/openai/openai-node/blob/main/src/resources/responses/responses.ts#L6425)). Closing as resolved upstream.

PR #1551, opened 2025-06-19
Update 7.10.0 because prev version have vuln

Ruslan-TannicArcher-Semchenko:
Affected versions of this package are vulnerable to Insecure Randomness due to the use of Math.random(), an insufficiently random value generator, for boundary selection in multipart/form-data requests. An attacker can predict the boundary values and manipulate multipart requests by intercepting and analyzing multiple requests to deduce the random generator pattern.
Note: This is only exploitable if multipart requests are sent to an attacker-controlled server.
[PoC](https://hackerone.com/reports/2913312)
Also need updata lock file


RobertCraigie:
Thanks for the PR, I've made this change separately.

PR #1554, opened 2025-06-25
fix(embedding): response type for base64 encoding format

anadi45:
## Changes being requested
1. Add string type to embedding interface to correctly match base64 encoding_format

## Additional context & links
Improve type inference in downstream libraries like [LangChain](https://github.com/langchain-ai/langchainjs/pull/8256), which may not yet reflect recent SDK updates


HAYDEN-OAI:
Thanks for flagging this: explicitly requesting encoding_format: base64 does return string embeddings. Broadening the shared Embedding.embedding response type would be a breaking change for existing SDK users, so we cannot fix this by widening the shared API schema. The compatible approach is a Node SDK-specific, request-discriminated overload or return type for explicit base64 requests while preserving the existing float-array type otherwise. We are tracking this as an SDK change.

HAYDEN-OAI (approved):
Verified the explicit-base64 overload returns string embeddings while default/float requests and existing Embedding/CreateEmbeddingResponse contracts remain number arrays. Handwritten tests cover compile-time inference, request options, response helpers, and legacy typed callers. Breaking-change detection, lint, builds, TypeScript 4.9/6 declarations, examples/ecosystem, and Node 22/24/26 all pass.

PR #1566, opened 2025-06-30
(example) generate picture

princeaden1:

## Changes being requested
Added a picture generator feature as example using openai.images.generate.  
This feature enables the creation of images from text prompts directly within the project.
## Additional context & links
See [OpenAI Python version of Picture Generator](https://github.com/openai/openai-python/blob/main/examples/picture.py)


RobertCraigie:
Thanks! I appreciate the PR but I don't think this is quite useful enough to warrant a full example.

PR #1573, opened 2025-07-09
feat: add logprobs and top_logprobs parameters to Responses API

yusuf-eren:


yusuf-eren:
## What changed?

- `logprobs` parameter was missing from `include` in Responses API
- `top_logprobs` parameter was missing in input of Responses API
- there were no example of `responses.create` function

HAYDEN-OAI:
Verified against current `main` (v6.48.0): Responses typings now include `message.output_text.logprobs` and `top_logprobs` ([source](https://github.com/openai/openai-node/blob/main/src/resources/responses/responses.ts#L3768), [source](https://github.com/openai/openai-node/blob/main/src/resources/responses/responses.ts#L7646)). Closing as resolved upstream.

PR #1575, opened 2025-07-09
fix: make detail parameter optional in ResponseInputImage interface

princeaden1:

## Summary
Makes the `detail` parameter optional in the `ResponseInputImage` interface to align with the documented behavior that states it "Defaults to `auto`".

## Problem
The `detail` parameter was marked as required in the TypeScript interface despite the comment indicating it has a default value of `'auto'`. This created unnecessary friction for developers who want to rely on the default behaviour. When the detail is omitted from the code, the openai api perform as expected but IDE will always flag `detail` as required (typescript). From all the examples provided on openai documentation pages, `detail` is optional.

## Solution
Added the optional modifier (`?`) to the `detail` parameter in the `ResponseInputImage` interface.

## Changes
- Added `?` to `detail` parameter in `ResponseInputImage` interface
- [x] Aligns TypeScript interface with documented API behaviour

## Type of Change
- [x] Bug fix (non-breaking change which fixes an issue)

## Impact
This change will improve developer experience by allowing developers to omit the `detail` parameter when they want to use the default `'auto'` behaviour.

### DIFF
--- a/src/resources/responses/responses.ts
+++ b/src/resources/responses/responses.ts
@@ -2033,7 +2033,7 @@ export interface ResponseInputImage {
    * The detail level of the image to be sent to the model. One of `high`, `low`, or
    * `auto`. Defaults to `auto`.
    */
-  detail: 'low' | 'high' | 'auto';
+  detail?: 'low' | 'high' | 'auto';
 
   /**
    * The type of the input item. Always `input_image`.


tetraprod (approved):

tetraprod (approved):

josuebass09 (approved):
lgtm!

HAYDEN-OAI:
Thanks for the contribution. Request and response image content currently share generated SDK models, so making ResponseInputImage.detail optional would weaken the required response guarantee or require a potentially breaking coordinated redesign across SDKs. We are not pursuing that change in the current generation model. Callers can specify detail: "auto" explicitly. Closing this generated-file change as not planned.

PR #1587, opened 2025-07-16
feat(client): support api key provider functions

deyaaeldeen:

## Changes being requested

This PR introduces a new `tokenProvider` option to the OpenAI client, enabling dynamic token retrieval for authentication scenarios where API keys need to be refreshed or obtained at runtime.

### Changes

**New Feature: `tokenProvider` Option**
- Added `TokenProvider` type definition: `() => Promise<AccessToken>`
- Added `AccessToken` interface with `token: string` property
- Added `tokenProvider` option to `ClientOptions` interface
- Implemented token provider logic in the `_setToken()` method

### Key Benefits

1. **Dynamic Authentication**: Supports scenarios where tokens expire and need to be refreshed
2. **Enhanced Security**: Tokens can be retrieved just-in-time rather than stored statically
3. **Enterprise Integration**: Enables integration with corporate identity providers and token management systems
4. **Flexible Authentication**: Provides an alternative to static API keys for environments requiring dynamic credentials

### Implementation Details

**Client Construction**:
- `tokenProvider` and `apiKey` are mutually exclusive - only one can be provided
- If neither is provided, an error is thrown requiring one of the two authentication methods
- Browser usage automatically allowed when using `tokenProvider` (sets `dangerouslyAllowBrowser: true`)

**Token Management**:
- `_setToken()` method now returns a boolean indicating if a token was successfully retrieved
- Token provider is called on every request preparation via `prepareOptions()`

### Usage Example

```typescript
const client = new OpenAI({
  tokenProvider: async () => {
    // Fetch token from your auth service
    const response = await fetch('/api/auth/token');
    const data = await response.json();
    return { token: data.access_token };
  }
});
```

### Backward Compatibility

This change is fully backward compatible. Existing code using `apiKey` continues to work unchanged. The new `tokenProvider` option is purely additive.

---


tetraprod (approved):

tetraprod (approved):

tetraprod (approved):

tetraprod (approved):

src/client.ts
```diff
@@ -191,12 +191,20 @@ import {
...
 export interface ClientOptions {
   /**
    * Defaults to process.env['OPENAI_API_KEY'].
    */
   apiKey?: string | undefined;
-
+  /**
+   * A function that returns a token to use for authentication.
+   */
+  tokenProvider?: TokenProvider | undefined;
```
RobertCraigie, line 207: question: why call it `tokenProvider` and not `apiKeyProvider`? I think the former matches the existing terminology better

src/client.ts
```diff
@@ -438,6 +457,31 @@ export class OpenAI {
     return Errors.APIError.generate(status, error, message, headers);
   }
 
+  async _setToken(): Promise<boolean> {
```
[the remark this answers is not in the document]
deyaaeldeen: The decision to refresh is taken by the input function and not by this method, i.e. it is not guaranteed that the token will be refreshed if this method is called. However, let me know if you feel strongly about it and I can update the name.

tests/lib/azure.test.ts
```diff
@@ -268,9 +268,9 @@ describe('instantiate azure client', () => {
       );
     });
 
-    test.skip('AAD token is refreshed', async () => {
+    test('AAD token is refreshed', async () => {
       let fail = true;
-      const testFetch = async (url: RequestInfo, req: RequestInit | undefined): Promise<Response> => {
+      const testFetch = async (url: any, { headers }: RequestInit = {}): Promise<Response> => {
```
RobertCraigie, line 273: q: why did you have to use `any`?

src/client.ts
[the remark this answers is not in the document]
RobertCraigie: I do dislike `_set` but don't feel strongly about it being called `_refresh` - I find it weird that a method called `_setX()` doesn't actually take any arguments.

fwiw I was also framing "refresh" as in, refresh the `apiKey` property - not necessarily do a full token refresh.

wdyt about `_callApiKey()`? I don't think that's great either... my main fud with `_setApiKey()` is that the return value is not intuitive imo and it'd be great if we could find a name that makes that work

src/beta/realtime/websocket.ts
```diff
@@ -49,6 +50,10 @@ export class OpenAIRealtimeWebSocket extends OpenAIRealtimeEmitter {
 
     client ??= new OpenAI({ dangerouslyAllowBrowser });
 
+    if (typeof (client as any)?._options?.apiKey !== 'string') {
+      throw new Error('Call the create method instead to construct the client');
```
RobertCraigie, line 54: nit: this error message could be more helpful

RobertCraigie: also, wait does this work? because `_setApiKey()` doesn't mutate options, won't this check always return `true` if you pass in a function?

should we instead define the `client.apiKey` fallback string as a constant and check that? e.g. `if (client.apiKey === API_KEY_MISSING) { ...`

src/client.ts
```diff
@@ -385,7 +386,7 @@ export class OpenAI {
 
     this._options = options;
 
-    this.apiKey = apiKey;
+    this.apiKey = typeof apiKey === 'string' ? apiKey : 'Missing Key';
```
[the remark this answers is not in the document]
RobertCraigie: actually, we should move `API_KEY_SENTINEL` our of the azure specific file and use that

src/client.ts
```diff
@@ -453,6 +454,32 @@ export class OpenAI {
...
+      } catch (err: any) {
+        if (err instanceof Errors.OpenAIError) {
+          throw err;
+        }
+        throw new Errors.OpenAIError(
+          `Failed to get token from 'apiKey' function: ${err.message}`,
+          // @ts-ignore
+          { cause: err },
+        );
+      }
```
RobertCraigie, line 478: having a try catch around the whole block and then checking inside the catch if we threw isn't great, could we instead just either put the try catch just around the `await apiKey()`? or `await apiKey().catch(...)`?

src/beta/realtime/websocket.ts
deyaaeldeen: error messages are updated in https://github.com/openai/openai-node/pull/1587/commits/390c131bd67d8edce315efb4c9088e7ac371cd33

src/client.ts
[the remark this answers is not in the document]
deyaaeldeen: It is only used once here so I am not sure if there is value to have it in a constant var.

src/beta/realtime/websocket.ts
RobertCraigie: good point about the token potentially being expired but I think my other question still stands? does this check work as intended? because we don't mutate `_options`, so if you originally pass in a function then `typeof (client as any)?._options?.apiKey !== 'string'` would always be true?

deyaaeldeen: oh I see what you mean now. The constructor is called in the factory method too. Let me address this.

RobertCraigie (approved):
Looks great!

src/beta/realtime/websocket.ts
```diff
@@ -49,6 +50,16 @@ export class OpenAIRealtimeWebSocket extends OpenAIRealtimeEmitter {
 
     client ??= new OpenAI({ dangerouslyAllowBrowser });
 
+    if (hasProvider && !props?.__resolvedApiKey) {
+      throw new Error(
+        [
+          'Cannot open Realtime WebSocket with a function-based apiKey.',
+          'Use the factory so the key is resolved before connecting:',
+          '- OpenAIRealtimeWebSocket.create(client, { model })',
```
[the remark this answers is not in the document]
deyaaeldeen: Addressed in https://github.com/openai/openai-node/pull/1587/commits/4033c6a9982e2971f6ef6cd192a3dff873c65d7d

src/beta/realtime/websocket.ts
```diff
@@ -94,20 +105,23 @@ export class OpenAIRealtimeWebSocket extends OpenAIRealtimeEmitter {
...
-    client: Pick<AzureOpenAI, '_getAzureADToken' | 'apiVersion' | 'apiKey' | 'baseURL' | 'deploymentName'>,
+    client: Pick<AzureOpenAI, '_callApiKey' | 'apiVersion' | 'apiKey' | 'baseURL' | 'deploymentName'>,
     options: { deploymentName?: string; dangerouslyAllowBrowser?: boolean } = {},
   ): Promise<OpenAIRealtimeWebSocket> {
-    const token = await client._getAzureADToken();
+    const isToken = await client._callApiKey();
     function onURL(url: URL) {
-      if (client.apiKey !== '<Missing Key>') {
-        url.searchParams.set('api-key', client.apiKey);
+      if (isToken) {
```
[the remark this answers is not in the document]
deyaaeldeen: Addressed in https://github.com/openai/openai-node/pull/1587/commits/4033c6a9982e2971f6ef6cd192a3dff873c65d7d

src/azure.ts
```diff
@@ -140,47 +134,6 @@ export class AzureOpenAI extends OpenAI {
...
-     */
-    if (opts.headers.values.get('Authorization') || opts.headers.values.get('api-key')) {
-      return super.prepareOptions(opts);
-    }
-
-    const token = await this._getAzureADToken();
-    if (token) {
-      opts.headers.values.set('Authorization', `Bearer ${token}`);
-    } else if (this.apiKey !== API_KEY_SENTINEL) {
-      opts.headers.values.set('api-key', this.apiKey);
```
JounQin, line 178: Why `api-key` header support is removed? related #1633

PR #1596, opened 2025-07-23
feat: reject emitted promise on error in EventEmitter

bigint:

## Changes being requested

- Implemented rejection logic in EventEmitter.emitted so that if an “error” occurs before the target event, the returned Promise now rejects accordingly
- Added a dedicated test suite for EventEmitter.emitted to verify both successful resolution and rejection when an “error” event precedes the target event


HAYDEN-OAI (approved):

HAYDEN-OAI:
@bigint really appreciate your patience while we got this updated and reviewed! thanks for catching the missing error behavior and putting the initial fix together. this is merged now 🎉

PR #1638, opened 2025-09-03
[azure] Edit examples

deyaaeldeen:
To use the `OpenAI` client instead of the `AzureOpenAI` one.


trrwilson (approved):

RobertCraigie:
Is this functionally equivalent now or are there still cases where users may want to fallback to AzureOpenAI if something isn't supported / working yet?

Issue #1542, opened 2025-06-12
Embedding dimensions are truncated to 256 in some cases


dflatline:
### Confirm this is a Node library issue and not an underlying OpenAI API issue

- [x] This is an issue with the Node library

### Describe the bug

With some API endpoints, this package truncates the embedding dimension to 256. For models with a larger embedding dimension than 256, this is a massive accuracy loss.

So far, ollama and LM Studio openai-compatible endpoints suffer from this issue. llama-server does not.

This issue impacts RooCode (see https://github.com/RooCodeInc/Roo-Code/issues/4462).

### To Reproduce

Download embedding an model with more than 256 dimensions in ollama and LM Studio. For example, snowflake-arctic-embed-large-v2 has 1024 dimensions.
`ollama pull Definity/snowflake-arctic-embed-l-v2.0-q8_0`

Then, run the attached node script, and notice that the package truncates the expected dimensions of 1024 to 256. If the same model from LM Studio is served from llama-server, there is no problem:
`llama-server -m /Users/tinman/.cache/lm-studio/models/Casual-Autopsy/snowflake-arctic-embed-l-v2.0-gguf/snowflake-arctic-embed-l-v2.0-q6_k.gguf --embeddings -c 4096 -ngl 99`

[package.json](https://github.com/user-attachments/files/20702959/package.json)

[test-openai-embeddings.js.txt](https://github.com/user-attachments/files/20702960/test-openai-embeddings.js.txt)

### Code snippets

```JavaScript

```

### OS

macOS

### Node version

Node v23.11.0

### Library version

openai 4.104.0

anishesg:
Opened https://github.com/openai/openai-node/pull/2150 with a fix for this.

HAYDEN-OAI:
The OpenAI API honors the requested embedding `encoding_format`. Some compatible providers return numeric arrays even when the request asks for base64, which violates that contract and causes the reported decoding behavior. For those providers, pass `encoding_format: 'float'` explicitly. We are not planning to add provider-specific handling for responses that do not honor the requested encoding.

PR #1647, opened 2025-09-13
fix(embeddings): add null/undefined checks for malformed responses

SkinnnyJay:
## Changes being requested

Add defensive null/undefined checks for malformed embedding responses to prevent runtime crashes when using third-party OpenAI-compatible endpoints.

  **Core changes:**
  - Add null/undefined validation for embedding objects and data
  - Skip null embedding objects gracefully during processing
  - Throw meaningful OpenAIError with context for missing embedding data
  - Maintain full backward compatibility with existing functionality

  **Test coverage:**
  - Add 5 new test cases covering null/undefined scenarios
  - Add helper function `makeClientWithCustomResponse` for custom mock responses
  - Verify error messages include proper context (item index)
  - Ensure existing functionality remains unaffected

## Additional context & links

 **Bug:** https://github.com/openai/openai-node/issues/1542

This addresses potential runtime errors when third-party providers (Ollama, LM Studio) return malformed responses with null/undefined embedding data. The current code lacks defensive checks:

```typescript
// Before (vulnerable to runtime errors)
response.data.forEach((embeddingBase64Obj) => {
  const embeddingBase64Str = embeddingBase64Obj.embedding as unknown as string;
  embeddingBase64Obj.embedding = toFloat32Array(embeddingBase64Str);
});

// After (defensive with meaningful errors)
response.data.forEach((embeddingBase64Obj) => {
  if (!embeddingBase64Obj) return; // Skip null/undefined items

  const embeddingBase64Str = embeddingBase64Obj.embedding as unknown;
  if (embeddingBase64Str == null) {
    throw new OpenAIError(`Missing embedding data for item at index ${embeddingBase64Obj.index ?? 'unknown'}`);
  }
  embeddingBase64Obj.embedding = toFloat32Array(embeddingBase64Str as string);
});
```

  Impact:
  - ✅ Prevents runtime crashes from malformed third-party responses
  - ✅ Provides meaningful error messages for debugging
  - ✅ Maintains full backward compatibility
  - ✅ No performance impact on normal operations
  
 ✅ I understand that this repository is auto-generated, and my pull request may not be merged


HAYDEN-OAI:
Thanks for the contribution — closing this because the patch does not resolve the linked issue [#1542](https://github.com/openai/openai-node/issues/1542).

The reported truncation happens when an OpenAI-compatible provider returns a valid numeric embedding array despite the SDK's internal base64 request. This patch still passes that array to `toFloat32Array`, so a 1024-number embedding is interpreted as 1024 bytes and becomes 256 floats. The null checks also leave null entries inside the public `Embedding[]` response shape.

A correct downstream fix would decode strings, pass through numeric arrays, and throw a clear error for null/invalid items. This behavior lives in the custom embedding decode path introduced by [#1312](https://github.com/openai/openai-node/pull/1312), rather than the generated API schema. In the meantime, explicitly setting `encoding_format: 'float'` avoids the decode path for compatible providers.

Issue #1590, opened 2025-07-18
TypeError: Cannot read private member from an object whose class did not declare it at __classPrivateFieldGet


FaresKi:
### Confirm this is a Node library issue and not an underlying OpenAI API issue

- [x] This is an issue with the Node library

### Describe the bug

When running my code in production, I got this very random error:

```
TypeError: Cannot read private member from an object whose class did not declare it at __classPrivateFieldGet 

(file:///app/node_modules/.pnpm/openai@5.10.1_ws@8.18.3_zod@3.25.76/node_modules/openai/internal/tslib.mjs:14:15) at Proxy.buildURL
(/app/node_modules/.pnpm/openai@5.10.1_ws@8.18.3_zod@3.25.76/node_modules/openai/src/client.ts:446:23) at Proxy.buildRequest 
(/app/node_modules/.pnpm/openai@5.10.1_ws@8.18.3_zod@3.25.76/node_modules/openai/src/client.ts:818:22) at Proxy.makeRequest 
(/app/node_modules/.pnpm/openai@5.10.1_ws@8.18.3_zod@3.25.76/node_modules/openai/src/client.ts:532:46) at process.processTicksAndRejections (node:internal/process/task_queues:105:5) at async file:///app/node_modules/.pnpm/@langchain+openai@0.6.2_@langchain+core@0.3.65_@opentelemetry+api@1.9.0_@opentelemetry+_6b14e933d3433c65dede7732832881c4/node_modules/@langchain/openai/dist/embeddings.js:182:29 at async RetryOperation._fn (/app/node_modules/.pnpm/p-retry@4.6.2/node_modules/p-retry/index.js:50:12)
```
What's strange is that this error typically happens in a container with a Linux environment. In my case, Ubuntu.
I've tried to replicate it but I simply cannot.
Considering downgrading for now but I'm not sure if it's a "langchain/openai" issue, or an "openai" one.

### To Reproduce

I'm not even sure, I'm sorry.
i'm creating the openai client via langchain:
```
new ChatOpenAI({
     apiKey: process.env.OPENAI_API_KEY,
     modelName: "gpt-4o",
})
```

### Code snippets

```JavaScript

```

### OS

Ubuntu

### Node version

v22.17.1

### Library version

openai 5.10.1

RobertCraigie:
Thanks for the report, can you share more details about your setup? How are you running your code?

nbonamy:
[...]

PR #1648, opened 2025-09-14
fix(core): prevent private-field brand-check runtime crashes

SkinnnyJay:
## Changes being requested

Replace class private fields with module-scoped `WeakMaps` to prevent runtime crashes caused by JavaScript private field brand checks across duplicate SDK installations.

**Core changes:**
- Replace class `#private` fields with `WeakMap-based` encapsulation in core components
- Modify `core/api-promise.ts`, `core/pagination.ts`, and `core/streaming.ts`
- Add minimal inline rationale comments explaining the `WeakMap` approach
- Maintain existing encapsulation guarantees without public API changes

**Test coverage:**
- Add regression tests to prevent future brand-check crashes
- Verify the `cross-copy` method compatibility works correctly

## Additional context & links

**Bug:**  #1590

This addresses `"TypeError: Cannot read private member"` runtime crashes that occur when methods from different SDK copies interact, particularly in environments with duplicate SDK installations.

**Problem:**
JavaScript private field brand checks fail when instances cross module boundaries between different SDK copies, causing runtime errors in production environments.

**Solution:**
WeakMaps provide the same encapsulation benefits as private fields, but don't enforce brand checks, allowing instances from different SDK copies to interoperate safely.

**Impact:**
- ✅ Prevents runtime crashes in duplicate SDK installation scenarios
- ✅ Maintains existing encapsulation and security guarantees
- ✅ Improves cross-copy method compatibility
- ✅ No public API changes or breaking changes
- ✅ Minimal performance impact

✅ I understand that this repository is auto-generated, and my pull request may not be merged


mariocandela (approved):

SkinnnyJay:
Happy to help, @Pierre-Assemat @mariocandela. Just to confirm - are you referring to reviewing the current fix in this PR, or reporting a new issue caused by it?

If it’s the latter, could you share a minimal example and the `langgraph` version where you’re seeing the problem? I’ll verify that the fix covers your case as well.

HAYDEN-OAI:
Thanks for investigating this. The SDK intentionally retains JavaScript runtime-enforced private fields, and replacing them across the generated client, promises, pagination, and streaming would weaken encapsulation for all consumers. Proxy-based integrations should instead preserve or bind the original method receiver; affected Sentry versions have already been fixed upstream. Closing this implementation as not planned.

PR #1659, opened 2025-09-25
Pass through fetch error messages in APIConnectionError

### LINKED ISSUE #1656: `APIConnectionError` does not pass `message` argument

### Describe the bug

The library always returns `Connection Error` when hitting an invalid url instead of the full error from fetch (such as "ADDRINFO not found" or similar). This is happening because [this line](https://github.com/openai/openai-node/blob/5b275a9ff5e89f07a2d7719f5750c4771392478d/src/client.ts#L656) does not pass the `message` parameter required in [`APIConnectionError`](https://github.com/openai/openai-node/blob/5b275a9ff5e89f07a2d7719f5750c4771392478d/src/core/error.ts#L113-L120) which makes `message` undefined and resulting in the fallback string - "Connection error".

### To Reproduce

1. Setup OpenAI client with a custom fetch function returning a random error - "this fn does not make req"
2. Make a request with the client
3. The error is always "connection error" irrespective of what the fetch function contains.

### Code snippets

```JavaScript

```

### OS

macos

### Node version

20.19.0

### Library version

openai v5.22

cnaples79:
## Summary
- include the underlying fetch error message when constructing APIConnectionError
- assert the network error test surfaces the original message in addition to the error type

## Rationale
- fixes openai/openai-node#1656 so callers receive the full connection failure details instead of the generic "Connection error"

## Changes
- update the retry logic to pass the error message into APIConnectionError
- extend the streaming network error test to check the message is preserved

Fixes #1656

### DIFF
--- a/src/client.ts
+++ b/src/client.ts
@@ -653,7 +653,7 @@ export class OpenAI {
       if (isTimeout) {
         throw new Errors.APIConnectionTimeoutError();
       }
-      throw new Errors.APIConnectionError({ cause: response });
+      throw new Errors.APIConnectionError({ message: response.message, cause: response });
     }
 
     const specialHeaders = [...response.headers.entries()]
--- a/tests/lib/ChatCompletionRunFunctions.test.ts
+++ b/tests/lib/ChatCompletionRunFunctions.test.ts
@@ -2407,11 +2407,15 @@ describe('resource completions', () => {
         throw new Error('mock request error');
       }).catch(() => {});
 
-      async function runStream() {
-        await stream.done();
-      }
-
-      await expect(runStream).rejects.toThrow(APIConnectionError);
+      await expect(async () => {
+        try {
+          await stream.done();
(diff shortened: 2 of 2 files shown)


HAYDEN-OAI:
Thanks for the contribution — closing this because forwarding arbitrary transport text into the top-level error message is not the behavior we want.

The original fetch error is already preserved as `APIConnectionError.cause`, so callers that need low-level detail can inspect `error.cause`. Raw fetch messages vary by runtime and can expose URLs, hostnames, proxy, TLS, or internal-network details in logs and user-facing messages. We add curated actionable diagnostics where appropriate, such as the Undici dispatcher mismatch guidance in [#1898](https://github.com/openai/openai-node/pull/1898), while keeping the generic `Connection error.` fallback stable.

If we expand connection diagnostics further, it should be through curated or sanitized mappings rather than forwarding `response.message` verbatim.

Issue #1641, opened 2025-09-09
Realtime SDK does not match API


sethmoon:
### Confirm this is a Node library issue and not an underlying OpenAI API issue

- [x] This is an issue with the Node library

### Describe the bug

The [documentation](https://platform.openai.com/docs/api-reference/realtime_client_events/session/update) describes sending a `session.update` event with the following shape:

```json
{
    "type": "session.update",
    "session": {
        "type": "realtime",
        "instructions": "...",
        "audio": {
            "input": {
                "format": {
                    "type": "audio/pcmu"
                }
            },
            "output": {
                "format": {
                    "type": "audio/pcmu"
                }
            }
        }
    }
}
```

However, the [guide](https://platform.openai.com/docs/guides/realtime-conversations#handling-audio-with-websockets) shows different events and shapes. The equivalent of the above which I currently use, based on your previous code examples pre-ga:

```json
{
    "type": "session.update",
    "session": {
        "instructions": "...",
        "input_audio_format": "g711_ulaw",
        "output_audio_format": "g711_ulaw"
    }
}
```

When using the latest 5.20.0 version of the sdk, I cannot use the events listed in the API documentation, only the events listed in the guide. I assume the API documentation and SDK are built together since the SDK doesn't like the events in the guide, only the events in the API documentation. Sending the `session.update` with the current format in the API documentation yields an error `Unknown parameter: 'session.audio'.` The websocket URL it's connecting to is `wss://api.openai.com/v1/realtime?model=gpt-realtime`

### To Reproduce

1. Install latest `openai` package
2. `import { OpenAIRealtimeWS } from 'openai/realtime/ws';`
3. Open the socket and send a `session.update` with the above payload.
4. OpenAI sends back an error event with message `Unknown parameter: 'session.audio'.`

### Code snippets

```JavaScript

```

### OS

Fedora Linux 42

### Node version

Bun 1.2.21

### Library version

openai 5.20.0

RobertCraigie:
Sorry about this, should be fixed in the 5.23.1 release

PR #1661, opened 2025-09-26
fix(realtime): remove beta headers

Chonne:

## Changes being requested

Closes issue #1641

## Additional context & links

The [beta to GA migration guide](https://platform.openai.com/docs/guides/realtime#beta-to-ga-migration) mentions that the "OpenAI-Beta" should no longer be used for the GA version of the API. The types are up to date, but the code still adds the headers


RobertCraigie:
Thanks for the PR! We actually *just* shipped this fix at the same time as you opened this PR 😅

PR #1666, opened 2025-10-01
feat: Add support for zod@4 schemas

karpetrosyan:


tests/helpers/zod.test.ts
```diff
@@ -286,7 +290,7 @@ describe('zodResponseFormat', () => {
     `);
   });
 
-  it('throws error on optional fields', () => {
+  (version === 'v4' ? it.skip : it)('throws error on optional fields', () => {
```
karpetrosyan, line 293: as we transform the schema, we actually add missing fields in `required` property, as we do in python

src/helpers/zod.ts
```diff
@@ -22,6 +25,14 @@ function zodToJsonSchema(schema: ZodType, options: { name: string }): Record<str
   });
 }
 
+function nativeToJsonSchema(schema: ZodTypeV4): Record<string, unknown> {
```
RobertCraigie, line 28: nit: small preference for renaming this to include zod 4, and the `zodToJsonSchema()` function above to include zod 3 so its super explicit.

src/helpers/zod.ts
```diff
@@ -22,6 +25,14 @@ function zodToJsonSchema(schema: ZodType, options: { name: string }): Record<str
...
+  return toJSONSchema(schema, { target: 'draft-7' }) as Record<string, unknown>;
```
RobertCraigie, line 29: nit: this should call `toStrictJsonSchema()` as well

examples/parsing.ts
```diff
@@ -1,6 +1,6 @@
 import { zodResponseFormat } from 'openai/helpers/zod';
 import OpenAI from 'openai/index';
-import { z } from 'zod/v3';
+import { z } from 'zod/v3'; // Also works for 'zod/v4'
```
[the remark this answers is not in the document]
RobertCraigie: ah yeah separate PR for updating all examples sgtm as well

src/lib/transform.ts
```diff
@@ -0,0 +1,186 @@
+import type { JSONSchema, JSONSchemaDefinition } from './jsonschema';
+
+export function toStrictJsonSchema(schema: JSONSchema): JSONSchema {
+  return ensureStrictJsonSchema(schema, [], schema);
+}
+
+function isNullable(schema: JSONSchemaDefinition): boolean {
```
RobertCraigie: this should check `schema.nullable` as well right?

src/lib/transform.ts
```diff
@@ -0,0 +1,186 @@
...
+  if (typeof schema === 'boolean') {
+    return false;
+  }
+  for (const oneOfVariant of schema.oneOf ?? []) {
+    if (typeof oneOfVariant !== 'boolean' && oneOfVariant.type === 'null') {
+      return true;
+    }
```
RobertCraigie: this should be recursive

```suggestion
    if (isNullable(oneOfVariant)) {
      return true;
    }
```
also looks like `isNullable({ type: 'null' })` isn't handled

examples/responses/streaming-tools.ts
```diff
@@ -2,7 +2,7 @@
 
 import { OpenAI } from 'openai';
 import { zodResponsesFunction } from 'openai/helpers/zod';
-import { z } from 'zod/v3';
+import { z } from 'zod/v4'; // Also works for 'zod/v3'
```
dtmeadows: @karpetrosyan any chance you could run all these examples with v3 on this PR just to make sure things are back-compatible?

tests/lib/parser.test.ts
```diff
@@ -1,10 +1,14 @@
...
+import { z as z3 } from 'zod/v3';
 import { zodResponseFormat } from 'openai/helpers/zod';
 import { makeSnapshotRequest } from '../utils/mock-snapshots';
 
 jest.setTimeout(1000 * 30);
 
-describe('.parse()', () => {
+describe.each([
+  { version: 'v4', z: z4 as any },
+  { version: 'v3', z: z3 as any },
```
RobertCraigie: the `as any` cast isn't ideal, I assume you did that as the types were too complicated and caused errors? if that's the case, then I think we should just cast to `z3` so it's not a complicated union type
```suggestion
  { version: 'v4', z: z4 as typeof z3 },
  { version: 'v3', z: z3 },
```

RobertCraigie: then you could remove the `// @ts-ignore`s further down and just use z3.infer instead of z.infer

src/lib/transform.ts
karpetrosyan: fixed, thanks!

src/helpers/zod.ts
```diff
@@ -22,6 +25,18 @@ function zodToJsonSchema(schema: ZodType, options: { name: string }): Record<str
   });
 }
 
+function zodV4ToJsonSchema(schema: z4.ZodType): Record<string, unknown> {
+  return toStrictJsonSchema(
+    z4.toJSONSchema(schema, {
+      target: 'draft-7',
```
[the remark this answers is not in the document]
karpetrosyan: I've lost the context here a little bit, but looking at it again, it seems like we shouldn't set the concrete target. I'll remove it.

src/lib/transform.ts
RobertCraigie: wait what, is that openapi only?

src/helpers/zod.ts
```diff
@@ -11,8 +12,16 @@ import {
 import { zodToJsonSchema as _zodToJsonSchema } from '../_vendor/zod-to-json-schema';
 import { AutoParseableResponseTool, makeParseableResponseTool } from '../lib/ResponsesParser';
 import { type ResponseFormatTextJSONSchemaConfig } from '../resources/responses/responses';
+import { toStrictJsonSchema } from '../lib/transform';
+import { JSONSchema } from '../lib/jsonschema';
 
-function zodToJsonSchema(schema: ZodType, options: { name: string }): Record<string, unknown> {
+// Add this helper type at the top
```
RobertCraigie: AI comment? 😅

src/helpers/zod.ts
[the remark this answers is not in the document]
karpetrosyan: Ah, I think I did that because the current vendored version that parses Zod schema to JSON Schema [is using draft-07 by default](https://github.com/openai/openai-node/blob/3595555a834de8c51cf00440494b7d0013475ec8/tests/helpers/zod.test.ts#L19)

PR #1675, opened 2025-10-07
docs(azure): add Responses API example

### LINKED ISSUE #701: [Feature Request] Assistants API integration/example with Azure OpenAI

I understand there is an Azure specific SDK, but I've built a lot of my app using this library and believe this shouldn't be too hard to integrate and maintains parity with the `openai` python SDK. 

From this example, it looks like the python SDK can handle this pretty easily:

```python
from openai import AzureOpenAI

# Load the environment variables - These are secrets.
load_dotenv()

api_URI = os.getenv("OPENAI_URI")
api_KEY = os.getenv("OPENAI_KEY")
api_version = os.getenv("OPENAI_VERSION")
deployment_name = os.getenv("OPENAI_DEPLOYMENT_NAME")

# Create an OpenAI Azure client
client = AzureOpenAI(api_key=api_key,
        api_version=api_version,
        azure_endpoint=api_endpoint)
```

That's all it takes and then all the "beta" assistant methods are available:

```py
assistant = client.beta.assistants.create()
thread = client.beta.threads.create()
message = client.beta.threads.messages.create()
# etc.
```

Source: https://github.com/Azure-Samples/azureai-samples/tree/main/scenarios/Assistants/assistants-api-in-a-box

If this can't be achieved or isn't in the roadmap through this node library, I'd rather make my own REST methods rather than try to use 2 different libraries

### Additional context

_No response_

dhofheinz:
## Description

Adds an Azure OpenAI Responses API example using the current v1 endpoint and Microsoft Entra ID authentication.

This PR originally proposed an Assistants API example. It has been updated to use Responses because the Assistants API is deprecated and scheduled to shut down on August 26, 2026.

## Changes

- Adds examples/azure/responses.ts as an executable TypeScript example
- Uses the standard OpenAI client with the Azure /openai/v1/ endpoint
- Uses DefaultAzureCredential with a refreshable token provider
- Reads the Azure endpoint and model deployment from environment variables
- Demonstrates stateful follow-up turns with previous_response_id
- Resends instructions on each turn, since they are not inherited through response chaining

## Context

The repository already includes Azure Chat Completions and Realtime examples, but no Azure Responses example. This keeps the original math-tutor and multi-turn intent while moving the contribution to the supported replacement for Assistants.

Related to #701.

## Validation

- Prettier
- ESLint
- Isolated TypeScript compilation
- Full GitHub Actions suite


Edf2w5 (changes requested):
Go

Edf2w5 (approved):

HAYDEN-OAI (approved):

HAYDEN-OAI:
@dhofheinz really appreciate your patience while we got this updated! we adapted the example to the current azure responses api, and it is merged now. thanks again for taking the time to contribute!

PR #1692, opened 2025-10-27
docs(azure): use standard client for v1 chat

### LINKED ISSUE #1684: [chat.completions.create] 404 Problem When Call Azure OpenAI V1 Chat Completion Api

### Describe the bug

I want to call Azure openai [v1 chat completion api](https://learn.microsoft.com/en-us/azure/ai-foundry/openai/latest#create-chat-completion)
Ideally the request should like `POST {endpoint}/openai/v1/chat/completions`
But our node library send request like `POST {endpoint}/openai/v1/deployments/{model}/chat/completions?api-version=preview` which finally got 404 error

Image

I tried to update package to latest version 6.3.0 but it doesn't help.

I did a bit trouble shooting in source code and found that:

if the api address exists in [_deployments_endpoints](https://github.com/openai/openai-node/blob/master/src/azure.ts#L148) array, the it will set
`/deployments/${model}${options.path}` in the path (See https://github.com/openai/openai-node/blob/master/src/azure.ts#L134 )

I saw there is a similar issue #1607 ,but response api is not in the `_deployments_endpoints` array so it won't impact response api

### To Reproduce

Run the code like following, then I got 404 error

``` js
const options: AzureClientOptions = {
        apiKey: AZURE_OPENAI_API_KEY,
        baseURL:AZURE_OPENAI_ENDPOINT.endsWith("/")? AZURE_OPENAI_ENDPOINT+"openai/v1/":AZURE_OPENAI_ENDPOINT+"/openai/v1/",
        deployment: "gpt-4.1",
        apiVersion:"preview",
        dangerouslyAllowBrowser: true,
      };

 this.client = new AzureOpenAI(options);
const stream = await this.client.chat.completions.create({
      model: "gpt-4.1",
      messages: [
        {
          role: "user",
          content: "Hello",
        },
      ],
      stream: true,
    });
```
Please let me know if there is other way to call azure openai v1 chat completion api successfully

### Code snippets

```JavaScript

```

### OS

win11

### Node version

Node v22

### Library version

openai 6.3.0

wuxiaojun514:
## Summary

Azure OpenAI's v1 endpoints use the standard `OpenAI` client and do not include `/deployments/{model}` in the URL. This updates the Azure chat example and guide to use that supported configuration:

- build the base URL as `{endpoint}/openai/v1/`
- pass the Azure deployment name through the request's `model` field
- use the current Azure identity scope with a refreshable token provider
- keep `AzureOpenAI` guidance for dated API versions

## Why

The original runtime check matched any base URL containing `/v1`, which could disable deployment routing for unrelated legacy endpoints. It also left `AzureOpenAI` requiring and sending `api-version`, so it did not fully support Azure's v1 contract.

Using the standard client follows Microsoft's current v1 guidance and fixes the 404 from #1684 without changing legacy `AzureOpenAI` behavior.

## Validation

- `pnpm exec prettier --check azure.md examples/azure/chat.ts`
- `pnpm exec eslint examples/azure/chat.ts`
- `pnpm exec tsc --noEmit`

Fixes #1684


HAYDEN-OAI (approved):

HAYDEN-OAI:
@wuxiaojun514 thanks so much for flagging this and for your patience while we got the v1 approach sorted out! we've updated the fix to follow azure's current client guidance, and it's merged now 🎉

PR #1705, opened 2025-11-15
Fix: Sanitize whitespace in $ref paths for OpenAI strict mode

### LINKED ISSUE #1679: zodResponseFormat emits $ref values and definition keys with spaces

### Describe the bug

Based on my conversations with OpenAI enterprise support, a lot of our OpenAI call errors are due to `$ref` values with literal spaces. 

zodResponseFormat generates JSON Schema where reused sub-schemas are hoisted into definitions/$defs using keys derived from property names without sanitizing spaces. Corresponding $ref values
also include literal spaces (e.g., "#/$defs/example-scope_properties_group_properties_Thing With Spaces"). The OpenAI API at least sometimes choke on refs/keys containing spaces. 

Expected: no literal spaces in any $ref values or definition/$defs keys.

### To Reproduce

To Reproduce:
    1. Install openai and zod.
    2. Create a Zod schema with a reused sub-schema under a property name that includes spaces.
    3. Call zodResponseFormat(schema, "example-scope") and inspect json_schema.schema.
    4. Observe that at least one $ref and the corresponding definition key contain spaces.

Example observed (problematic):
refs: ["#/$defs/example-scope_properties_group_properties_Thing With Spaces"]
defs keys: ["example-scope_properties_group_properties_Thing With Spaces"]

Expectation: no spaces in any $ref or defs keys.

### Code snippets

```JavaScript
import { z } from "zod";
import { zodResponseFormat } from "openai/helpers/zod";

// Collect all "$ref" strings from a JSON Schema object
function collectRefs(obj: any, acc: string[] = []): string[] {
    if (typeof obj.$ref === "string") acc.push(obj.$ref);
    for (const v of Object.values(obj)) collectRefs(v, acc);
    return acc;
}

// Reusable sub-schema to encourage hoisting into definitions/$defs
const Thing = z.object({ id: z.string() });

// Property name intentionally includes spaces
const Root = z.object({
    group: z.object({
      "Thing With Spaces": Thing,
      AnotherUsage: Thing, // second use -> likely hoisted + $ref
    }),
});

const scopeName = "example-scope";
const rf = zodResponseFormat(Root, scopeName);
const schema = rf.json_schema.schema;

const refs = collectRefs(schema);
console.log("refs:", refs);
console.log("defs keys:", Object.keys(schema.definitions ?? schema.$defs ?? {}));
```

### OS

macOS (Darwin 24.6.0, arm64)

### Node version

v22.14.0

### Library version

5.23.0

PaulyBearCoding:
Fixes #1679

## Problem Statement

When using `zodResponseFormat` with Zod schemas that contain property names with whitespace characters, the generated JSON Schema produces `$ref` values and definition keys containing literal spaces. This causes validation failures when submitting the schema to the OpenAI API.

## Root Cause

In `src/_vendor/zod-to-json-schema/parseDef.ts` line 134, the `extract-to-root` case joins path segments with underscores but doesn't sanitize the segments themselves. Path segments like `"Thing With Spaces"` preserve their internal spaces in the final `$ref` value.

The `join('_')` method only adds separators between array elements - it does not modify the content within individual elements.

## Solution

Map over each path segment and replace all whitespace characters with underscores before joining:

\`\`\`typescript
const name = item.path
  .slice(refs.basePath.length + 1)
  .map((segment) => segment.replace(/\s+/g, '_'))
  .join('_');
\`\`\`

The regex `/\s+/g` matches all types of whitespace: spaces, tabs, newlines, and Unicode whitespace characters.

## Testing

**Unit tests:**
- All 18 tests pass (16 existing + 2 new)
- Both Zod v3 and v4 compatibility confirmed
- New test verifies $ref values and definition keys contain no spaces

**Comprehensive edge cases validated:**
- Multiple consecutive spaces
- Tabs and newlines
- Unicode spaces (nbsp, em space, etc.)
- Deeply nested structures
- Valid names with underscores/hyphens preserved
- Emoji and CJK characters
- Performance: 0.03ms per schema generation (1000 iterations)

**No breaking changes:**
- Valid identifiers (alphanumeric, underscores, hyphens) remain unchanged
- Only whitespace characters are affected

## Changes

**Modified files:**
- `src/_vendor/zod-to-json-schema/parseDef.ts` (4 lines)
- `tests/helpers/zod.test.ts` (added comprehensive test)

### DIFF
--- a/src/_vendor/zod-to-json-schema/parseDef.ts
+++ b/src/_vendor/zod-to-json-schema/parseDef.ts
@@ -131,7 +131,10 @@ const get$ref = (
     // `["#","definitions","contactPerson","properties","person1","properties","name"]`
     // then we'll extract it out to `contactPerson_properties_person1_properties_name`
     case 'extract-to-root':
-      const name = item.path.slice(refs.basePath.length + 1).join('_');
+      const name = item.path
+        .slice(refs.basePath.length + 1)
+        .map((segment) => segment.replace(/\s+/g, '_'))
+        .join('_');
 
       // we don't need to extract the root schema in this case, as it's already
       // been added to the definitions
--- a/tests/helpers/zod.test.ts
+++ b/tests/helpers/zod.test.ts
@@ -359,4 +359,33 @@ describe.each([
 
     expect(consoleSpy).toHaveBeenCalledTimes(0);
   });
+
+  it('sanitizes property names with spaces in $ref values', () => {
+    const Thing = z.object({ id: z.string() });
+    const Root = z.object({
+      group: z.object({
+        'Thing With Spaces': Thing,
+        AnotherUsage: Thing,
+      }),
+    });
+
+    const result = zodResponseFormat(Root, 'example-scope');
+    const schema = result.json_schema.schema;
+
+    // Check definitions keys (draft-07 uses "definitions" not "$defs")
+    const defs = (schema as any).definitions || (schema as any).$defs || {};
+    const defsKeys = Object.keys(defs);
+    const defsWithSpaces = defsKeys.filter((key: string) => key.includes(' '));
+    expect(defsWithSpaces).toEqual([]);
+
+    // Check all $ref values don't contain spaces
+    const schemaStr = JSON.stringify(schema);
+    const refMatches = schemaStr.match(/"\$ref"\s*:\s*"([^"]+)"/g) || [];
+
+    refMatches.forEach((match) => {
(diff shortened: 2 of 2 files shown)


HAYDEN-OAI:
@PaulyBearCoding thanks so much for digging into this and for the thoughtful report and test coverage! we found that replacing whitespace with underscores can collide with existing underscore names, so we are closing this in favor of #1903, which handles the same issue with collision-safe encoding. really appreciate you identifying the underlying problem and your patience here!

PR #1708, opened 2025-11-25
fix(types): incorrect value for prompt_cache_retention

redbmk:

## Changes being requested

When sending a request with `prompt_cache_retention: 'in-memory'`, the server responds with an error saying the only valid values are `in_memory` or `24h`. The typescript types though say that `in_memory` is not a valid value and should be `in-memory` instead. This updates  the types to reflect the server's requirements.


HAYDEN-OAI:
Verified against current `main` (v6.48.0): `prompt_cache_retention` now uses the `in_memory` literal ([source](https://github.com/openai/openai-node/blob/main/src/resources/chat/completions/completions.ts#L2215)). Closing as resolved upstream.

PR #1713, opened 2025-12-01
Add new tool call streaming system

404Wolf:

## Changes being requested

Update the tool call pattern to use a new async iterable style pattern that makes more easily accessible tool calls as they are fired, and API responses.


zanllan23-dev (approved):
Ok

src/lib/beta/BetaToolRunner.ts
```diff
@@ -0,0 +1,453 @@
...
+    this.#options = {
+      ...options,
+      headers: buildHeaders([{ 'x-stainless-helper': 'BetaToolRunner' }, options?.headers]),
+    };
+    this.#completion = promiseWithResolvers();
+  }
+
+  async *[Symbol.asyncIterator](): AsyncIterator<
+    Stream extends true ?
+      ChatCompletionStream // TODO: for now!
```
RobertCraigie: 🤨

src/lib/beta/BetaToolRunner.ts
```diff
@@ -0,0 +1,453 @@
...
+            this.#message?.catch(() => {});
+            yield stream as any;
+          } else {
+            this.#chatCompletion = this.client.beta.chat.completions.create(
+              {
+                ...apiParams, // spread and explicit so we get better types
+                stream: false,
+                tools: params.tools,
+                messages: params.messages,
+                model: params.model,
```
RobertCraigie: q: why are these explicitly listed?

src/lib/beta/BetaToolRunner.ts
```diff
@@ -0,0 +1,453 @@
...
+
+          const prevMessage = await this.#message;
+
+          if (!prevMessage) {
+            throw new OpenAIError('ToolRunner concluded without a message from the server');
+          }
+
+          if (!this.#mutated) {
+            // TODO: what if it is empty?
+            if (prevMessage) {
```
RobertCraigie: this seems redundant given we check `prevMessage` above?

src/lib/beta/BetaToolRunner.ts
```diff
@@ -0,0 +1,453 @@
...
+   * @example
+   * // Using a mutator function
+   * runner.setMessagesParams((params) => ({
+   *   ...params,
+   *   max_tokens: 100,
+   * }));
+   */
+  setMessagesParams(params: BetaToolRunnerParams): void;
+  setMessagesParams(mutator: (prevParams: BetaToolRunnerParams) => BetaToolRunnerParams): void;
+  setMessagesParams(
```
RobertCraigie: I think this should be `setCompletionsParams`, `setChatParams` or something along those lines as we should indicate it's for the API, with this naming I'd expect it to just take the `messages` property

RobertCraigie (commented):
Looking good.

Once we've ironed out some of the naming questions, could you also add docs for this?

src/lib/beta/BetaToolRunner.ts
404Wolf: Because typescript can't infer that `stream: false` even though we're in the else. tools/messages/model are not necessary

404Wolf: In completions.ts I saw we always explicitly provide stream (we don't leave it as undefined) so I think it's okay to override it here to get nicer inference

src/lib/beta/BetaToolRunner.ts
404Wolf: I was also a bit tripped up since anthropic also does it this way. Honestly I think the whole idea of thinking about this as a param manager is a little confusing. Part of me likes that it's so direct, but also, I think most users using this function just want ".updateChatHistory(prevHistory)". I'll update the name

RobertCraigie (commented):
Thinking through naming more _again_ (sorry), we'll likely want to add a similar helper for the Responses API as well in the future, so we should namespace these helpers in some fashion to make it clear they're for Chat Completions instead of Responses, e.g. `betaChatFunction()` etc.

Could you review the current naming and suggest some alternatives taking this into account?

README.md
```diff
@@ -124,6 +124,91 @@ await client.files.create({
...
+main();
+```
+
+With `.stream()` you get event handlers, accumulation, and an async iterable.
+
+Alternatively, you can use `client.chat.completions({ ..., stream: true })` which only returns an async iterable of the events in the stream and thus uses less memory (it does not build up a final message object for you).
+
+## Tool Helpers
+
+The SDK makes it easy to create and run [function tools with the chats API](https://platform.openai.com/docs/guides/function-calling). You can use Zod schemas or direct JSON schemas to describe the shape of tool input, and then you can run the tools using the `client.beta.messages.toolRunner` method. This method will automatically handle passing the inputs generated by the model into your tools and providing the results back to the model.
```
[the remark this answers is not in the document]
404Wolf: It's where they explain what a _function tool call_ actually is, but yes it focuses on the responses API. Should we re-explain it here instead and not link to it, since it could cause confusion?

zanllan23-dev (commented):
Good day

HAYDEN-OAI:
Thanks for the substantial proposal — closing this as stale/superseded by the current helper surface.

`chat.completions.runTools()` already handles automated multi-round tool execution and exposes completion and tool events. Merged [#1977](https://github.com/openai/openai-node/pull/1977) added typed `events(eventName)` async iterators for pull-based consumption, and the existing `afterCompletion` hook provides an awaited between-round control point.

The one remaining idea here — a per-round completion iterator — is a separate product enhancement, not something we want to land through a parallel Chat-only beta runner. If that control point is still needed, please open a focused issue describing the missing behavior so it can be designed against the current Chat Completions and Responses surfaces.

PR #1746, opened 2026-02-11
fix: fall back to entire response body when 'error' field is missing

### LINKED ISSUE #1734: The node.js client only reads errors from the "error" field of a response, with no fallback if it doesn't exist

# Describe the bug

## Quick summary

If an OpenAI compatible API endpoint returns error information in the response in a field other than "error", the error message won't get reported.

This differs in behavior from the python client for openai, which will fall back to putting the entire response body in the status code

## Is this a bug? 
Is the python implementation more correct than the node.js one?

I'm not sure which implementation is "more correct". Let me know what you think.
- From my personal experience, my feedback is the python implementation was more helpful for me, because the endpoint I was connecting to did not put error information in "error"
- Feel free to close this if it's considered "out of spec" and unsupported for an openai compatible api to not put errors in the "error" field. 
- I couldn't find a reference saying openai compatible apis must return errors in the "error" field.

I'm mostly making this issue to point out that the python client acts differently than the node.js one, and make people aware of that.
- One other issue had the"(no body)" errors, e.g. https://github.com/openai/openai-node/issues/1528
- As far as I can tell, the openai api returns errors in the "error" field.

## Sample working / non-working responses

Sample error response that will not be read by the openai node.js client (this will result in an error like `422 status code (no body)`):
```json
{
    "detail":"422: The model gpt-5-gibberish does not exist."
}
```
 
Sample error response from the OpenAI API that does work:
```json 
{
    "error": {
        "message":"The model `gpt-5-gibberish` does not exist or you do not have access to it.",
        "type":"invalid_request_error",
        "param":null,
        "code":"model_not_found"
    }
}
```

## node.js code trace

How the Node.js openai client looks for error messages:
1. At https://github.com/openai/openai-node/blob/44b7ac27bc989b2fc2bb2e831191821b11133d2e/src/core/error.ts#L69 , const error = (errorResponse as Record<string, any>)?.['error']; --  error will be undefined
2. At https://github.com/openai/openai-node/blob/44b7ac27bc989b2fc2bb2e831191821b11133d2e/src/core/error.ts#L50-52 , message and error will be undefined, msg will be undefined
    a. Yes, message is undefined, it's not a bug that it's undefined, that's from the code here https://github.com/openai/openai-node/blob/44b7ac27bc989b2fc2bb2e831191821b11133d2e/src/client.ts#L714 
3. You end up with error messages like 422 (no body)

In case this doesn't get fixed, a hacky workaround for this is to use dev tools and put a breakpoint at https://github.com/openai/openai-node/blob/44b7ac27bc989b2fc2bb2e831191821b11133d2e/src/core/error.ts#L69 
- Then run this command `errorResponse.error = errorResponse.detail`
- Let the execution continue, the error message will be better.

## Python code trace

See this code: https://github.com/openai/openai-python/blob/v2.15.0/src/openai/_client.py#L408
- In the python implementation, if it can't find a field named "error", it will just put the entire body in the status code 

See this pdb output for a demonstration

```log
python -m pdb openai_error_test.py
(Pdb) b <path/to>/python3.13/site-packages/openai/_client.py:409
(Pdb) c
-> if response.status_code == 400:
(Pdb) p body
{'detail': '422: The model claude-sonnet-4-5-gibberish does not exist or you do not have access to it.'}
(Pdb) p data
{'detail': '422: The model claude-sonnet-4-5-gibberish does not exist or you do not have access to it.'}
(Pdb) p body.get("error")
None
```

# To Reproduce

For my testing I used the chat completions API.

## Node.js reproduction steps

1. Take the basic example here https://github.com/openai/openai-quickstart-node/blob/master/chat_completions/index.js
2. Cause an error. The easiest way I know of is to set the model id to a gibberish value like gpt-5-gibberish
3. Connect to an API that puts error information in a field other than "error"

You should get an error like `422 status code (no body)`

## Python reproduction steps

1. Take the basic example for chat completions from this page:  https://github.com/openai/openai-python?tab=readme-ov-file#usage
2. Cause an error. The easiest way I know of is to set the model id to a gibberish value like gpt-5-gibberish
3. Connect to an API that puts error information in a field other than "error"

You should get something like this

> openai.UnprocessableEntityError: Error code: 422 - {'detail': '422: The model gpt-5-gibberish does not exist or you do not have access to it.'}

### OS

macOS

### Node version

v24.11.1

### Library version

v6.16.0

Limitless2023:
## Summary

Fixes #1734

When an OpenAI-compatible API returns error information in fields other than `error` (e.g. `detail` or `message`), the Node client currently reports `422 status code (no body)` instead of showing the actual error content. This differs from the Python client, which falls back to the entire response body.

## Changes

- **`src/core/error.ts`**: In `APIError.generate()`, when `errorResponse.error` is `undefined`, fall back to `errorResponse` itself (the entire parsed JSON body) instead of passing `undefined` to the error constructor. This is a one-line change: `?? errorResponse`.
- **`tests/error.test.ts`**: Added tests covering:
  - Standard OpenAI error response (with `error` field) — still works as before
  - Non-standard response with `detail` field — now shows the error content
  - Non-standard response with `message` field — now shows the error content  
  - Truly empty response — still shows `(no body)`
  - Raw text fallback — still works as before

## Before / After

**Before:** An error response like `{"detail": "422: The model gpt-5-gibberish does not exist."}` would produce:
```
UnprocessableEntityError: 422 status code (no body)
```

**After:** The same response now produces:
```
UnprocessableEntityError: 422 {"detail": "422: The model gpt-5-gibberish does not exist."}
```

This matches the Python client's behavior of falling back to the full response body when the `error` field is absent.

### DIFF
--- a/src/core/error.ts
+++ b/src/core/error.ts
@@ -66,7 +66,7 @@ export class APIError<
       return new APIConnectionError({ message, cause: castToError(errorResponse) });
     }
 
-    const error = (errorResponse as Record<string, any>)?.['error'];
+    const error = (errorResponse as Record<string, any>)?.['error'] ?? errorResponse;
 
     if (status === 400) {
       return new BadRequestError(status, error, message, headers);
--- a/tests/error.test.ts
+++ b/tests/error.test.ts
@@ -1,0 +1,51 @@
+import { APIError, BadRequestError, UnprocessableEntityError } from 'openai/core/error';
+
+describe('APIError.generate', () => {
+  const headers = new Headers({ 'x-request-id': 'req_123' });
+
+  test('extracts error from standard OpenAI error response', () => {
+    const err = APIError.generate(
+      422,
+      { error: { message: 'Invalid model', type: 'invalid_request_error', code: 'model_not_found' } },
+      undefined,
+      headers,
+    );
+    expect(err).toBeInstanceOf(UnprocessableEntityError);
+    expect(err.message).toContain('Invalid model');
+    expect(err.code).toBe('model_not_found');
+    expect(err.type).toBe('invalid_request_error');
+  });
+
+  test('falls back to entire response body when "error" field is missing', () => {
+    const err = APIError.generate(
+      422,
+      { detail: '422: The model gpt-5-gibberish does not exist.' },
+      undefined,
+      headers,
+    );
+    expect(err).toBeInstanceOf(UnprocessableEntityError);
+    expect(err.message).not.toContain('(no body)');
+    expect(err.message).toContain('gpt-5-gibberish');
+  });
+
+  test('falls back to response body with message field (non-standard)', () => {
+    const err = APIError.generate(
+      400,
+      { message: 'Something went wrong', code: 'bad_request' },
+      undefined,
+      headers,
+    );
+    expect(err).toBeInstanceOf(BadRequestError);
+    expect(err.message).toContain('Something went wrong');
+  });
+
+  test('still shows "(no body)" when response is truly empty', () => {
+    const err = APIError.generate(422, undefined, undefined, headers);
+    expect(err.message).toContain('(no body)');
+  });
+
+  test('uses errMessage string fallback when JSON parsing fails', () => {
+    const err = APIError.generate(422, undefined, 'raw error text', headers);
+    expect(err.message).toContain('raw error text');
+  });
+});


HAYDEN-OAI:
Thanks for the contribution. The canonical fix has now merged upstream in the shared TypeScript client generator: unwrapped JSON error bodies are normalized before APIError generation, preserving the parsed body, HTTP status, response headers, and structured error details without changing the handwritten error module. Closing this SDK-local change as superseded; the fix will be available after SDK regeneration and release.

PR #1762, opened 2026-03-09
fix(zod): convert oneOf to anyOf in toStrictJsonSchema for Zod v4 compatibility

### LINKED ISSUE #1709: Zod schema conversion broken for unions in Zod 4.1.13+

### Describe the bug

Zod recently started emitting `oneOf` instead of `anyOf` for discriminated unions:
https://github.com/colinhacks/zod/pull/5453

This is not supported for OpenAI strict mode, so using the Zod conversion helpers will cause API failures upon upgrading to the latest Zod.

### To Reproduce

- Define a discriminated union with Zod v4.1.13
- Use one of the `openai/helpers/zod` functions
- Use the result with the appropriate OpenAI endpoint
- It fails because the JSON schema includes `oneOf`, which is unsupported by OpenAI

### Code snippets

```JavaScript
import { z } from "zod";
import { zodResponseFormat } from "openai/helpers/zod";

const ResponseSchema = z.object({
  data: z.discriminatedUnion("type", [
    z.object({ type: z.literal("a") }),
    z.object({ type: z.literal("b") }),
  ]),
});

const client = new OpenAI({
  apiKey: process.env.OPENAI_API_KEY
});

const completion = await client.chat.completions.parse({
  model: "gpt-4o-mini",
  messages: [{ role: "user", content: "Choose a or b"}],
  response_format: zodResponseFormat(ResponseSchema, "choice"),
});

// Result:
// {
//   status: 400,
//   headers: Headers {},
//   requestID: 'req_a7f2a1dc6e19439289ff6992bec4a24d',
//   error: {
//     message: "Invalid schema for response_format 'choice': In context=('properties', 'data'), 'oneOf' is not permitted.",
//     type: 'invalid_request_error',
//     param: 'response_format',
//     code: null
//   },
//   code: null,
//   param: 'response_format',
//   type: 'invalid_request_error'
// }
```

### OS

N/A

### Node version

N/A

### Library version

openai v6.9.1

s-zx:
## Problem

Zod v4.1.13 changed discriminated unions to emit `oneOf` in JSON Schema output ([colinhacks/zod#5453](https://github.com/colinhacks/zod/pull/5453)). The OpenAI strict-mode API does not accept `oneOf` — it returns HTTP 400:

```
"Invalid schema for response_format 'choice': In context=('properties', 'data'), 'oneOf' is not permitted."
```

Any user on Zod v4.1.13+ who passes a discriminated union to `zodResponseFormat` / `zodTextFormat` / `zodTool` will hit this failure.

Reported in #1709.

## Fix

After the existing `anyOf` handling block in `ensureStrictJsonSchema`, add a `oneOf` → `anyOf` conversion step:

```ts
const oneOf = (jsonSchema as any).oneOf;
if (Array.isArray(oneOf)) {
  jsonSchema.anyOf = oneOf.map((variant, i) =>
    ensureStrictJsonSchema(variant, [...path, 'oneOf', String(i)], root),
  );
  delete (jsonSchema as any).oneOf;
}
```

Each variant still passes through `ensureStrictJsonSchema`, so nested objects get `additionalProperties: false` and required fields are set correctly. The conversion is semantically safe: Zod `oneOf` variants are mutually exclusive object schemas, and `anyOf` is the form the OpenAI API expects.

Fixes #1709

### DIFF
--- a/src/lib/transform.ts
+++ b/src/lib/transform.ts
@@ -110,11 +110,16 @@ function ensureStrictJsonSchema(
   // Handle oneOf — Zod v4 emits `oneOf` for discriminated unions (https://github.com/colinhacks/zod/pull/5453),
   // but the OpenAI strict-mode API only accepts `anyOf`.  The semantics are equivalent for the union types
   // that Zod produces here, so we convert in-place and recurse into each variant.
+  // Merge into an existing anyOf array when one is already present (schemas that legitimately
+  // declare both anyOf and oneOf would otherwise lose the anyOf constraints).
   const oneOf = (jsonSchema as any).oneOf;
   if (Array.isArray(oneOf)) {
-    jsonSchema.anyOf = oneOf.map((variant: JSONSchemaDefinition, i: number) =>
+    const convertedOneOf = oneOf.map((variant: JSONSchemaDefinition, i: number) =>
       ensureStrictJsonSchema(variant, [...path, 'oneOf', String(i)], root),
     );
+    jsonSchema.anyOf = Array.isArray(jsonSchema.anyOf)
+      ? [...jsonSchema.anyOf, ...convertedOneOf]
+      : convertedOneOf;
     delete (jsonSchema as any).oneOf;
   }
 


s-zx:
Good point — fixed. The previous code replaced `anyOf` outright with the converted `oneOf` variants. The updated logic merges them:

```ts
jsonSchema.anyOf = Array.isArray(jsonSchema.anyOf)
  ? [...jsonSchema.anyOf, ...convertedOneOf]
  : convertedOneOf;
```

This preserves any pre-existing `anyOf` alternatives when both arrays are present on the same schema node.

PR #1765, opened 2026-03-13
fix: correct prompt_cache_retention type to use underscore

### LINKED ISSUE #1756: Prompt Caching retention policy value mismatch

### Describe the bug

I noticed an inconsistency between the official documentation and the openai-node SDK regarding the prompt caching retention policy values.

[The allowed values in retention policy](https://developers.openai.com/api/docs/guides/prompt-caching#configure-per-request) are documented as `in_memory` and `24h` 

However, in the openai-node implementation, the value appears to be defined as 
``` typescript
  /**
   * The retention policy for the prompt cache. Set to `24h` to enable extended
   * prompt caching, which keeps cached prefixes active for longer, up to a maximum
   * of 24 hours.
   * [Learn more](https://platform.openai.com/docs/guides/prompt-caching#prompt-cache-retention).
   */
  prompt_cache_retention?: 'in-memory' | '24h' | null;
```

This is currently type-enforced in the SDK, so developers cannot use `in_memory` without modifying the type definitions. This makes it difficult to follow the official documentation as written.

This mismatch makes it unclear which value is officially supported and may cause confusion for developers following the documentation.

Could you please clarify:

* Which value is correct?
* Whether the documentation or the SDK should be updated for consistency?

If this is confirmed as an issue, I’d be happy to work on a fix (either updating the SDK or the documentation, depending on the intended behavior).

Thanks!

### To Reproduce

1. Install the latest version of openai (openai-node).
2. Create a minimal TypeScript file with the following code:

### Code snippets

```JavaScript
import OpenAI from 'openai';

const client = new OpenAI({
  apiKey: process.env.OPENAI_API_KEY!,
});

async function main() {
    model: 'gpt-4o-mini',
    input: 'Hello',
    // According to docs: "in_memory"
    prompt_cache_retention: 'in-memory',
  });

  console.log(response.output_text);
}

main();
```

### OS

macOS

### Node version

Node v25.2.1

### Library version

openai v6.25.0

fuleinist:
## Summary

This PR fixes a type definition mismatch in the `prompt_cache_retention` parameter.

According to the [OpenAI API documentation](https://developers.openai.com/api/docs/guides/prompt-caching), the allowed values for `prompt_cache_retention` are `in_memory` and `24h` (with underscore, not hyphen).

> If you don't specify a retention policy, the default is in_memory. Allowed values are in_memory and 24h.

However, the SDK type definition was using `in-memory` (with hyphen) instead of `in_memory` (with underscore), causing TypeScript errors when developers try to use the documented value.

## Changes

- Updated `prompt_cache_retention` type from `in-memory | 24h` to `in_memory | 24h` in:
  - `src/resources/responses/responses.ts` (3 occurrences)
  - `src/resources/chat/completions/completions.ts` (1 occurrence)
- Updated test file to use correct value

## Testing

The type change allows developers to use the documented `in_memory` value without TypeScript errors.

Fixes #1756

### DIFF
--- a/src/resources/chat/completions/completions.ts
+++ b/src/resources/chat/completions/completions.ts
@@ -1651,7 +1651,7 @@ export interface ChatCompletionCreateParamsBase {
    * of 24 hours.
    * [Learn more](https://platform.openai.com/docs/guides/prompt-caching#prompt-cache-retention).
    */
-  prompt_cache_retention?: 'in-memory' | '24h' | null;
+  prompt_cache_retention?: 'in_memory' | '24h' | null;
 
   /**
    * Constrains effort on reasoning for
--- a/src/resources/responses/responses.ts
+++ b/src/resources/responses/responses.ts
@@ -1085,7 +1085,7 @@ export interface Response {
    * of 24 hours.
    * [Learn more](https://platform.openai.com/docs/guides/prompt-caching#prompt-cache-retention).
    */
-  prompt_cache_retention?: 'in-memory' | '24h' | null;
+  prompt_cache_retention?: 'in_memory' | '24h' | null;
 
   /**
    * **gpt-5 and o-series models only**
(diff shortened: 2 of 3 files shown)


fuleinist:
Stale PR (no activity for 4+ weeks). Feel free to reopen if still relevant.

PR #1766, opened 2026-03-13
fix(zod): fully resolve extracted wrapper definitions

### LINKED ISSUE #1739: `zodTextFormat` generates invalid circular `$ref` in JSON schema definitions

### Describe the bug

When using `zodTextFormat` to convert a Zod schema, the generated JSON schema contains a `$ref` that references itself, creating an invalid circular definition that cannot be resolved.

### To Reproduce

1. Create a Zod schema as shown in the code snippet.
2. Pass it to zodTextFormat(schema, "slidePlan").
3. Inspect the generated definitions block in the output.

Notes:
1. Only occurs for `zod/v3`
2. Seems to be related to the `brand` applied to `BlockIdSchema`
3. If the result is used for `text.format` of a `responses.parse()` call, a rather cryptic error is shown:
```
SyntaxError: Unexpected end of JSON input 
    at JSON.parse (<anonymous>)
```

### Code snippets

```JavaScript
import { zodTextFormat } from "openai/helpers/zod";

import { z } from "zod/v3";

const BlockIdSchema = z.string().brand<"SlideId">();

const SlidePlanSchema = z.object({
  slides: z.array(
    z.object({
      subtitle: z.array(BlockIdSchema).nullable(),
      content: z.array(BlockIdSchema).min(1),
    }),
  ),
});

const jsonSchema = zodTextFormat(SlidePlanSchema, "slidePlan");
console.dir(jsonSchema, { depth: null });

```
This outputs:

```json
{
  type: 'json_schema',
  name: 'slidePlan',
  strict: true,
  schema: {
    type: 'object',
    properties: {
      slides: {
        type: 'array',
        items: {
          type: 'object',
          properties: {
            subtitle: {
              anyOf: [
                { type: 'array', items: { type: 'string' } },
                { type: 'null' }
              ]
            },
            content: {
              type: 'array',
              items: {
                '$ref': '#/definitions/slidePlan_properties_slides_items_properties_subtitle_anyOf_0_items'
              },
              minItems: 1
            }
          },
          required: [ 'subtitle', 'content' ],
          additionalProperties: false
        }
      }
    },
    required: [ 'slides' ],
    additionalProperties: false,
    definitions: {
      slidePlan_properties_slides_items_properties_subtitle_anyOf_0_items: {
        '$ref': '#/definitions/slidePlan_properties_slides_items_properties_subtitle_anyOf_0_items'
      },
      slidePlan: {
        type: 'object',
        properties: {
          slides: {
            type: 'array',
            items: {
              type: 'object',
              properties: {
                subtitle: {
                  anyOf: [
                    { type: 'array', items: { type: 'string' } },
                    { type: 'null' }
                  ]
                },
                content: {
                  type: 'array',
                  items: {
                    '$ref': '#/definitions/slidePlan_properties_slides_items_properties_subtitle_anyOf_0_items'
                  },
                  minItems: 1
                }
              },
              required: [ 'subtitle', 'content' ],
              additionalProperties: false
            }
          }
        },
        required: [ 'slides' ],
        additionalProperties: false
      }
    },
    '$schema': 'http://json-schema.org/draft-07/schema#'
  }
}
```

Under definitions, there is a circular reference:
```json
    definitions: {
      slidePlan_properties_slides_items_properties_subtitle_anyOf_0_items: {
        '$ref': '#/definitions/slidePlan_properties_slides_items_properties_subtitle_anyOf_0_items'
      },
    ...
```

### OS

Ubuntu

### Node version

v24.13.0

### Library version

openai v6.17.0

matantsach:
## Summary

Fixes #1739

When a branded Zod type (e.g. `z.string().brand<"Id">()`) is reused across multiple fields, the generated JSON schema contains a self-referencing definition — a definition whose value is just a `$ref` back to itself. This makes the schema invalid and causes a cryptic `SyntaxError: Unexpected end of JSON input` when used with `zodTextFormat` or `zodResponseFormat`.

## Root Cause

`parseBrandedDef` delegates to `parseDef` for the inner type but doesn't pass through the `forceResolution` parameter. When a definition is being force-resolved, the branded parser's inner call hits `refs.seen` and creates a `$ref` back to the same definition path — producing the circular reference.

The same structural issue exists in all transparent wrapper parsers (`catch`, `default`, `promise`, `readonly`) that delegate to `parseDef` without forwarding `forceResolution`. Only `parseEffectsDef` was already doing this correctly.

## Changes

- Pass `forceResolution` through `parseBrandedDef`, `parseCatchDef`, `parseDefaultDef`, `parsePromiseDef`, and `parseReadonlyDef` — matching the existing pattern in `parseEffectsDef`
- Add test reproducing the exact circular `$ref` from the issue (branded type reused in multiple fields)

## Testing

- New test verifies no self-referencing `$ref` in definitions and that the branded type resolves to `{type: "string"}`
- All 116 helper/lib tests pass
- Test is Zod v3 only (v4 uses `z4.toJSONSchema()` directly, not the vendored converter)


HAYDEN-OAI (approved):

HAYDEN-OAI:
@matantsach thanks so much for digging into this and for the thoughtful fix! really appreciate your patience while we got the broader wrapper coverage and reference handling sorted out. merging now 🎉

PR #1768, opened 2026-03-13
fix: add external_web_access option to WebSearchTool type

### LINKED ISSUE #1716: Types do not acknowledge web search's external_web_access option

### Describe the bug

When using the Responses API with the web_search tool, TypeScript flags `external_web_access: false` as an invalid option. It does indeed work, however (e.g. try a query like "What is NVDA's latest stock price?" - it will give you the latest without that option but with it, will say it doesn't have that information).

As far as I can tell, there's no option like this that works for the chat completions API using the specialized search models. I was thinking setting `web_search_options.external_web_access` to `false` would work but the API throws an error in that case.

### To Reproduce

install typescript and the latest openai sdk, e.g.: `bun add typescript openai`

create a typescript file, e.g. `index.ts`:

```typescript

new OpenAI().responses.create({
  input: "What is NVDA's latest stock price?",
  tools: [{ type: "web_search", external_web_access: false }],
  // tsc flags this ------------^
})
```

run `bunx tsc --noEmit`

### Code snippets

```JavaScript

```

### OS

macOS

### Node version

v22.19.0, bun v1.3.2

### Library version

openai v6.9.1

fuleinist:
## Summary

Adds the missing `external_web_access` option to the `WebSearchTool` interface.

## Changes

- Added `external_web_access?: boolean | null` property to the `WebSearchTool` interface
- Added JSDoc documentation for the new property

## Testing

- TypeScript compilation passes (errors are only from missing example dependencies)
- The new property matches the API behavior described in the issue

Fixes #1716

### DIFF
--- a/src/resources/responses/responses.ts
+++ b/src/resources/responses/responses.ts
@@ -7208,6 +7208,12 @@ export interface WebSearchTool {
    * The approximate location of the user.
    */
   user_location?: WebSearchTool.UserLocation | null;
+
+  /**
+   * Whether to allow the model to use external web access during the search.
+   * When `false`, the model will not access external web sources.
+   */
+  external_web_access?: boolean | null;
 }
 
 export namespace WebSearchTool {


fuleinist:
Stale PR (no activity for 4+ weeks). Feel free to reopen if still relevant.

PR #1779, opened 2026-03-19
fix(responses): type external_web_access on web_search tools

theamodhshetty:
## Summary
- add `external_web_access?: boolean` to the `WebSearchTool` response request type
- add a compile-time regression covering `tools: [{ type: "web_search", external_web_access: false }]`

## Why
The Responses API accepts `external_web_access` on `web_search` tools at runtime, but the Node SDK type currently rejects it. This patch aligns the request type with the runtime behavior the issue reports.

## Testing
- `./node_modules/.bin/jest tests/responses.test.ts --runInBand`
- `./node_modules/.bin/eslint src/resources/responses/responses.ts tests/responses.test.ts`
- `./scripts/build`

### DIFF
--- a/src/resources/responses/responses.ts
+++ b/src/resources/responses/responses.ts
@@ -7199,6 +7199,12 @@ export interface WebSearchTool {
   type: 'web_search' | 'web_search_2025_08_26';
 
   /**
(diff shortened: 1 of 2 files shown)


HAYDEN-OAI:

PR #1760, opened 2026-03-06
fix: export AzureClientOptions

dcol91863:
## Summary
Re-export `AzureClientOptions` from the package entrypoint so `import type { AzureClientOptions } from "openai"` works again.

## Changes
- export `AzureClientOptions` alongside `AzureOpenAI` from `src/index.ts`

## Testing
- Not run in this environment (`node` / `yarn` unavailable locally)

Fixes #1735

PR #1785, opened 2026-03-22
fix(types): re-export AzureClientOptions

theamodhshetty:
## Summary
- re-export `AzureClientOptions` from the package root again
- add a regression in the Azure package-surface test so `import type { AzureClientOptions } from 'openai'` stays covered

## Testing
- `./node_modules/.bin/jest tests/lib/azure.test.ts --runInBand`
- `./scripts/build`
- `./node_modules/.bin/eslint src/index.ts tests/lib/azure.test.ts`

### DIFF
--- a/src/index.ts
+++ b/src/index.ts
@@ -23,4 +23,4 @@ export {
   InvalidWebhookSignatureError,
 } from './core/error';
 
-export { AzureOpenAI } from './azure';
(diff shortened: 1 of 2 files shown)


HAYDEN-OAI:
Thanks for the contribution! This proposes the same AzureClientOptions root export already tracked in #1760, so I’m closing this as a duplicate and keeping the discussion there.

PR #1788, opened 2026-03-23
update

kristentr:


src/resources/responses/ws.ts
```diff
@@ -65,6 +65,135 @@ export class ResponsesWS extends ResponsesEmitter {
...
+        break;
+      case WS.WebSocket.CLOSING:
+        push({ type: 'closing' });
+        break;
+      case WS.WebSocket.CLOSED:
+        push({ type: 'close' });
+        done = true;
+        cleanup();
+        break;
+    }
```
abrahamcuenca: This looks like it could be a race condition. The socket can transition after listeners are attached BUT before the `readyState` switch runs. That might end up enqueueing dupes of open or close events. I wonder if we should route these through a guarded call like `emitOpen` / `emitClose`

ghost (changes requested):

HAYDEN-OAI:
Verified as superseded: this was an old 6.33.0 release PR, while the release train has advanced through later merged releases such as [#1944](https://github.com/openai/openai-node/pull/1944) and current `main` is v6.48.0 ([package](https://github.com/openai/openai-node/blob/main/package.json#L3)). Closing as done.

PR #1792, opened 2026-03-25
docs: fix client comment wording

Rohan5commit:
## Summary
- Update the `withOptions` comment in `src/client.ts` to use `reusing` instead of `re-using`.

## Related issue
- N/A (trivial comment wording fix)

## Guideline alignment
- Reviewed `CONTRIBUTING.md` before editing.
- Kept the diff to one hand-written source file with no behavior changes.

## Validation
- Not run (comment-only change).

### DIFF
--- a/src/client.ts
+++ b/src/client.ts
@@ -433,7 +433,7 @@ export class OpenAI {
   }
 
   /**
-   * Create a new client instance re-using the same options given to the current client with optional overriding.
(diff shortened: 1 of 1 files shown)


HAYDEN-OAI:
Thanks for the contribution! This is a minor wording-only change with no behavior or API-surface impact, so we’re going to pass on it and keep the generated SDK focused on substantive fixes.

PR #1793, opened 2026-03-25
feat(responses): add response stream accumulator helper

theamodhshetty:
## Summary

- add a public `openai/lib/responses/ResponseAccumulator` helper for building `Response` snapshots from `ResponseStreamEvent` streams
- refactor `ResponseStream` to reuse the shared accumulator
- handle response-representable delta, done, lifecycle, and tool-status events
- keep raw events safe to retain and replay by cloning payloads before snapshot mutation
- synthesize the SDK-only `output_text` field when raw lifecycle payloads omit it

## Behavior

Callers should always retain the snapshot returned by `accumulateResponse`. Incremental events update the supplied snapshot in place, while response lifecycle events return a detached authoritative snapshot.

Events without corresponding state in the `Response` object, such as audio chunks and partial image previews, intentionally leave the snapshot unchanged.

## Validation

- `pnpm exec jest tests/lib/ResponseAccumulator.test.ts tests/lib/ResponseStream.test.ts --runInBand`
- `pnpm exec tsc --noEmit`
- focused ESLint and Prettier checks
- `pnpm build`


HAYDEN-OAI (approved):

HAYDEN-OAI:
@theamodhshetty thanks so much for putting this helper together! really appreciate your patience while we got the accumulator behavior fully rounded out. merging now 🎉

PR #1816, opened 2026-04-06
docs: fix reusing wording in client docstring

Rohan5commit:
## Summary

Fix the client docstring by changing `re-using` to `reusing`.

## Related issue

N/A, trivial docstring wording fix.

## Guideline alignment

Single-file, non-behavioral wording change kept narrowly scoped.

## Validation/testing note

Not run; docstring-only change.

### DIFF
--- a/src/client.ts
+++ b/src/client.ts
@@ -433,7 +433,7 @@ export class OpenAI {
   }
 
   /**
-   * Create a new client instance re-using the same options given to the current client with optional overriding.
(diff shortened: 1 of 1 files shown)


HAYDEN-OAI:
Thanks for the contribution! This is a minor wording-only change with no behavior or API-surface impact, so we’re going to pass on it and keep the generated SDK focused on substantive fixes.

PR #1828, opened 2026-04-18
Fix abort listener cleanup in fetchWithTimeout

MukundaKatta:
## Summary
- remove the custom abort listener after fetch completion in fetchWithTimeout
- add a regression test covering successful cleanup of the listener

## Testing
- ./node_modules/.bin/jest tests/index.test.ts --runInBand --testNamePattern=custom signal|fetchWithTimeout removes custom abort listener after success|normalized method
- ./scripts/build

Closes #1811


HAYDEN-OAI:

Issue #1807, opened 2026-03-31
skills.create: getName() strips directory paths from File names, breaking required directory structure


akhileshrangani4:
## Bug

`client.skills.create({ files: [file] })` fails with `400 All files must be under a single top-level directory` when the `File` name includes a directory path (e.g. `my-skill/SKILL.md`).

The OpenAI Skills API requires files to be under a top-level directory. However, the SDK's internal `getName()` function in `internal/uploads.js` strips directory paths:

```js
function getName(value) {
    return (((typeof value === 'object' && value !== null &&
        (('name' in value && value.name && String(value.name)) || ...)) || '')
        .split(/[\\/]/).pop() || undefined);  // <-- strips directory
}
```

So `new File([content], "my-skill/SKILL.md")` becomes just `SKILL.md` in the multipart form.

## Expected behavior

`skills.create()` should preserve directory paths in file names since the Skills API requires them. The Anthropic SDK handles this correctly by passing `stripFilenames: false` for `skills.create`.

## Workaround

Build the `FormData` manually and use `fetch` directly:

```ts
const form = new FormData();
form.append("files", new Blob([content], { type: "text/markdown" }), "my-skill/SKILL.md");
const response = await fetch("https://api.openai.com/v1/skills", {
  method: "POST",
  headers: { Authorization: `Bearer ${apiKey}` },
  body: form,
});
```

## Environment

- `openai` package version: 5.8.0 (also confirmed on earlier versions)
- Node.js 22

matantsach:
Hi, I'd like to take this on.

## Proposed approach

Thread an opt-in `preservePath: true` option through the multipart helpers (`getName`, `createForm`, `multipartFormRequestOptions`, `maybeMultipartFormRequestOptions`) so `getName()` can skip its `.split(/[\\/]/).pop()` step when the caller needs the original `File.name`.

Then pass `{ preservePath: true }` from the two skills endpoints (`skills.create` and `skills.versions.create`), matching the Anthropic SDK's `stripFilenames: false` pattern you referenced.

Default behavior stays strip-on for all other endpoints — no change for existing callers.

Will submit a PR shortly.

PR #1833, opened 2026-04-19
Preserve nested filenames for skill uploads

MukundaKatta:
## Summary
- preserve nested file names for `skills.create()` and `skills.versions.create()` multipart uploads
- keep the existing filename stripping behavior for other upload endpoints
- add regression tests that assert `my-skill/SKILL.md` is preserved in the multipart body

## Testing
- not run locally: `yarn` is not installed in this environment


HAYDEN-OAI:
Thanks for investigating this! The Skills directory-upload filename issue is already tracked in #1807, and #1856 contains the more complete focused implementation. I’m closing this PR as a duplicate so we can consolidate the remaining discussion and review there.

Issue #1818, opened 2026-04-07
Chat Completions is not returning the metatada


juninhodeluca:
### Confirm this is a Node library issue and not an underlying OpenAI API issue

- [x] This is an issue with the Node library

### Describe the bug

Hi, I just noticed the `metadata` key is not being returned in response object when using the `https://api.openai.com/v1/chat/completions` endpoint despite of the current references says at [Create chat completion](https://developers.openai.com/api/reference/resources/chat/subresources/completions/methods/create#(resource)%20chat.completions%20%3E%20(method)%20create%20%3E%20(params)%200.non_streaming%20%3E%20(param)%20metadata%20%3E%20(schema))

Not sure if this is the right place to open this issue, but the `openai-openapi` looks a bit older.

Request log
```bash
[inx-nlp] [SILLY] [69d528cd4d585c4b66e400cf] runFlowExecutionJobWithAxios: making request to https://api.openai.com/v1/chat/completions with body 
{
    "frequency_penalty": 0,
    "presence_penalty": 0,
    "temperature": 0.8,
    "max_completion_tokens": 4096,
    "store": true,
    "model": "gpt-4.1",
    "messages": [
        {
            "role": "system",
            "content": "give a joke about IT support\n\n\n\n"
        }
    ],
    "metadata": {
        "executionId": "69d528cd4d585c4b66e400cf",
        "jobId": "69d528cd4d585c4b66e400da",
        "taskId": "65d4d28ab7b9b5bc5f77db4d"
    }
}
```

Response log
```bash
[inx-nlp] [INFO] [69d528cd4d585c4b66e400cf] runFlowExecutionJobWithAxios: received response from https://api.openai.com/v1/chat/completions: 
{
    "id": "chatcmpl-DS33ORr15wtnJiC5wIvGLIYtb1ubG",
    "object": "chat.completion",
    "created": 1775577294,
    "model": "gpt-4.1-2025-04-14",
    "choices": [
        {
            "index": 0,
            "message": {
                "role": "assistant",
                "content": "Why did the IT support technician refuse to fix the broken elevator?\n\nBecause he didn’t want to deal with another level of abstraction!",
                "refusal": null,
                "annotations": []
            },
            "logprobs": null,
            "finish_reason": "stop"
        }
    ],
    "usage": {
        "prompt_tokens": 19,
        "completion_tokens": 26,
        "total_tokens": 45,
        "prompt_tokens_details": {
            "cached_tokens": 0,
[...]

PR #1834, opened 2026-04-20
fix(types): include optional metadata on ChatCompletion responses (#1818)

Adityakk9031:

## Changes being requested

This PR addresses #1818 (`Chat Completions is not returning the metadata`).

### Root cause
`metadata` is accepted in `chat.completions.create` params and supported in stored completion flows (`retrieve`, `update`, `list`), but the `ChatCompletion` response type did not include a `metadata` field.  
This caused a contract/type mismatch in the SDK even when the API returns metadata for stored objects.

### What changed
- Added `metadata?: Shared.Metadata | null` to `ChatCompletion` in:
  - `src/resources/chat/completions/completions.ts`
- Added contextual docs indicating metadata is commonly present on stored completions and may be absent in immediate `create` responses.

### Scope
- Type/docs-only change.
- No runtime request/response parsing logic changed.

### DIFF
--- a/src/resources/chat/completions/completions.ts
+++ b/src/resources/chat/completions/completions.ts
@@ -279,6 +279,14 @@ export interface ChatCompletion {
   object: 'chat.completion';
 
   /**
+   * Developer-defined metadata attached to the completion.
+   *
+   * This is commonly present on stored completions (e.g. via `retrieve`/`list`) and
+   * may be omitted in immediate `create` responses.
+   */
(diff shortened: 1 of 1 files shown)


HAYDEN-OAI:
Verified against current `main` (v6.48.0): `ChatCompletion.metadata` is now present ([source](https://github.com/openai/openai-node/blob/main/src/resources/chat/completions/completions.ts#L2128)). Closing as resolved upstream.

PR #1835, opened 2026-04-21
fix(client): remove caller abort listener after successful fetch

MukundaKatta:
## Summary
- remove the forwarded abort listener from the caller signal once fetchWithTimeout completes
- avoid leaving an orphaned listener attached to successful requests
- add regression coverage for listener cleanup on the success path

## Testing
- npm test -- --runInBand tests/index.test.ts --testNamePattern="custom signal|removes the caller abort listener"
- npx prettier --check src/client.ts tests/index.test.ts


HAYDEN-OAI:

PR #1841, opened 2026-04-25
fix(client): restore 60s cap on Retry-After header

### LINKED ISSUE #1840: Retry-After 60-second cap silently removed in chore commit, was this intentional?

### Describe the bug

**Before (had cap):**
```typescript
}
```

**After (no cap):**
```typescript
}
```

**Questions:**

### To Reproduce

### Code snippets

```JavaScript
const client = new OpenAI();

```

### OS

### Node version

Node v25.9.0

### Library version

openai v6.34.0

MukundaKatta:
Closes #1840.

A refactor in 583240f dropped the 60s cap on Retry-After parsing, so the SDK now sleeps for whatever the server returns (sometimes 30+ minutes), freezing long-running scripts. Restored the original guard so values outside [0, 60s) fall back to calculated backoff.

Added test in tests/index.test.ts. Couldn't run yarn test locally due to sandbox restrictions; happy to adjust if CI flags anything.

### DIFF
--- a/src/client.ts
+++ b/src/client.ts
@@ -946,9 +946,9 @@ export class OpenAI {
       }
     }
 
-    // If the API asks us to wait a certain amount of time, just do what it
-    // says, but otherwise calculate a default
-    if (timeoutMillis === undefined) {
+    // If the API asks us to wait a certain amount of time (and it's a reasonable amount),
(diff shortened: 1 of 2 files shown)


HAYDEN-OAI:

PR #1843, opened 2026-04-26
docs: clarify abort-on-function-call example

### LINKED ISSUE #996: `Abort on a function call` example is confusing

### Describe the bug

Current doc ["Abort on a function call"](https://github.com/openai/openai-node/blob/19a5ba3b09299375b50b49d43a13bd89214b8575/helpers.md) is confusing. 

In general, I think use `abort` is not a good thing, we actually need `end/finish` method.

### To Reproduce

The example code actually not work if use `{ tool_choice: "required" }`? It seems `await runner.finalFunctionCall()` would just throw Error, and all async methods (eg. `finalXXX` and `totalUsage`) will throw error anyway after aborted.

### Code snippets

_No response_

### OS

macOS

### Node version

Node 22

### Library version

4.55.7

Rohan5commit:

## Changes being requested

Clarify the `Abort on a function call` example so it no longer calls `finalFunctionCall()` after `runner.abort()`, and document that the `final*` helpers reject once the run has been aborted.

## Additional context & links

- Fixes #996
- Documentation-only change in `helpers.md`
- Validation: `git diff --check`


HAYDEN-OAI (approved):

HAYDEN-OAI:
@Rohan5commit really appreciate your patience while we got this sorted out! thanks for flagging the confusing abort example and for putting the fix together. we tightened up the cancellation and error-handling details, and everything is green now. merging this in!

PR #1681, opened 2025-10-12
fix(responses): restore output_text on streamed final responses

alessiopelliccione:
## Description

Ensures ResponseStream.finalResponse() consistently exposes output_text for streamed Responses.

Raw streaming events may omit the SDK-only convenience field, so the parser now rebuilds it from the response output whenever the field is missing or null.

## Changes

- Synthesizes output_text in both parsed and unparsed response branches
- Preserves an existing non-null output_text value
- Updates the stream fixture to match the raw response.completed payload
- Adds focused regression coverage for finalResponse()

## Context

Fixes #1662.

## Validation

- Focused ResponseStream Jest suite
- Prettier and ESLint
- Full TypeScript check
- Full GitHub Actions suite

PR #1845, opened 2026-04-28
Add Responses testing example

MukundaKatta:
Adds a lightweight Responses API testing example that demonstrates how to assert response text, inspect function-call arguments, and collect streamed text/tool argument deltas without calling the API.\n\nVerified locally with:\n\n```\nnpm run tsn -T examples/responses/testing.ts\n```

### DIFF
--- a/examples/responses/testing.ts
+++ b/examples/responses/testing.ts
@@ -1,0 +1,92 @@
+#!/usr/bin/env -S npm run tsn -T
+
+import assert from 'node:assert/strict';
+
+type ResponseLike = {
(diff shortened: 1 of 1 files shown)


HAYDEN-OAI:
@MukundaKatta thanks so much for putting this together, and appreciate your patience while we worked through the queue! since this was opened, #1681 and #1793 landed the `output_text` and typed offline stream accumulation pieces in the SDK itself, so this example has been superseded. i'm going to close this one, but really appreciate the thoughtful contribution and the testing use case it highlighted!

Issue #1791, opened 2026-03-24
Responses API: undocumented reasoning+message pairing constraint breaks multi-turn conversations


achandmsft:
### Confirm this is a Node library issue and not an underlying OpenAI API issue

- [X] This is an issue with the Node library

### Describe the bug

Reasoning+message items must appear as consecutive pairs in `input`, but nothing documents this. Common multi-turn patterns silently produce orphaned items → 400 on the next turn.

This broke [OpenClaw](https://github.com/openclaw/openclaw/issues/49167) (64.9k forks) on `gpt-5.3-codex`. They had to add `downgradeOpenAIReasoningBlocks()` to strip orphan reasoning items.

The most common failure: filtering `response.output` to keep only messages (dropping reasoning) → `400: Item 'msg_...' of type 'message' was provided without its required preceding item of type 'reasoning'`.

Tested in JS, Python, Go, Java, .NET — identical results across all 5 SDKs. This is an API-level constraint, not SDK-specific. The SDK types don't prevent building `input` arrays that violate it.

| Model | A (all items) | B (msgs only) |
|---|---|---|
| gpt-5.3-codex (reasoning=high) | PASS | FAIL |
| o4-mini | PASS | FAIL* |

\* Nondeterministic — o4-mini sometimes returns reasoning-only output (no message), so B passes on those runs since there's no message to orphan. Codex with reasoning=high reliably returns both items, making it a deterministic repro.

Workaround: `previous_response_id`. For manual history, always pass reasoning+message pairs together.

Related:
- openai/openai-openapi#536 (spec root cause)
- openai/openai-python#3009 (same bug, Python SDK)
- openai/openai-dotnet#1050 (same bug, .NET SDK)
- openai/openai-go#631 (same bug, Go SDK)
- openai/openai-java#710 (same bug, Java SDK)
- [openclaw/openclaw#49167](https://github.com/openclaw/openclaw/issues/49167) (64.9k forks, production breakage)
- openai/openai-python#2561 (open since Aug 2025)
- openai/openai-agents-python#1660 (Agents SDK)

### To Reproduce

```TypeScript
import OpenAI from "openai";

const client = new OpenAI();
const conversation: any[] = [];

for (const msg of ["Write a Python prime checker.", "Add type hints.", "Add docstrings."]) {
    conversation.push({ role: "user", content: msg });
    const response = await client.responses.create({
        model: "gpt-5.3-codex",
        input: conversation,
        max_output_tokens: 300,
[...]

PR #1808, opened 2026-04-01
Add example and docs for reasoning-message pairing constraint (#1791)

iammannu:
Fixes #1791.

## Summary

Document the safe way to carry Responses API conversation state across turns.

- explain why filtering `response.output` to messages can drop required reasoning or tool-call context
- demonstrate preserving ordered, replayable output with `toResponseInputItems()`
- point to `previous_response_id` for simpler stateful continuation
- add a runnable TypeScript manual-history example under `examples/responses/`

## Validation

- `pnpm lint`
- focused TypeScript compilation for the new example
- GitHub Actions: build, lint, examples, ecosystem, breaking-change detection, Agents SDK regression detection, and Node 20/22/24/26 tests

PR #1846, opened 2026-04-28
Add Responses manual conversation state example

MukundaKatta:
Adds a Responses API example showing the safe manual conversation-state pattern: append the full `response.output` before the next user message so reasoning-capable models keep reasoning and message items paired across turns.\n\nAddresses #1791.\n\nVerification:\n- Not run: example calls the live API and requires `OPENAI_API_KEY`.

### DIFF
--- a/examples/responses/manual-conversation-state.ts
+++ b/examples/responses/manual-conversation-state.ts
@@ -1,0 +1,44 @@
+#!/usr/bin/env -S npm run tsn -T
+
+import OpenAI from 'openai';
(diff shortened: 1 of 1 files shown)


HAYDEN-OAI:
@MukundaKatta thanks again for digging into this and for calling attention to a tricky reasoning-history edge case! we landed #1808, which addresses #1791 with the same example plus safer `toResponseInputItems()` normalization and guidance for `previous_response_id`, so this one is now superseded. closing it out, and really appreciate the contribution!

PR #1847, opened 2026-04-28
Fix prompt cache retention literal

MukundaKatta:
Updates the `prompt_cache_retention` type literal from `in-memory` to `in_memory` to match the official prompt caching guide, which documents the allowed values as `in_memory` and `24h`.\n\nCloses #1756.\n\nVerification:\n- `npm test -- tests/api-resources/chat/completions/completions.test.ts`


HAYDEN-OAI:
Verified against current `main` (v6.48.0): `prompt_cache_retention` now uses the `in_memory` literal ([source](https://github.com/openai/openai-node/blob/main/src/resources/chat/completions/completions.ts#L2215)); this is the same resolved gap as [#1708](https://github.com/openai/openai-node/pull/1708). Closing as resolved upstream.

PR #1851, opened 2026-04-28
fix: preserve non-standard error response bodies

Genmin:
## Summary\n\n- fall back to the full JSON error response body when a response does not include the standard `error` envelope\n- keep the existing standard OpenAI error-envelope behavior unchanged\n- add a regression test for a 422 response body that only contains `detail`\n\nFixes #1734.\n\n## Testing\n\n- `yarn test tests/index.test.ts -t "error message falls back"`\n- `yarn build`\n

### DIFF
--- a/src/core/error.ts
+++ b/src/core/error.ts
@@ -67,7 +67,8 @@ export class APIError<
       return new APIConnectionError({ message, cause: castToError(errorResponse) });
     }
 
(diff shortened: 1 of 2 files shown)


HAYDEN-OAI:
Thanks for the contribution! Preserving error responses that do not contain a nested error field is already tracked in #1746, so I’m closing this as a duplicate and keeping the discussion there.

PR #1852, opened 2026-04-29
fix(responses): avoid parsing incomplete structured output

### LINKED ISSUE #1839: Structured responses fail with JSON parse error when incomplete

When using structured responses and an incomplete result is returned, the library throws a "SyntaxError: Unterminated string in JSON" or a similar JSON parsing error.

I'm filing this as a feature request because its reasonable that it throws a JSON parsing error, but it would make a huge difference if it returned a result indicating the problem. This is a very tricky issue to debug because it happens entirely within the library and by the time the exception is thrown its too late to figure out what happened.

Knowing the exact reason for the incomplete response would be helpful in knowing if the request should be retried, modified, or just skipped entirely.

This example is a bit contrived, but it needed to meet the minimum of 16 tokens to trigger the error.
```ts
import OpenAI from 'openai';
import { zodTextFormat } from 'openai/helpers/zod.mjs';
import { z } from 'zod';

const openai = new OpenAI({ /* ... */ });

const schema = z.object({
	size: z.enum(['small', 'medium', 'large']),
	quality: z.enum(['good', 'bad']),
	pea_description: z.string(),
})

const response = await openai.responses.parse({
	model: 'gpt-5.4-mini',
	instructions: 'You are a pea QA specialist. Determine if these peas are good.',
	input: 'Good large pea',
	max_output_tokens: 16, // This will trigger the error. Minimum is 16 tokens.
	text: {
		format: zodTextFormat(schema, 'pea_schema')
	},
});

console.log(response.output_parsed);
```

### Additional context

_No response_

cyphercodes:

## Changes being requested
Fixes #1839.

- Skip auto-parsing structured Responses output when the response is not completed, so incomplete responses can be inspected via `status` / `incomplete_details` instead of surfacing a JSON parse error first.
- Keep completed response parsing behavior unchanged and add focused coverage for both paths.

## Additional context & links
Verification:
- `yarn test tests/lib/ResponsesParser.test.ts --runInBand`
- `yarn build`
- `./node_modules/.bin/prettier --check src/lib/ResponsesParser.ts tests/lib/ResponsesParser.test.ts`
- `./node_modules/.bin/eslint src/lib/ResponsesParser.ts tests/lib/ResponsesParser.test.ts`
- `git diff --check`


HAYDEN-OAI (approved):

HAYDEN-OAI:
@cyphercodes thanks so much for the thoughtful fix, and appreciate your patience while we got this reviewed! this keeps incomplete structured responses inspectable without changing completed parsing behavior. everything is green, merging now!

PR #1855, opened 2026-04-29
fix: pass file batch helper options through

Genmin:
## Summary
- pass request options from `fileBatches.createAndPoll()` into the underlying batch create request
- pass request options from `fileBatches.uploadAndPoll()` into `createAndPoll()` so custom headers and `pollIntervalMs` survive the helper chain
- add a regression test covering upload, batch creation, and polling headers

## Validation
- `yarn test tests/api-resources/vector-stores/file-batches.test.ts`
- `yarn lint` reaches Prettier, ESLint, and build successfully; local typecheck then fails on pre-existing optional example dependencies not installed in this checkout: `@azure/identity`, `express`, and `next`


HAYDEN-OAI:
Thanks for the contribution! This fixes request-option propagation in the SDK’s vector-store polling helpers rather than the generated API schema. I’m moving it to SDK triage for review of the helper implementation.

HAYDEN-OAI (approved):
Verified batch create/upload/poll consistently forward request options and custom polling intervals. Regression coverage now lives in handwritten helper tests and checks headers on every request. Full handwritten/generated suites, lint, TypeScript, builds, breaking-change detection, ecosystem/examples, and Node 22/24/26 CI all pass.

PR #1856, opened 2026-04-29
fix: preserve Skills upload file paths

### LINKED ISSUE #1807: skills.create: getName() strips directory paths from File names, breaking required directory structure

Genmin:
Fixes #1807.

This keeps the existing safe default for multipart uploads: filenames inferred from `File`, `Response`, stream `path`, or `filename` are still stripped down to their basename unless a caller opts out.

The Skills API is the exception because its upload contract requires files to live under a single top-level directory, such as `my-skill/SKILL.md`. `skills.create()` and `skills.versions.create()` now opt into preserving filename paths when constructing multipart form data.

Validation:
- `yarn test tests/form.test.ts`


HAYDEN-OAI:
Thanks for the contribution! This is an SDK multipart-upload/helper change needed to preserve explicit Skills directory filenames, rather than an API-schema change. I’m moving it to SDK triage for review of the upload implementation.

HAYDEN-OAI (approved):
Verified both Skills creation endpoints preserve explicit directory filenames across buffered and streaming multipart uploads while inferred filesystem paths and response URLs remain sanitized. Confirmed ordinary uploads retain basename behavior, bearer authentication remains intact, and the full Node 22/24/26 CI matrix, breaking-change check, lint, build, examples, and ecosystem tests all pass.

PR #1857, opened 2026-04-29
fix: honor defaultHeaders null for helper method header

### LINKED ISSUE #1499: Setting `x-stainless-helper-method` to null in defaultHeaders gets ignored

### Describe the bug

Hardcoding the header in the function declaration is causing the defaultHeader override to not work.

https://github.com/openai/openai-node/blob/bc9f15fc7d1f4acf625adc3603577b06d59cdc5c/src/lib/ChatCompletionStreamingRunner.ts#L56
https://github.com/openai/openai-node/blob/bc9f15fc7d1f4acf625adc3603577b06d59cdc5c/src/lib/ChatCompletionRunner.ts#L52

Image

### To Reproduce

1. Create a new OpenAI instance
2. Set the defaultHeaders to include 'x-stainless-retry-count': null
3. See the header being included nonetheless

### Code snippets

```JavaScript
new OpenAI({
        baseURL: baseURL,
        apiKey: '',
        defaultHeaders: {
          'x-stainless-os': null,
          'x-stainless-runtime': null,
          'x-stainless-arch': null,
          'x-stainless-lang': null,
          'x-stainless-helper-method': null,
          'x-stainless-timeout': null,
          'x-stainless-package-version': null,
          'x-stainless-runtime-version': null,
          'x-stainless-retry-count': null,
        },
      }),
```

### OS

macOS

### Node version

v20.17.0

### Library version

openai@4.92.1:

Genmin:
Fixes #1499.

Helper-method headers were previously injected as ordinary per-request headers by helpers such as `runTools()`, `stream()`, and `chat.completions.parse()`. Since per-request headers are merged after `defaultHeaders`, users could not suppress `X-Stainless-Helper-Method` with `defaultHeaders: { 'x-stainless-helper-method': null }` even though that works for the other Stainless telemetry headers.

This moves helper-method markers into internal request metadata and emits the header from the client’s internal header block, before `defaultHeaders` are applied. Existing helper telemetry remains present by default, while explicit default-header nulls can now remove it consistently.

Validation:
- `yarn fix src/client.ts src/lib/ChatCompletionRunner.ts src/lib/ChatCompletionStreamingRunner.ts src/lib/ChatCompletionStream.ts src/lib/responses/ResponseStream.ts src/lib/AssistantStream.ts src/resources/chat/completions/completions.ts tests/lib/ChatCompletionRunFunctions.test.ts`
- `yarn test tests/lib/ChatCompletionRunFunctions.test.ts`


HAYDEN-OAI:
Thanks for the contribution! This fixes how the SDK’s chat, assistant, and response-streaming helpers propagate their helper-method headers so defaultHeaders can suppress them consistently. I’m moving it to SDK triage because the change spans the handwritten helper implementations and shared client runtime.

HAYDEN-OAI (approved):
Verified helper telemetry is emitted before configured defaults across chat parsing, tool runners, assistant streams, chat streams, and response streams. Confirmed case-insensitive default/request suppression and overrides, preservation of Headers instances, tuple headers, existing metadata, and unchanged default telemetry. TypeScript, lint, build, examples, ecosystem tests, breaking-change detection, and Node 22/24/26 CI all pass.

PR #1859, opened 2026-05-01
fix: run chat completion tools in parallel

### LINKED ISSUE #1131: runTools function isn't running tools in parallel

### Describe the bug

I'm attempting to have tools run in parallel using runTools and it doesn't work

### To Reproduce

1. Create a completion with multiple tools 
2. Call runTools with a prompt that will likely invoke multiple tool calls 
3. Wait for stream with it clearly being the case that the functions don't run in parallel

### Code snippets

_No response_

### OS

Mac

### Node version

Latest

### Library version

Latest

Genmin:
## Summary

Fixes #1131.

- Execute multiple function tool calls from one `runTools()` assistant turn concurrently when normal `auto` tool choice is in use.
- Preserve tool-result message order by appending results in the original `tool_calls` order before the next model request.
- Keep the existing specific `tool_choice` behavior serial so a single requested function still returns after the first successful execution.
- Add a regression test that observes concurrent execution while asserting ordered tool-result messages.

## Tests

- `yarn test tests/lib/ChatCompletionRunFunctions.test.ts --runInBand`
- `./node_modules/.bin/prettier --check src/lib/AbstractChatCompletionRunner.ts tests/lib/ChatCompletionRunFunctions.test.ts`
- `./node_modules/.bin/eslint src/lib/AbstractChatCompletionRunner.ts tests/lib/ChatCompletionRunFunctions.test.ts`
- `./scripts/build`
- `git diff --check`

Note: `./node_modules/typescript/bin/tsc --noEmit` is blocked in this local checkout by missing optional example dependencies: `@azure/identity`, `express`, and `next`.

### DIFF
--- a/src/lib/AbstractChatCompletionRunner.ts
+++ b/src/lib/AbstractChatCompletionRunner.ts
@@ -334,8 +334,9 @@ export class AbstractChatCompletionRunner<
         return;
       }
 
-      for (const tool_call of message.tool_calls) {
-        if (tool_call.type !== 'function') continue;
+      const runFunctionToolCall = async (
+        tool_call: ChatCompletionMessageFunctionToolCall,
+      ): Promise<{ tool_call_id: string; content: string; didExecute: boolean }> => {
         const tool_call_id = tool_call.id;
         const { name, arguments: args } = tool_call.function;
         const fn = functionsByName[name];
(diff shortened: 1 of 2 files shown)


HAYDEN-OAI:
@Genmin thanks so much for putting this together, especially the care around concurrent execution and preserving tool-result order! we've now merged #1949, which closes #1131 with the same core behavior plus the explicit parallel_tool_calls: false sequential path and safer all-settled error handling. we're going to close this one as superseded, but we really appreciate the thoughtful contribution and regression coverage!

PR #1861, opened 2026-05-01
docs: clarify ResponseStreamEvent description

### LINKED ISSUE #1699: Incorrect documentation ("Emitted when there is a partial audio response") for ResponseStreamEvent
See: https://github.com/openai/openai-node/blob/7a97bfcf1ea071bc605c654161cc0e4867d22960/src/resources/responses/responses.ts#L4500

The documentation that appears here ("Emitted when there is a partial audio response.") is intended for the first member of the union (ResponseAudioDeltaEvent), rather than for the union itself (ResponseStreamEvent)

HenryLee789:
## Summary
- replace the ResponseStreamEvent union doc comment with a general streaming-event description
- keep the existing ResponseAudioDeltaEvent-specific comment on the audio event itself

Fixes #1699.

## Tests
- git diff --check

Note: this is a comment-only documentation fix; node_modules is not installed locally in this worktree.

### DIFF
--- a/src/resources/responses/responses.ts
+++ b/src/resources/responses/responses.ts
@@ -5949,7 +5949,7 @@ export interface ResponseRefusalDoneEvent {
 export type ResponseStatus = 'completed' | 'failed' | 'in_progress' | 'cancelled' | 'queued' | 'incomplete';
 
 /**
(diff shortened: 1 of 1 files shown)


HAYDEN-OAI:

PR #1863, opened 2026-05-01
fix: cap retry-after retry delay

### LINKED ISSUE #1840: Retry-After 60-second cap silently removed in chore commit, was this intentional?

Genmin:
## Summary

- restore the prior safety cap for server-provided `Retry-After` values
- fall back to the SDK's default exponential retry backoff when `Retry-After` is missing, zero/invalid, or >= 60 seconds
- add a regression test so a large `Retry-After: 600` does not make the SDK wait ten minutes before retrying

Fixes #1840.

## Validation

- `./node_modules/.bin/jest tests/index.test.ts --runInBand -t 'retry on 429 with long retry-after falls back to default backoff|retry on 429 with retry-after|retry on 429 with retry-after-ms'`
- `./scripts/build`
- `./node_modules/typescript/bin/tsc --noEmit -p tsconfig.build.json`
- `git diff --check`

### DIFF
--- a/src/client.ts
+++ b/src/client.ts
@@ -946,9 +946,10 @@ export class OpenAI {
       }
     }
 
-    // If the API asks us to wait a certain amount of time, just do what it
-    // says, but otherwise calculate a default
-    if (timeoutMillis === undefined) {
+    // If the API asks us to wait a reasonable amount of time, just do what it
+    // says, but otherwise calculate a default.
+    const maxRetryAfterMillis = 60 * 1000;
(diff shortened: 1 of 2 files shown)


HAYDEN-OAI:
Thanks for the contribution! The Retry-After delay limit is already tracked in #1841 and issue #1840, so I’m closing this as a duplicate and keeping the discussion there.

Issue #1850, opened 2026-04-28
images.generate types advertise output_format="webp" for gpt-image-2, but API ignores it and returns PNG


pedrobarretocw:
### Confirm this is a Node library issue and not an underlying OpenAI API issue

- [x] This is an issue with the Node library

### Describe the bug

The SDK type `ImageGenerateParamsBase` declares

```ts
output_format?: 'png' | 'webp' | 'jpeg'
```

as valid for all GPT image models. In practice, when calling `images.generate` with `model: "gpt-image-2"` and `output_format: "webp"`, the API returns PNG-encoded bytes in `data[0].b64_json` regardless of the requested format. JPEG is honored correctly, so the SDK is sending the field — the API silently ignores `webp` for this model.

The SDK type therefore lies to callers: the `.jpeg` path works, the `.webp` path returns PNG bytes with no error and a misleading filename if you derive the extension from the request.

Two acceptable fixes:

1. **SDK side** — narrow `output_format` to `'png' | 'jpeg'` for `gpt-image-2`, surfacing the limitation at compile time.
2. **API side** — honor `output_format: "webp"` for `gpt-image-2`.

### To Reproduce

1. Call `images.generate` with `model: "gpt-image-2"`, `output_format: "webp"`, any quality.
2. Decode `result.data[0].b64_json` to bytes.
3. Inspect the first 8 bytes — they are `89 50 4E 47 0D 0A 1A 0A` (PNG magic) instead of `52 49 46 46 .. .. .. .. 57 45 42 50` (RIFF/WEBP).
4. Repeat with `output_format: "jpeg"` — bytes start with `FF D8 FF` (JPEG magic) as expected.

Reproduces identically on `quality: "auto" | "low" | "medium" | "high"`.

### Code snippets

```JavaScript
import OpenAI from "openai";
const client = new OpenAI();

const r = await client.images.generate({
  model: "gpt-image-2",
  prompt: "solid red square",
  size: "1024x1024",
  quality: "low",
  output_format: "webp",
  output_compression: 30,
  n: 1,
});

const bytes = Buffer.from(r.data[0].b64_json, "base64");
console.log(bytes.subarray(0, 4).toString("hex")); // "89504e47" → PNG, not WEBP
```

Raw curl confirms it is API-side, not SDK-side:

```bash
curl -X POST https://api.openai.com/v1/images/generations \
  -H "Authorization: Bearer \$OPENAI_API_KEY" \
  -H "Content-Type: application/json" \
[...]

PR #1865, opened 2026-05-04
fix(images): restrict webp for gpt-image-2 and add missing model

Deepakhub202:
## Changes being requested

Fix incorrect `output_format` typing for image generation.

Instead of removing `webp` globally, this change restricts `webp` only for `gpt-image-2` using conditional types.

- `gpt-image-2` → allows only `png | jpeg`
- other models → still allow `webp`

Also updates related types to work with the new union-based structure.

## Additional context & links

Related issue: #1850

The API ignores `output_format: "webp"` for `gpt-image-2` and always returns PNG, so the SDK should reflect this behavior accurately.


Deepakhub202:
Hi, I’ve updated the fix to use conditional typing so that `webp` is only restricted for `gpt-image-2` while remaining available for other models. Let me know if any further changes are needed. @pedrobarretocw

HAYDEN-OAI:
Thanks for reporting this! The gpt-image-2 model identifiers are already included in the current SDK. The remaining WebP behavior reproduces in direct API requests, so it needs to be addressed as an API issue rather than by narrowing otherwise supported SDK types. Please continue tracking the API behavior in #1850; closing this SDK PR.

PR #1871, opened 2026-05-11
Clarify abort behavior in function call helper docs

### LINKED ISSUE #996: `Abort on a function call` example is confusing

Jaksenc:
## Summary
- clarify that calling `runner.abort()` rejects promise-returning runner helpers with `APIUserAbortError`
- update the function-call abort example to capture the needed value before aborting
- avoid awaiting `finalFunctionCall()` after aborting the runner

Closes #996

## Testing
- `npm exec --yes --package prettier@3.7.1 -- prettier --check helpers.md`
- `git diff --check`

### DIFF
--- a/helpers.md
+++ b/helpers.md
@@ -601,6 +601,9 @@ The underlying `AbortController` for the runner.
 
 If you have a function call flow which you intend to _end_ with a certain function call, then you can use the second
 argument `runner` given to the function to either mutate `runner.messages` or call `runner.abort()`.
+Calling `runner.abort()` cancels the runner, so promise-returning helpers like `runner.done()`,
+`runner.finalChatCompletion()`, and `runner.finalFunctionToolCall()` will reject with an `APIUserAbortError`.
+Capture any data you need before aborting, or listen for the relevant event.
 
 ```ts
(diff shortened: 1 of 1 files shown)


HAYDEN-OAI:
@Jaksenc thanks so much for putting this together! #1843 has now merged and closes #996 with the same core fix, plus the conditional abort behavior, typed APIUserAbortError handling, and a warning that already-running tool callbacks continue. since that fully covers this change, we're going to close this as superseded. really appreciate you catching the confusing example and taking the time to improve it 💚

PR #1672, opened 2025-10-06
docs(files): clarify files.content returns Response object

dhofheinz:
## Changes being requested

Updated documentation for `openai.files.content()` to clarify that it returns a `Response` object, not raw file content. The previous JSDoc comment was ambiguous, leading users to expect file content directly.

### What changed:
- Enhanced JSDoc comment to explicitly state Response object is returned
- Added documentation for `.text()`, `.blob()`, and `.arrayBuffer()` consumption methods
- Provided two practical code examples in JSDoc
- Created a comprehensive example file demonstrating all consumption patterns

### Files modified:
- `src/resources/files.ts` - Updated JSDoc (lines 66-85)
- `examples/file-content-retrieval.ts` - New example file (NEW)

## Additional context & links

This addresses the 15-month-old issue #958 where users were confused about the return type. The fix follows the exact pattern established in `src/resources/audio/speech.ts` for documenting Response-returning methods.

**Validation**:
- ✅ All but one tests passing (680/681)
- ✅ Build successful
- ✅ Linting clean

Fixes #958

---

**Note**: I understand that this repository is auto-generated and my pull request may not be merged. However, this documentation enhancement provides immediate value to users and follows the established patterns in the codebase. The `examples/` directory is protected from generator overwrites, ensuring the example file persists.

PR #1874, opened 2026-05-12
Document file content response handling

### LINKED ISSUE #958: `openai.files.content` returns raw Response object

### Describe the bug

In the documentation, `openai.files.content` is described as returns content as string type.
However, it returns raw Response type.

```TS
import OpenAI from "openai";

const openai = new OpenAI();

async function main() {
  const file = await openai.files.content("file-abc123");

  console.log(file); // Expected to show a string (ref. https://platform.openai.com/docs/api-reference/files/retrieve-contents)
}

main();
```

Output:
Screenshot 2024-07-27 at 17 47 19

### To Reproduce

1. Write the code above (copy&pasted from https://platform.openai.com/docs/api-reference/files/retrieve-contents)
2. It shows `file` is a Response object

### Code snippets

_No response_

### OS

macOS

### Node version

Node v18.17.0

### Library version

openai v4.53.1

Jaksenc:
## Summary
- document that `client.files.content()` returns a web-standard `Response` and show reading it with `.text()`
- align the exported `FileContent` alias with the actual response type

Closes #958.

## Validation
- `git diff --check`
- `npx --yes prettier@3.7.1 --check src/resources/files.ts`
- `./node_modules/.bin/tsc --noEmit --pretty false --target es2020 --lib es2020 --module commonjs --moduleResolution node --esModuleInterop --strict --skipLibCheck src/resources/files.ts`

Note: full `tsc --noEmit` still requires optional example dependencies such as `@azure/identity`, `express`, and `next`.

### DIFF
--- a/README.md
+++ b/README.md
@@ -223,6 +223,16 @@ await client.files.create({
 });
 ```
 
+To retrieve uploaded file contents, use `files.content()` and consume the returned
+web standard `Response`:
+
+```ts
+const response = await client.files.content('file-abc123');
+const fileContents = await response.arrayBuffer();
+```
+
+For text files, you can use `await response.text()` instead.
+
 ## Webhook Verification
 
 Verifying webhook signatures is _optional but encouraged_.
--- a/src/resources/files.ts
+++ b/src/resources/files.ts
@@ -120,7 +120,7 @@ export class Files extends APIResource {
 
 export type FileObjectsPage = CursorPage<FileObject>;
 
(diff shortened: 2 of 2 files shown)


HAYDEN-OAI:
Thanks for the contribution! The same files.content documentation issue is already tracked in #958 and addressed more comprehensively in #1672, including binary-safe response consumption. I’m closing this as a duplicate so we can keep discussion there.

PR #1877, opened 2026-05-15
fix: remove abort listener after fetch settles

MukundaKatta:
Fixes #1811.

## Summary

- remove the forwarded caller abort listener after `fetchWithTimeout` settles
- keep the existing internal timeout cleanup behavior
- add a regression test that verifies the caller signal listener is detached after a successful request

## Verification

- `yarn test tests/index.test.ts --runInBand`
- `npx -p node@20 -c './node_modules/.bin/prettier --check src/client.ts tests/index.test.ts && ./node_modules/.bin/eslint src/client.ts tests/index.test.ts'`
- `npx -p node@20 -c './scripts/build'`

Note: full `yarn lint` proceeds through Prettier, ESLint, and build under Node 20, then local type-checking fails because optional example dependencies are not installed in this checkout: `@azure/identity`, `express`, and `next`.

### DIFF
--- a/src/client.ts
+++ b/src/client.ts
@@ -969,6 +969,7 @@ export class OpenAI {
       return await this.fetch.call(undefined, url, fetchOptions);
     } finally {
       clearTimeout(timeout);
+      if (signal) signal.removeEventListener('abort', abort);
     }
   }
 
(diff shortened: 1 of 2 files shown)


HAYDEN-OAI:

PR #1878, opened 2026-05-15
Fix abort signal listener cleanup

MukundaKatta:


HAYDEN-OAI:

PR #1881, opened 2026-05-15
fix: preserve options in vector store file batch upload polling

### LINKED ISSUE #1700: Function vectorStores.fileBatches.uploadAndPoll returns vectorStoreId instead of batchId

### Describe the bug

Hola :)

TLDR;

`vectorStores.fileBatches.uploadAndPoll` is returning the vector store id instead of the batch id

I'm using this API `vectorStores.fileBatches.uploadAndPoll` but when the return value seems like vector store instead of vector store file batch.

If using the id of the return value in subsequent API calls of `vectorStores.fileBatches` it will failed because of the incorrect id.

So I logged what is the value from `vectorStores.fileBatches.uploadAndPoll` and get this:

```js
{
  id: 'vs_my-vector-store-id',
  object: 'vector_store',
  created_at: 1762447816,
  name: 'The name of the vector store',
  description: null,
  usage_bytes: 0,
  file_counts: { in_progress: 0, completed: 1, failed: 0, cancelled: 0, total: 1 },
  status: 'completed',
  expires_after: null,
  expires_at: null,
  last_active_at: 1762448186,
  metadata: {}
}
```

### To Reproduce

- Call `vectorStores.fileBatches.uploadAndPoll`
- Get the id of the return object
- Call `vectorStores.fileBatches.listFiles`
- API call will fail because of wrong id, `400 Invalid 'batch_id': 'vs_my-vector-store-id'. Expected an ID that begins with 'vsfb_'."`

### Code snippets

```ts
const fileBatchUpload = await openai.vectorStores.fileBatches.uploadAndPoll(
  'vs_my-vector-store-id'
  {
    files: [fileStream],
  },
);

const files = await openai.vectorStores.fileBatches.listFiles(
  fileBatchUpload.id,
  { vector_store_id: 'vs_my-vector-store-id'},
);
```

### OS

macOS

### Node version

v20.19.0

### Library version

openai v6.8.1

MukundaKatta:
Fixes #1700.

## Summary
- pass request/poll options through vector store file batch create-and-poll helpers
- add a regression test ensuring uploadAndPoll resolves to the completed file batch (`vsfb_...`) object

## Tests
- yarn test tests/api-resources/vector-stores/file-batches.test.ts
- yarn format src/resources/vector-stores/file-batches.ts tests/api-resources/vector-stores/file-batches.test.ts
- git diff --check

### DIFF
--- a/src/resources/vector-stores/file-batches.ts
+++ b/src/resources/vector-stores/file-batches.ts
@@ -71,7 +71,7 @@ export class FileBatches extends APIResource {
     body: FileBatchCreateParams,
     options?: RequestOptions & { pollIntervalMs?: number },
   ): Promise<VectorStoreFileBatch> {
-    const batch = await this.create(vectorStoreId, body);
+    const batch = await this.create(vectorStoreId, body, options);
     return await this.poll(vectorStoreId, batch.id, options);
   }
 
@@ -190,9 +190,13 @@ export class FileBatches extends APIResource {
     // Wait for all processing to complete.
     await allSettledWithThrow(workers);
 
-    return await this.createAndPoll(vectorStoreId, {
-      file_ids: allFileIds,
-    });
+    return await this.createAndPoll(
+      vectorStoreId,
+      {
+        file_ids: allFileIds,
+      },
+      options,
+    );
   }
 }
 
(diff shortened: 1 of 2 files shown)


HAYDEN-OAI:
Thanks for the contribution! The remaining vector-store file-batch request-options propagation issue is already tracked in #1855; the separate return-value issue in #1700 has already been addressed. I’m closing this as a duplicate so we can keep discussion of the remaining bug in #1855.

PR #1882, opened 2026-05-15
fix(client): omit User-Agent in browser requests

### LINKED ISSUE #1694: req header user-agent is not a normal value in safari

### Describe the bug

use safari , in chat/completions API , the header  User-Agent  is D/JS 4.104.0,

then it  spawn a error : Request header field User-Agent is not allowed by Access-CControl-Allow-Headers

### To Reproduce

write something , and fetch  chat/completions

### Code snippets

```JavaScript

```

### OS

macOS

### Node version

Node v22.16.0

### Library version

4.104.0

MukundaKatta:
Fixes #1694.

## Summary
- omit the SDK-generated `User-Agent` header when the client is running in a browser-like environment
- keep the existing `User-Agent` behavior for Node, Deno, edge, and other non-browser runtimes
- add a regression test for `dangerouslyAllowBrowser` requests so browser usage does not send a forbidden header

## Tests
- `./node_modules/.bin/jest tests/index.test.ts --runInBand -t "omits forbidden User-Agent"`
- `PATH=/Users/ubl/.cache/codex-runtimes/codex-primary-runtime/dependencies/node/bin:$PATH ./scripts/build`

### DIFF
--- a/src/client.ts
+++ b/src/client.ts
@@ -1105,7 +1105,7 @@ export class OpenAI {
       idempotencyHeaders,
       {
         Accept: 'application/json',
-        'User-Agent': this.getUserAgent(),
+        ...(!isRunningInBrowser() ? { 'User-Agent': this.getUserAgent() } : undefined),
         'X-Stainless-Retry-Count': String(retryCount),
         ...(options.timeout ? { 'X-Stainless-Timeout': String(Math.trunc(options.timeout / 1000)) } : {}),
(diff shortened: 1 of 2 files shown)


HAYDEN-OAI:
Resolved upstream in https://github.com/openai/openai/pull/1277053; the fix will be available in the next SDK release. Closing this downstream implementation as superseded.

PR #1885, opened 2026-05-17
fix: validate workload identity access tokens

pragnyanramtha:
## Summary

- validate successful workload identity token exchange responses before caching them
- throw an `OpenAIError` when a 200 response omits or returns an empty `access_token`
- preserve the existing default `expires_in` behavior while respecting explicit `0` with nullish coalescing

## Why

A malformed token exchange success response was previously trusted and cached. If `access_token` was absent, `getToken()` could return `undefined`, allowing downstream requests to construct an invalid bearer token instead of failing at the authentication boundary.

## Validation

- `yarn test tests/auth/workload-identity-auth.test.ts --runInBand`
- `yarn test tests/lib/workload-identity.test.ts --runInBand`
- `./node_modules/.bin/prettier --check src/auth/workload-identity-auth.ts tests/auth/workload-identity-auth.test.ts`
- `./node_modules/.bin/eslint src/auth/workload-identity-auth.ts tests/auth/workload-identity-auth.test.ts`
- `./scripts/build`
- `git diff --check`


HAYDEN-OAI (approved):

HAYDEN-OAI:
@pragnyanramtha thanks again for catching this and tightening the auth boundary so malformed token responses fail cleanly. really appreciate the careful work here. merging now! 💚

PR #1893, opened 2026-05-21
fix(client): remove abort listener on success in fetchWithTimeout

LeSingh1:
Fixes #1811.

`fetchWithTimeout` attaches an abort listener to the caller's `signal` that forwards to its internal controller. The listener is added with `{ once: true }`, so it self-removes when the signal actually fires, but it is never removed when the request completes successfully.

On Node this is harmless because `AbortSignal.timeout()` keeps its timer unref'd regardless of listeners. On Deno, attaching a listener to a signal returned by `AbortSignal.timeout()` refs the underlying timer. The orphaned listener keeps that timer alive for the full timeout duration, so a request that completes in 500ms blocks process exit for the remainder of a caller's 30s timeout.

The fix removes the listener in the existing `finally` block alongside `clearTimeout`. The `abort` reference is already hoisted into the enclosing scope, so no other changes are needed.

### Repro (from the linked issue)

```ts
import OpenAI from "openai";

const client = new OpenAI();
const signal = AbortSignal.timeout(30000);

const response = await client.chat.completions.create(
  { model: "gpt-4o", messages: [{ role: "user", content: "Say hi" }], max_tokens: 10 },
  { signal },
);

console.log("Result:", response.choices[0].message.content);
// Before: Deno hangs ~30s after logging. After: exits immediately.
```

### Tests

Added a regression test in `tests/index.test.ts` that spies on `addEventListener` / `removeEventListener` of a caller-supplied `AbortSignal` and asserts the same listener reference is both added and removed once the request resolves. Full `tests/index.test.ts` suite (61 tests) passes against the mock server.

### DIFF
--- a/src/client.ts
+++ b/src/client.ts
@@ -969,6 +969,12 @@ export class OpenAI {
       return await this.fetch.call(undefined, url, fetchOptions);
     } finally {
       clearTimeout(timeout);
+      // Remove the listener on success so we don't keep the caller's signal
+      // (and any underlying timer) ref'd for the full timeout duration.
+      // The `{ once: true }` option only self-removes when the signal actually
+      // fires, which on Deno keeps `AbortSignal.timeout()` timers ref'd and
+      // blocks process exit until the timeout elapses. See #1811.
+      if (signal) signal.removeEventListener('abort', abort);
     }
   }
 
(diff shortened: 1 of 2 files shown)


HAYDEN-OAI:

PR #1897, opened 2026-05-27
fix(error): preserve unwrapped API error bodies

### LINKED ISSUE #1577: Support raw error responses

With the current version its is possible to have [undocumented request parameters and response parameters](https://github.com/openai/openai-node?tab=readme-ov-file#making-customundocumented-requests). But if one also returns undocumente errors which are not matching the openai layout the response is basically not accessible. 

It would be nice if the error could contain a `.raw` property which allows to access the original body and headers. 

### Additional context

e.g. this would be possible: 

````js
try {
  new OpenAI({...}).chat.completions.create({
        model,
        messages,
        stream: true
      });
} catch (error) {
  // Access undocumented properties, e.g. RFC specified error properties
  console.log(error.detail)
}
```

vixco:

## Changes being requested

Preserve JSON error response bodies that do not use the OpenAI { error: ... } wrapper. Today those responses produce an APIError with error === undefined even though the response body was valid JSON. This keeps the existing behavior for wrapped OpenAI errors, while falling back to the full parsed body when the error key is absent.

## Additional context & links

Fixes #1577.

Tests:
- ./node_modules/.bin/jest.cmd tests/index.test.ts
- ./node_modules/.bin/eslint.cmd src/core/error.ts tests/index.test.ts

Note: 
ode ./node_modules/typescript/bin/tsc currently fails in this local checkout because example files import optional packages that are not installed here (@azure/identity, express, 
ext).

### DIFF
--- a/src/core/error.ts
+++ b/src/core/error.ts
@@ -67,7 +67,10 @@ export class APIError<
       return new APIConnectionError({ message, cause: castToError(errorResponse) });
     }
 
-    const error = (errorResponse as Record<string, any>)?.['error'];
+    const error =
+      errorResponse && typeof errorResponse === 'object' && 'error' in errorResponse ?
+        (errorResponse as Record<string, any>)['error']
+      : errorResponse;
 
     if (status === 400) {
       return new BadRequestError(status, error, message, headers);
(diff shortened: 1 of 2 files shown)


HAYDEN-OAI:
Thanks for the contribution! Preserving error responses that do not contain a nested error field is already tracked in #1746, so I’m closing this as a duplicate and keeping the discussion there.

PR #1899, opened 2026-05-27
feat(api): Add Amazon Bedrock Responses support

jim-openai:
## Summary
- add first-class `BedrockOpenAI` provider client
- support Bedrock base URL, region, static bearer token, and refreshable token-provider config
- keep Bedrock support thin: SDK adapts auth and base URL, while AWS owns endpoint and feature validation
- add focused tests, docs, and an example

## Review Notes
- This rewrite removes the earlier SDK-side endpoint/tool/include/websocket guards. Formerly guarded calls now send requests and surface normal AWS HTTP errors through the standard SDK error pipeline.
- Bedrock does not implicitly reuse `OPENAI_API_KEY`; it reads `AWS_BEARER_TOKEN_BEDROCK` unless the caller passes `apiKey` or `bedrockTokenProvider`.
- Generated admin-security routes reuse the Bedrock bearer token so they pass through to AWS instead of failing SDK auth resolution before the request.
- Kept regional Mantle endpoint derivation for now; no global endpoint behavior is added before AWS defines it.
- Deliberately deferred a separate Bedrock auth/config struct. Inline config keeps this PR small and self-contained; a dedicated struct may age better if AWS adds more auth or endpoint modes.
- Kept one Bedrock-local compatibility repair for `responses.stream().finalResponse().output_text`; a future core cleanup could make that generic instead of Bedrock-specific.

## Testing
- focused Node Bedrock tests: `17 passed`
- Prettier write/check on touched provider docs/source/test/example files and touched `README.md`
- ESLint on touched Bedrock source/test/example files
- Node build passed
- live `us-east-2` probes with refreshed Bedrock credentials:
  - earlier same-day `openai.gpt-5.5` pass completed static bearer-token SSE and two token-provider `responses.create` calls
  - current recheck confirmed `/v1/models` returns `200` and lists `openai.gpt-5.4` / `openai.gpt-5.5`
  - current direct `/openai/v1/responses` and SDK `responses.create` probes return Bedrock `500 internal_server_error` for both `openai.gpt-5.4` and `openai.gpt-5.5`; SDKs surface normal `InternalServerError` with request ids, so no SDK workaround was added
  - `ResponsesWS` still passes through and AWS rejects the websocket handshake with HTTP `405`

### DIFF
--- a/src/bedrock.ts
+++ b/src/bedrock.ts
@@ -1,0 +1,193 @@
+import * as Errors from './error';
+import { OpenAI } from './client';
+import type { ApiKeySetter, ClientOptions } from './client';
+import type { NullableHeaders } from './internal/headers';
+import { buildHeaders } from './internal/headers';
+import type { FinalRequestOptions, RequestOptions } from './internal/request-options';
+import { readEnv } from './internal/utils';
+import { addOutputText } from './lib/ResponsesParser';
+import type { ResponseStreamParams } from './lib/responses/ResponseStream';
+import * as API from './resources/index';
+import type * as ResponsesAPI from './resources/responses/responses';
+
+export interface BedrockClientOptions extends Omit<
+  ClientOptions,
+  'apiKey' | 'adminAPIKey' | 'baseURL' | 'workloadIdentity'
+> {
+  /**
+   * Bedrock bearer token used for authentication.
+   *
+   * Defaults to process.env['AWS_BEARER_TOKEN_BEDROCK'].
+   */
+  apiKey?: string | null | undefined;
+
+  /**
+   * Bedrock API root.
+   *
+   * Defaults to process.env['AWS_BEDROCK_BASE_URL'], or derives
+   * `https://bedrock-mantle.<region>.api.aws/openai/v1` from `awsRegion`,
+   * process.env['AWS_REGION'], or process.env['AWS_DEFAULT_REGION'].
+   */
+  baseURL?: string | null | undefined;
+
+  /**
+   * BedrockOpenAI only supports Bedrock bearer token authentication.
+   */
+  adminAPIKey?: never;
+
+  /**
+   * BedrockOpenAI only supports Bedrock bearer token authentication.
+   */
+  workloadIdentity?: never;
+
+  /**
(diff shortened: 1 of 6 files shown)


src/bedrock.ts
```diff
@@ -0,0 +1,382 @@
...
+import { addOutputText } from './lib/ResponsesParser';
+import * as API from './resources/index';
+import * as ResponsesAPI from './resources/responses/responses';
+
+const BEDROCK_RESPONSES_ONLY_ERROR =
+  'Amazon Bedrock support in the OpenAI SDK only supports the Responses API. Use `client.responses` with `BedrockOpenAI` instead.';
+
+const BEDROCK_CAPABILITIES = Object.freeze({
+  responsesWebsocket: false,
+  unsupportedToolTypes: new Set([
```
apcha-oai: Similar here to python, would strongly consider letting this validation live server-side similar to azure.

If we want some validation can at least validate high level endpoints but this is a little fragile/will need to be updated constantly

src/bedrock.ts
```diff
@@ -0,0 +1,382 @@
...
+    'code_interpreter_call.',
+    'computer_call_output.',
+    'file_search_call.',
+    'web_search_call.',
+  ],
+});
+
+export interface BedrockClientOptions extends Omit<
+  ClientOptions,
+  'apiKey' | 'adminAPIKey' | 'baseURL' | 'workloadIdentity'
```
apcha-oai: nice

apcha-oai:
(also just note will need to run prettier)

jim-openai:
Pushed simplified version

apcha-oai (dismissed):
Looks a lot better, thanks for fixes! Minor note: just need to quick run prettier

jim-openai:
Updated after follow-up: ran Prettier and kept its touched README formatting; reran focused tests, ESLint, Prettier check, and build. Apologies that I could not complete the planned fresh manual Responses recheck: Bedrock `/openai/v1/responses` was returning provider-side `500 internal_server_error` for both `openai.gpt-5.4` and `openai.gpt-5.5`, reproduced with direct curl and confirmed by others, so I did not add an SDK workaround. The PR body has the current manual probe details.

PR #1908, opened 2026-06-01
fix(client): apply timeout to response body reads

### LINKED ISSUE #1825: SDK `timeout` option does not cover response-body read phase (can hang forever)

redactdeveloper:
Summary
- keep request timeouts active during stalled response body reads, not only until headers arrive
- start the body timeout lazily when the response body is actually consumed, so asResponse() callers do not leave active timers behind
- release the response body reader when the monitored stream closes, errors, or is cancelled
- add regression coverage for a response that sends headers and then stalls before completing the body

Tests
- ./scripts/test tests/index.test.ts
- ./node_modules/.bin/jest tests/index.test.ts --runInBand --detectOpenHandles
- ./node_modules/.bin/prettier --check src/client.ts tests/index.test.ts
- ./node_modules/.bin/tsc --noEmit --skipLibCheck --target es2020 --lib es2020,dom,dom.iterable --module commonjs --moduleResolution node --esModuleInterop --strict --exactOptionalPropertyTypes --noUncheckedIndexedAccess --noImplicitOverride --noPropertyAccessFromIndexSignature src/client.ts

Fixes #1825

### DIFF
--- a/src/client.ts
+++ b/src/client.ts
@@ -949,7 +949,28 @@ export class OpenAI {
     const abort = this._makeAbort(controller);
     if (signal) signal.addEventListener('abort', abort, { once: true });
 
-    const timeout = setTimeout(abort, ms);
+    let timeout: ReturnType<typeof setTimeout> | undefined = setTimeout(abort, ms);
+    let timeoutCleared = false;
+    // Re-arm the timer so that the timeout applies to inactivity during the
+    // body-read phase as well, not just to header arrival. A stalled body read
+    // (headers received, then the server stops sending) is aborted after `ms`
+    // of inactivity instead of hanging forever, while a steadily-streaming
+    // response keeps resetting the timer and is never cut off.
+    // See https://github.com/openai/openai-node/issues/1825
+    const rearmTimeout = () => {
+      if (timeoutCleared) return;
+      if (timeout) clearTimeout(timeout);
+      timeout = setTimeout(abort, ms);
+    };
+    const clearCurrentTimeout = () => {
+      if (timeout) clearTimeout(timeout);
+      timeout = undefined;
+    };
+    const clearTimeoutOnce = () => {
+      if (timeoutCleared) return;
+      timeoutCleared = true;
+      clearCurrentTimeout();
+    };
 
     const isReadableBody =
       ((globalThis as any).ReadableStream && options.body instanceof (globalThis as any).ReadableStream) ||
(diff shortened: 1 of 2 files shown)


HAYDEN-OAI:
Thanks for investigating stalled response bodies! A complete fix needs coordinated SDK response parsing, timeout classification, and retry handling; the current implementation bypasses retries after headers arrive and also needs to preserve custom fetch compatibility. I’m moving this to SDK triage so the complete response lifecycle can be reviewed together.

HAYDEN-OAI:
Thanks for investigating this; stalled response bodies are a real bug. I traced the current request lifecycle to the shared Castiron TypeScript client template, which clears its timeout as soon as fetch resolves and owns timeout classification and retries. Fixing this safely therefore requires coordinated upstream changes across generated request retries, response parsing, and streaming rather than an SDK-local response wrapper. The proposed wrapper also bypasses APIConnectionTimeoutError/retries after headers, eagerly consumes raw responses, breaks custom fetch bodies without getReader(), and can silently truncate SSE streams. I am marking this upstream so the complete fix can be made in Castiron and propagated safely.

HAYDEN-OAI:
Thanks for investigating this. The canonical response-body timeout fix has now merged upstream in the shared TypeScript client generator, including timeout classification and retry handling while preserving raw responses, streaming, and custom fetch behavior. Because src/client.ts is generated, closing this SDK-local proposal as superseded. The fix will be available after SDK regeneration and release.

PR #1913, opened 2026-06-02
Avoid rescanning SSE delimiter buffers

camdencheek:

## Changes being requested

Modify `iterSSEChunks` to skip already-scanned bytes so streaming in small chunks does not cause `findDoubleNewlineIndex` to get called with more data than is necessary.

## Additional context & links

In many of our profiles, `findDoubleNewlineIndex` was consuming a large amount of CPU. In some cases, even pushing our event loop utilization to 100%. This was surprising to me since that should be a pretty cheap operation. 

It seems the issue is not `findDoubleNewlineIndex` itself, but rather that it is re-scanning bytes that it has already checked. Whenever a network chunk comes in, it gets appended to `data` then that _full_ buffer is searched for a double newline. This means that, in the pathological case of streaming one byte at a time, finding the double newline scans `1 + 2 + ... + n = n * (n + 1) / 2` bytes. 

When we measured this in our service, roughly 95% of the bytes scanned were redundantly scanned.

This fixes the issue by skipping the prefix we know can't contain any double newline (or partial double newline) and only scanning new bytes. We've patched this on our end, and it dropped byte scans here by 95% in our production env (as expected). 

screenshot_2026-06-02_09-27-42


Mamalimashayekh (approved):

HAYDEN-OAI:
Thanks for the contribution! This optimizes the SDK’s server-sent-event decoder rather than changing generated API types. I’m moving it to SDK triage so the streaming runtime and regression coverage can be reviewed together.

HAYDEN-OAI (approved):
Verified the bounded three-byte SSE delimiter overlap against fragmented LF/CR/CRLF events, immediate event delivery, multiple buffered records, and an instrumented scan-complexity regression. Full handwritten and generated tests, TypeScript, lint, build, streaming benchmarks, and all required CI checks pass.

PR #1918, opened 2026-06-04
fix(azure): preserve model-to-deployment mapping for images.edit

### LINKED ISSUE #1917: Azure: images.edit ignores per-request model, requests go to wrong URL (404)

### Describe the bug

When using `AzureOpenAI` without a client-level `deployment` and passing the deployment via the per-request `model` param (the documented pattern, which works for chat completions, embeddings, audio transcriptions/translations and `images.generate`), `images.edit()` does not get its path rewritten to `/deployments/{model}/images/edits`. The request is sent to `https://{endpoint}/openai/images/edits`, which returns 404 on Azure.

Root cause: `AzureOpenAI.buildRequest` resolves the deployment with

```ts
const model = this.deploymentName || options.body['model'] || options.__metadata?.['model'];
```

`/images/edits` is a multipart endpoint, so by the time `buildRequest` runs, `options.body` is already a `FormData` instance — `options.body['model']` is `undefined`. The other multipart endpoints handle this by passing the model through request metadata (see `audio/transcriptions.ts` and `audio/translations.ts`, which both set `__metadata: { model: body.model }`), but the `images.edit` method does not, so the fallback is also `undefined` and the path rewrite is skipped.

Note `images.generate` is unaffected because its body stays a plain JSON object. `images.edit` is the only deployments-routed endpoint (`_deployments_endpoints` includes `/images/edits`) with this gap, which is probably why it went unnoticed — the Azure test suite covers model→deployment mapping for generate/transcriptions/translations but has no `images.edit` case.

### To Reproduce

```ts

const client = new AzureOpenAI({
  endpoint: 'https://my-resource.openai.azure.com',
  apiKey: '...',
  apiVersion: '2025-04-01-preview',
  // no client-level `deployment`
});

await client.images.edit({
  model: 'my-gpt-image-deployment', // per-request deployment
  image: fs.createReadStream('input.png'),
  prompt: 'add a hat',
});
// → POST https://my-resource.openai.azure.com/openai/images/edits?api-version=...
// → 404 (missing /deployments/my-gpt-image-deployment/)
```

Expected: `POST https://my-resource.openai.azure.com/openai/deployments/my-gpt-image-deployment/images/edits?api-version=...`

Workaround: instantiate the client with a client-level `deployment` — but that forces a separate client instance per deployment.

### OS

macOS / Linux

### Node version

Node v26.0.0

### Library version

openai v6.42.0 (current master)

serhiizghama:
## Problem

Fixes #1917

On Azure, `images.edit()` ignores the per-request `model` param for deployment routing. With no client-level `deployment`, the request goes to `/openai/images/edits` instead of `/openai/deployments/{model}/images/edits` and Azure returns 404. Every other deployments-routed endpoint (chat completions, embeddings, `images.generate`, audio transcriptions/translations) handles this correctly.

Root cause: `AzureOpenAI.buildRequest` resolves the deployment via

```ts
const model = this.deploymentName || options.body['model'] || options.__metadata?.['model'];
```

`/images/edits` is multipart, so `options.body` is already a `FormData` when `buildRequest` runs and `options.body['model']` is `undefined`. The multipart audio endpoints solve exactly this by passing `__metadata: { model: body.model }` — `images.edit` is the only entry in `_deployments_endpoints` with a multipart body that doesn't.

## Solution

Pass `__metadata: { model: body.model }` in `images.edit`, mirroring the existing pattern in `audio/transcriptions.ts` and `audio/translations.ts`. No behavior change for non-Azure clients (`__metadata` is only read by `AzureOpenAI.buildRequest`).

Since this file is generated, the underlying fix may need to land in the Stainless config — the change here reproduces exactly what the generator already emits for the audio multipart endpoints, so it should be a straightforward config mirror.

## Testing

Added `handles image edit` to the existing "with no client-level deployment" suite in `tests/lib/azure.test.ts` (the suite covered `images.generate` but not `images.edit`, which is how this slipped through). The test fails on master with:

```
- "url": "https://example.com/openai/deployments/deployment/images/edits?api-version=..."
+ "url": "https://example.com/openai/images/edits?api-version=..."
```

and passes with the fix. Full offline suite (`jest --testPathIgnorePatterns tests/api-resources`): 567/567 passing; lint, prettier and `./scripts/build` clean.


serhiizghama:
Bumping this one — the Azure model→deployment mapping for `images.edit` is still broken on main, and the PR's clean and mergeable. Happy to adjust the approach if that helps get it in.

HAYDEN-OAI:
Thanks for the contribution! This restores Azure deployment routing for multipart image edits by carrying the selected model through the SDK’s request metadata, matching the existing audio-upload behavior. I’m moving it to SDK/Azure triage so the routing change and regression test can be reviewed together.

HAYDEN-OAI (approved):
Verified Azure image-edit deployment routing for buffered and streaming uploads, configured deployment precedence, existing metadata preservation, omitted-model fallback, and nullable models. Full handwritten/generated suites, TypeScript, lint, build, and all required Node 22/24/26 CI checks pass.

PR #1922, opened 2026-06-05
feat: add token estimation utility for chat completions

### LINKED ISSUE #318: Estimate token usage + dry run option

There a couple nodejs tokenizers laying around npm land:
- [gpt-tokenizer](https://github.com/niieani/gpt-tokenizer)
- [tiktoken](https://github.com/dqbd/tiktoken)

However, they have to be kept maintained and up to date with the official [tiktoken](https://github.com/openai/tiktoken). It feels very natural that such a tool be apart of this repository and updated regularly by the openai team. I feel that pre-request token estimation is vital to ensuring hobbyist developers don't unknowingly succumb to financial demise. 

Why apart of this library?
Take [function calling](https://platform.openai.com/docs/guides/gpt/function-calling) for example:
```
    const response = await openai.chat.completions.create({
        model: "gpt-3.5-turbo",
        messages: messages,
        functions: functions,
        function_call: "auto",  // auto is default, but we'll be explicit
    });
```
I could intercept this function call above to view what data is sent to chatGPT so I can figure out how many tokens this call takes with the `message` + the `functions`, pipe them into gpt-tokenizer and see how many tokens are used. But if it were to be built into the API itself then you can have a dry run option on your nodejs api functions that estimate cost and then the developer could determine in the logic if they want to proceed with the actual run.

What else could be very helpful would be to include some parameter whether in the configuration or when calling api functions such that if the tokens and/or usd cost exceeds a certain limit the call auto fails and doesn't actually get executed.

### Additional context

_No response_

Oxygen56:
Adds token counting utility for cost estimation and dry-run support.

Closes #318

### DIFF
--- a/src/index.ts
+++ b/src/index.ts
@@ -27,3 +27,5 @@ export {
 
 export { AzureOpenAI } from './azure';
 export { BedrockOpenAI, type BedrockClientOptions } from './bedrock';
+
+export { estimateTokens, type EstimateTokensParams, type TokenEstimate } from './lib/token-estimation';
--- a/src/lib/token-estimation.ts
+++ b/src/lib/token-estimation.ts
@@ -158,6 +158,17 @@ function estimateMessageContent(content: unknown): number {
         } else if ('input_audio' in part) {
           // Audio is tokenized separately; rough estimate
           tokens += 50;
+        } else if ('file' in part && part.file && typeof part.file === 'object') {
+          const file = part.file as { file_data?: string; file_id?: string; filename?: string };
+          if (typeof file.file_data === 'string') {
+            tokens += estimateStringTokens(file.file_data);
+          }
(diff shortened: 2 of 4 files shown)


Oxygen56:
Addressed the Codex review feedback in a118358e: file content parts now contribute to input token estimates, and context-window lookup no longer uses broad prefix matching for unknown future model IDs. Validation: `npx jest tests/lib/token-estimation.test.ts --runInBand` and `git diff --check`.

HAYDEN-OAI:
Thanks for the thoughtful contribution! A character-count heuristic and manually maintained model limits cannot provide reliable token, cost, or context-window estimates across models. We don’t want to expose potentially misleading estimates as an SDK API, so we’re closing this approach while retaining the broader feature request in #318.

PR #1924, opened 2026-06-05
fix: restore 60-second cap on Retry-After header values

### LINKED ISSUE #1840: Retry-After 60-second cap silently removed in chore commit, was this intentional?

Oxygen56:
Restores the 60-second maximum on Retry-After header values.

Fixes #1840

### DIFF
--- a/src/client.ts
+++ b/src/client.ts
@@ -1026,9 +1026,9 @@ export class OpenAI {
       }
     }
 
-    // If the API asks us to wait a certain amount of time, just do what it
-    // says, but otherwise calculate a default
-    if (timeoutMillis === undefined) {
+    // If the API asks us to wait a certain amount of time (capped at 60 seconds),
+    // just do what it says, but otherwise calculate a default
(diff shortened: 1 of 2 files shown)


Oxygen56:
Addressed the Codex review feedback in d0eaae6: Retry-After values exactly at the 60-second cap are now honored, including both `Retry-After: 60` and `Retry-After-Ms: 60000`. Validation: `npx jest tests/index.test.ts -t "60-second cap" --runInBand` and `git diff --check`.

HAYDEN-OAI:
Thanks for the contribution! The Retry-After delay limit is already tracked in #1841 and issue #1840, so I’m closing this as a duplicate and keeping the discussion there.

PR #1925, opened 2026-06-05
fix: add gpt-5.1-codex-mini model to type definitions

### LINKED ISSUE #1706: 'gpt-5.1-mini' model does not work.

### Describe the bug

version : 6.9.0

It seems that `gpt-5.1-mini` is not working properly. 

I think it might have been mistakenly labeled as `gpt-5.1-mini` instead of `gpt-5.1-codex-mini`.

- https://github.com/openai/openai-node/blob/master/src/resources/shared.ts#L24
- https://platform.openai.com/docs/models/gpt-5.1-codex-mini

### To Reproduce

```
import OpenAI from "openai";
const client = new OpenAI();

  model: "gpt-5.1-mini",
  input: "Write a short bedtime story about a unicorn.",
});

console.log(response.output_text);
```

### Code snippets

```JavaScript

```

### OS

macOs

### Node version

Node v20.19.3

### Library version

openai v6.9.0

Oxygen56:
Adds support for the gpt-5.1-codex-mini model.

Fixes #1706

### DIFF
--- a/src/resources/responses/responses.ts
+++ b/src/resources/responses/responses.ts
@@ -8199,7 +8199,7 @@ export interface ResponseCompactParams {
     | 'gpt-5.1'
     | 'gpt-5.1-2025-11-13'
     | 'gpt-5.1-codex'
-    | 'gpt-5.1-mini'
+    | 'gpt-5.1-codex-mini'
     | 'gpt-5.1-chat-latest'
     | 'gpt-5'
     | 'gpt-5-mini'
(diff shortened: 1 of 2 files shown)


Oxygen56:
Addressed the Codex review feedback in 3a1b10e: `gpt-5.1-mini` is retained and `gpt-5.1-codex-mini` is added alongside it in both model literal lists. Validation: `git diff --check`.

HAYDEN-OAI:
Thanks for the contribution! gpt-5.1-codex-mini was retired on July 23, 2026 and is no longer available. The recommended replacement is gpt-5.6-terra, so we’re not adding the retired model to current SDK types. See https://platform.openai.com/docs/deprecations.

PR #1932, opened 2026-06-10
fix(responses): accumulate response.refusal.delta in ResponseStream

yen0304:
## Summary

`ResponseStream#accumulateResponse` has explicit cases for accumulating streaming deltas into the response snapshot — but `response.refusal.delta` was never handled. This means:

- `response.refusal.delta` events are emitted to subscribers correctly (via the default `maybeEmit` path in `#addEvent`), but
- The in-memory snapshot's `content[n].refusal` is never updated — it stays `""` for the entire stream.

Any code that reads `finalResponse().output[0].content[0].refusal` after streaming gets an empty string instead of the model's actual refusal text.

**Root cause:** The switch in `#accumulateResponse` handles `output_text.delta`, `function_call_arguments.delta`, and `reasoning_text.delta`, but has no case for `refusal.delta`.

## Changes

- `src/lib/responses/ResponseStream.ts`: Added `response.refusal.delta` case to `#accumulateResponse` — appends `event.delta` to `content.refusal`, matching the exact pattern used by the other delta handlers.
- `tests/lib/ResponseStream.test.ts`: New test `'refusal accumulates correctly'` — sends three refusal deltas and asserts both that the `response.refusal.delta` events fire with the right values and that `finalResponse()` contains the fully concatenated refusal string.
- `tests/lib/__snapshots__/ResponseStream.test.ts.snap`: Snapshot for the new test.

## Test plan

- [x] New test added covering the fix (`'refusal accumulates correctly'`)
- [x] All existing tests pass (`standard text works`, `reasoning works`)
- [x] Verified manually with a standalone driver: without the fix, `content.refusal` stays `""` after 3 deltas; with the fix it reads `"I cannot help with that."`

### DIFF
--- a/src/lib/responses/ResponseStream.ts
+++ b/src/lib/responses/ResponseStream.ts
@@ -10,7 +10,11 @@ import { RequestOptions } from '../../internal/request-options';
 import { APIUserAbortError, OpenAIError } from '../../error';
 import OpenAI from '../../index';
 import { type BaseEvents, EventStream } from '../EventStream';
-import { type ResponseFunctionCallArgumentsDeltaEvent, type ResponseTextDeltaEvent } from './EventTypes';
+import {
+  type ResponseFunctionCallArgumentsDeltaEvent,
+  type ResponseRefusalDeltaEvent,
+  type ResponseTextDeltaEvent,
+} from './EventTypes';
 import { maybeParseResponse, ParseableToolsParams } from '../ResponsesParser';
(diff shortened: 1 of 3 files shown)


HAYDEN-OAI:
@yen0304 thanks so much for catching this and for the clear write-up! this fix has now landed as part of #1793, including refusal delta coverage, and you're credited on the implementation there. closing this one as superseded. really appreciate the contribution and your patience here!

PR #1933, opened 2026-06-10
fix(internal): lowercase header names with locale-pinned `en-US` rule

### LINKED ISSUE #1928: Bug: TypeError: Header name must be a valid HTTP token ["openaı-organization"] on Turkish Windows/Locale

- [ ] This is an issue with the Node library

### Bug Description
The SDK throws a fatal error `Header name must be a valid HTTP token ["openaı-organization"]` on systems running with a Turkish OS locale (e.g., Windows set to Turkish language).

### Exact Root Cause
The issue stems directly from [src/internal/headers.ts (Line 77)](https://github.com/openai/openai-node/blob/5436f42de7f8ac2aafed4da23951c61e353da9cc/src/internal/headers.ts#L77):
```typescript
const lowerName = name.toLowerCase();
```
When the internal Stainless API engine passes header keys like `"OpenAI-Organization"` or `"OpenAI-Project"`, JavaScript's standard `.toLowerCase()` on a Turkish locale converts the uppercase `I` to a dotless `ı`, turning the keys into `"openaı-organization"` and `"openaı-project"`. Since `ı` is a non-ASCII character, the underlying Node.js network layer (undici/http) fails validation and terminates the process.

### Temporary Workaround (Proven to Work)
As a temporary hotfix, manually modifying the packaged bundle (e.g., `extension.js` in IDE extensions) and changing the literal definitions from camelCase to strict lowercase entirely prevents this bug:
- Change `"OpenAI-Organization"` to `"openai-organization"`
- Change `"OpenAI-Project"` to `"openai-project"`
By declaring them lowercase from the start, we bypass the need for runtime normalization which triggers the Turkish I-problem.

### Suggested Official Fix
To ensure locale-invariant header normalization, replace the standard `.toLowerCase()` with a locale-safe conversion in `headers.ts`:
```typescript
const lowerName = name.toLocaleLowerCase('en-US');
```

### To Reproduce

1. Set the operating system language/locale to Turkish (tr-TR).
2. Install the openai-node library.
3. Attempt to initialize the client and make an API request using a third-party proxy provider (e.g., OpenRouter) that passes through header builders.
4. The system throws a fatal token validation error due to "openaı-organization".

### Code snippets

```JavaScript

```

### OS

Windows (with Turkish locale / tr-TR)

### Node version

Managed internally by Continue IDE extension

### Library version

Latest version bundled inside Continue IDE extension

tsushanth:
## Why

Closes #1928.

\`String.prototype.toLowerCase()\` applies the host locale to certain Unicode mappings. On a Turkish process locale the uppercase \`I\` in \`OpenAI-Organization\` becomes the dotless \`ı\`, producing \`openaı-organization\`, which fails Node's HTTP token validation (\`TypeError: Header name must be a valid HTTP token\`) and tears down the request before it is ever sent. The reporter traced this all the way to \`src/internal/headers.ts:77\` and verified the workaround.

I also looked at the other call sites where this SDK lowercases header names for matching, and found the redactor in \`src/internal/utils/log.ts\` does the same thing on the same hot path. There the consequence is more subtle but more serious: an \`Authorization\` header arriving as \`authorızatıon\` after the Turkish casing rule **never matches** the redaction list, so the bearer token would be written verbatim into the logged request representation. Worth fixing in the same PR since it's the same root cause and a security-shaped exposure.

## What

- \`src/internal/headers.ts\`: \`name.toLowerCase()\` → \`name.toLocaleLowerCase('en-US')\` in \`buildHeaders\`, with a comment linking back to this issue.
- \`src/internal/utils/log.ts\`: pre-compute \`lowerName = name.toLocaleLowerCase('en-US')\` once and use that for the redaction comparisons. (Also removes the five repeated \`.toLowerCase()\` calls.) Same comment + issue link.

HTTP header names are RFC 9110 ASCII so anchoring on en-US is the correct invariant.

## Tests

\`tests/buildHeaders.test.ts\` — new \`lowercases header names with ASCII rules even under a Turkish process locale\` test. Monkey-patches \`String.prototype.toLowerCase\` for the duration of the test to mimic the Turkish casing rule (\`I\` → \`ı\`), then verifies the canonical key in the resulting \`NullableHeaders\` is \`openai-organization\` and ASCII-only.

\`yarn jest tests/buildHeaders.test.ts\` → 11 passed. \`yarn jest tests/log.test.ts\` → 4 passed.

## Stainless note

This change is in \`src/internal/\`, which is generated by Stainless. I considered whether the right path was to wait for the upstream codegen change. Felt it was worth proposing the diff directly anyway because (a) the bug currently corrupts requests / leaks credentials in production on a non-trivial install base, (b) the fix is mechanical and easy to mirror in the generator template, and (c) precedent from openai/anthropic SDK repos suggests external PRs to generated paths are merged when correctness-shaped.

Happy to defer to whatever process you'd prefer for landing this — let me know if you'd like the change in a different form.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -23,6 +23,8 @@ jobs:
     name: lint
     runs-on: ${{ github.repository == 'stainless-sdks/openai-typescript' && 'depot-ubuntu-24.04' || 'ubuntu-latest' }}
     if: (github.event_name == 'push' || github.event.pull_request.head.repo.fork) && (github.event_name != 'push' || github.event.head_commit.message != 'codegen metadata')
+    permissions:
+      contents: read
     steps:
       - uses: actions/checkout@de0fac2e4500dabe0009e67214ff5f5447ce83dd # v6
 
@@ -89,6 +91,8 @@ jobs:
     timeout-minutes: 10
     runs-on: ${{ github.repository == 'stainless-sdks/openai-typescript' && 'depot-ubuntu-24.04' || 'ubuntu-latest' }}
     if: github.event_name == 'push' || github.event.pull_request.head.repo.fork
+    permissions:
+      contents: read
     strategy:
       fail-fast: false
       matrix:
@@ -128,6 +132,8 @@ jobs:
     runs-on: ${{ github.repository == 'stainless-sdks/openai-typescript' && 'depot-ubuntu-24.04' || 'ubuntu-latest' }}
     if: github.repository == 'openai/openai-node'
     environment: ci
+    permissions:
+      contents: read
 
     steps:
       - uses: actions/checkout@34e114876b0b11c390a56381ad16ebd13914f8d5 # v4
@@ -160,6 +166,8 @@ jobs:
     runs-on: ubuntu-latest
     timeout-minutes: 20
     environment: ci
+    permissions:
+      contents: read
 
     steps:
       - uses: actions/checkout@34e114876b0b11c390a56381ad16ebd13914f8d5 # v4
--- a/.github/workflows/detect-breaking-changes.yml
+++ b/.github/workflows/detect-breaking-changes.yml
@@ -13,6 +13,8 @@ jobs:
     runs-on: 'ubuntu-latest'
     name: detect-breaking-changes
     if: github.repository == 'openai/openai-node'
+    permissions:
+      contents: read
     steps:
       - name: Calculate fetch-depth
         run: |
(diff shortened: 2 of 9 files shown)


tsushanth:
Rebased onto current upstream.

tsushanth:
Rebased onto current upstream.

HAYDEN-OAI:
Reverified during the upstream audit: this was superseded by [#1941](https://github.com/openai/openai-node/pull/1941), merged in [18e8c290](https://github.com/openai/openai-node/commit/18e8c290d422ef1dfb09a4940f2a0b7703691cc1). This PR is already closed; audit result: done.

PR #1934, opened 2026-06-10
chore: Convert repository to pnpm

jbeckwith-oai:
## Summary

- convert the repository from Yarn classic to pnpm 11.5.1
- add pnpm workspace security settings and a pnpm lockfile
- update CI, release workflows, bootstrap scripts, and contributor docs for pnpm
- upgrade `@types/node` to 24.x so the old `undici-types@6.21.0` trust-policy exclusion can be removed
- pin `@typespec/ts-http-runtime` to 0.3.5 to avoid the current registry trust metadata warning

## Notes

The pnpm version is centralized in `package.json#packageManager`; GitHub Actions uses `pnpm/action-setup` without duplicating the version in workflow env.

The remaining `trustPolicyExclude` and `overrides` entries include comments explaining why they are needed.

## Validation

- `npm view pnpm version`
- `pnpm i`
- `pnpm install --lockfile-only --frozen-lockfile`
- `pnpm build`
- `pnpm lint` currently fails only on pre-existing prettier drift outside this change set: `MIGRATION.md`, `README.md`, several `src/` files, and several tests.


CONTRIBUTING.md
```diff
@@ -32,7 +39,7 @@ All files in the `examples/` directory are not modified by the generator and can
 ```sh
 $ chmod +x examples/<your-example>.ts
 # run the example against your api
-$ yarn tsn -T examples/<your-example>.ts
+$ npm run tsn -T examples/<your-example>.ts
```
HAYDEN-OAI: could we forward `-T` after npm's `--` separator here? npm currently consumes it as an unknown cli option, so the documented command invokes ts-node without transpile-only mode. `npm run tsn -- -T examples/<your-example>.ts` would match the corrected shebangs above.

jbeckwith-oai:
Hi @HAYDEN-OAI, I addressed and resolved the review threads, and CI is green at the latest head. Could you please take another look?

.github/workflows/ci.yml
```diff
@@ -105,26 +128,33 @@ jobs:
     runs-on: ${{ github.repository == 'stainless-sdks/openai-typescript' && 'depot-ubuntu-24.04' || 'ubuntu-latest' }}
     if: github.repository == 'openai/openai-node'
     environment: ci
+    env:
+      OPENAI_API_KEY: ${{ secrets.OPENAI_API_KEY }}
```
HAYDEN-OAI: could we keep `OPENAI_API_KEY` scoped to the demo step here? moving it to job-level `env` also exposes it to the setup actions and `pnpm install`, including dependency lifecycle scripts. the existing shell guard already handles a missing key, so restoring the step-level `env` would preserve the narrower credential boundary.

jbeckwith-oai:
Hi @HAYDEN-OAI, I addressed the latest API key scoping feedback and resolved the thread. Could you please review again?

HAYDEN-OAI (approved):
looks good, thanks for tightening the api key scope!

PR #1937, opened 2026-06-11
test: Use CI environment for OpenAI API key workflows

jbeckwith-oai:
## Summary

- Add the `ci` GitHub Environment to the CI jobs that read `secrets.OPENAI_API_KEY`.
- Limit the environment change to the `examples` and `ecosystem_tests` jobs, which are the workflow jobs currently using that secret.
- Update live example/ecosystem chat-completions smoke tests from legacy `gpt-4` to `gpt-4o-mini`, because the CI-scoped API key resolves but does not have access to `gpt-4`.
- Tighten live smoke-test prompts to request the exact expected phrase, avoiding flaky failures from extra conversational text.

## Why

`OPENAI_API_KEY` is now environment-scoped for CI, so jobs that consume it must explicitly run in the `ci` environment. Once the secret was available, the examples job reached the API and failed on model access for `gpt-4`; the live smoke tests now use a broadly available small model while preserving the test intent. Follow-up ecosystem failures showed the new chat model returned valid responses with extra assistant text, so the prompt now asks for the exact expected phrase.

## Validation

- Parsed `.github/workflows/ci.yml` with Ruby YAML loader.
- Ran `git diff --check`.
- Ran Prettier via `npx prettier --write` on the non-ignored touched example file; ecosystem test fixtures are intentionally ignored by repo Prettier config.
- Monitored PR CI and inspected failed logs; confirmed `OPENAI_API_KEY` was injected, fixed `model_not_found` for `gpt-4`, then tightened prompts after valid `gpt-4o-mini` responses contained extra text.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -98,6 +98,7 @@ jobs:
     name: examples
     runs-on: ${{ github.repository == 'stainless-sdks/openai-typescript' && 'depot-ubuntu-24.04' || 'ubuntu-latest' }}
     if: github.repository == 'openai/openai-node'
+    environment: ci
 
     steps:
       - uses: actions/checkout@34e114876b0b11c390a56381ad16ebd13914f8d5 # v4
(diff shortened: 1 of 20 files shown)


apcha-oai (approved):

aolivares85-collab (commented):
``

PR #1938, opened 2026-06-12
[codex] Add AWS-native authentication for Bedrock

HAYDEN-OAI:
## Summary

This adds first-class Amazon Bedrock support through the standard `OpenAI` client and a dedicated AWS entrypoint:

```ts
import OpenAI from 'openai';
import { bedrock } from 'openai/providers/bedrock/aws';

const client = new OpenAI({
  provider: bedrock({ region: 'us-west-2' }),
});
```

The change:

- introduces an internal provider seam so third-party providers can own routing and final request authentication
- adds Bedrock bearer-token and AWS SigV4 authentication
- keeps the existing bearer-only `BedrockOpenAI` compatibility class unchanged
- separates the dependency-free bearer entrypoint from the statically analyzable AWS entrypoint
- packages both entrypoints for CommonJS, ESM, browser bearer use, and JSR

## Provider seam

When `provider` is configured, the provider owns the base URL and authentication. The client rejects conflicting top-level `apiKey`, `adminAPIKey`, `workloadIdentity`, and `baseURL` options and does not inherit ambient OpenAI credentials or headers when switching from the OpenAI route to a provider.

The provider receives the fully serialized request after subclass preparation and before every request attempt. This lets Bedrock sign the exact method, URL, headers, query, and body while refreshing credentials and signatures on retries.

Provider factories capture immutable configuration, while every `OpenAI` client gets fresh runtime state. An opaque provider registry prevents arbitrary JavaScript objects from impersonating providers and uses a `globalThis` symbol plus `WeakMap` so providers work across duplicate CommonJS and ESM module instances without retaining discarded configurations.

`withOptions()` preserves provider-owned headers when cloning the same provider. It drops inherited OpenAI organization, project, and custom headers only when crossing from the core OpenAI route to a provider or switching provider objects.

## Bedrock routing and authentication

The provider uses the regional Mantle endpoint:

```text
https://bedrock-mantle.<region>.api.aws/openai/v1
```

The `/openai/v1` prefix is intentional. AWS confirmed that it is the OpenAI compatibility contract, while similarly named routes under `/v1`, including `models`, can return different response shapes. Requests are signed with the `bedrock-mantle` SigV4 service name.

The AWS entrypoint selects authentication in this order:

1. an explicit bearer token, token provider, static AWS credentials, AWS profile, or credential provider
2. `AWS_BEARER_TOKEN_BEDROCK`
3. the standard AWS credential chain

Explicit bearer and AWS credential modes are mutually exclusive. Passing `apiKey: null` skips an ambient Bedrock bearer token and deliberately selects AWS credentials.

SigV4 signs the exact serialized, replayable request body and runs again for every retry. The provider also:

- validates canonical endpoint and signing-region agreement
- preserves repeated query parameters
- signs the transmitted uppercase HTTP method
- rejects conflicting custom `Authorization` headers
- disables automatic redirects for signed requests
- redacts AWS session tokens from request logs

Request streaming through `UNSIGNED-PAYLOAD` or AWS-chunked signing is intentionally outside this first milestone. Standard JSON requests are replayable, response streaming is unaffected, and Mantle waits for the complete request body before authentication and authorization.

## Packaging

Bearer authentication is available from `openai/providers/bedrock` without AWS dependencies. This keeps normal `openai` imports, the legacy `BedrockOpenAI` class, and browser bearer bundles independent from AWS packages.

AWS credential authentication is available from `openai/providers/bedrock/aws`. It uses normal static imports so Vite, Webpack, Vercel-style file tracing, and other serverless packagers can discover:

- `@aws-sdk/credential-provider-node`
- `@smithy/hash-node`
- `@smithy/signature-v4`

These remain optional peer dependencies for the package as a whole, but are required before importing the `/aws` entrypoint. If one is absent, import fails immediately with the runtime's normal module-not-found error.

The AWS subpath is exported from npm and JSR. The standalone Bedrock guide remains in the repository rather than being added to build distributions, consistent with other provider guides.

## Compatibility

The existing `BedrockOpenAI` class remains bearer-only and available for current applications. New AWS-authenticated applications should use `new OpenAI({ provider: bedrock(...) })` from `openai/providers/bedrock/aws`.

The shared Responses parser also restores the SDK-synthesized `output_text` convenience property when a raw streamed final response omits it. The underlying distinction between raw wire response types and SDK-enriched response types is broader schema work and is not expanded in this PR.

## Testing

The focused suites cover provider ownership, ambient OpenAI configuration isolation, same-provider cloning, request preparation order, retry signing, bearer and AWS credential precedence, the real AWS environment credential chain with a temporary session token, credential refresh, endpoint normalization, region validation, custom authorization rejection, non-replayable body rejection, browser bearer authentication, dependency isolation, compatibility behavior, and a deterministic SigV4 fixture clearly marked with AWS documentation test credentials.

V8 coverage across the four provider implementation modules is:

- 100% statements
- 99.36% branches
- 100% functions
- 100% lines

Live validation remains isolated from the standard test suite under `tests/live/bedrock.live.test.ts`. It requires an explicit safety flag, disables retries, uses `store: false`, and can exercise bearer, environment bearer, default-chain, named-profile, static, and custom-provider authentication:

```sh
BEDROCK_LIVE_TEST=1 \
BEDROCK_LIVE_AUTH=profile \
AWS_PROFILE=my-profile \
AWS_REGION=us-west-2 \
BEDROCK_MODEL=openai.gpt-oss-120b \
pnpm test:live:bedrock
```

Set `BEDROCK_LIVE_STREAM=1` to include a streaming request. `AWS_BEDROCK_BASE_URL` can override the derived endpoint for live endpoint comparisons.

## Attribution

This builds on and substantially expands Jim's initial implementation in [jim-openai/openai-node#1](https://github.com/jim-openai/openai-node/pull/1). Thank you, Jim, for getting the first version moving and letting us take it over.

## Validation

- [x] 61 focused provider tests and 78 provider/compatibility tests
- [x] full 1,092-test Jest suite
- [x] focused V8 coverage for the provider registry and both Bedrock entrypoints
- [x] full lint pipeline: Prettier, ESLint, SDK and Deno builds, TypeScript, Are The Types Wrong?, Publint, and JSR publish dry run
- [x] browser Webpack build with the dependency-free bearer entrypoint
- [x] Vite 7 SSR build with `noExternal` across OpenAI, AWS SDK, and Smithy packages
- [x] Vercel Node File Trace verification of the packed AWS entrypoint and all three direct AWS dependencies
- [x] live static SigV4 validation with model discovery and Responses inference
- [ ] second-region live validation in `us-east-2`


package.json
```diff
@@ -18,6 +18,7 @@
   },
   "scripts": {
     "test": "./scripts/test",
+    "test:live:bedrock": "./node_modules/.bin/jest --config jest.live.config.ts --runInBand tests/live/bedrock.live.test.ts",
```
jbeckwith-oai: going to node_modules here feels a little funky instead of just saying `jest` should be linked right?

package.json
```diff
@@ -28,6 +29,10 @@
   },
   "devDependencies": {
     "@arethetypeswrong/cli": "^0.18.3",
+    "@aws-sdk/credential-provider-node": "3.972.47",
```
jbeckwith-oai: We're not pinning deps elsewhere, so maybe `^` unless we want to make a broader change?

README.md
```diff
@@ -419,19 +421,18 @@ const response = await client.responses.create({
...
-Set `AWS_BEARER_TOKEN_BEDROCK` to an [Amazon Bedrock API key](https://docs.aws.amazon.com/bedrock/latest/userguide/api-keys.html). To refresh tokens yourself, pass a provider instead of `apiKey`:
+Bearer authentication requires no additional packages. For SigV4 authentication, install the optional AWS dependencies:
 
-```ts
-const client = new BedrockOpenAI({
-  awsRegion: 'us-west-2',
-  bedrockTokenProvider: async () => refreshBedrockToken(),
-});
+```bash
+npm install @aws-sdk/credential-provider-node @smithy/hash-node @smithy/signature-v4
```
jbeckwith-oai: what will the runtime error look like if I try to use this stuff but don't have the appropriate peer deps installed?

src/client.ts
```diff
@@ -501,7 +529,18 @@ export class OpenAI {
       organization: this.organization,
       project: this.project,
       webhookSecret: this.webhookSecret,
+    };
+    if (provider) {
+      delete inheritedOptions.apiKey;
+      delete inheritedOptions.adminAPIKey;
+      delete inheritedOptions.workloadIdentity;
+      delete inheritedOptions.baseURL;
```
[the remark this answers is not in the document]
HAYDEN-OAI: good catch, thank you! fixed in 980c99d5. when `withOptions({ provider })` crosses from an OpenAI client to a provider, it now drops inherited `organization`, `project`, and `defaultHeaders` along with OpenAI auth and the base URL. i added a regression test covering ambient `OPENAI_CUSTOM_HEADERS`, org, and project values.

package.json
HAYDEN-OAI: yep, agreed. changed the script to invoke the linked `jest` binary directly, and verified that `pnpm test:live:bedrock --listTests` resolves the isolated live test.

README.md
HAYDEN-OAI: good question. i documented the lazy failure in both the readme and the Bedrock guide. client construction still succeeds, but the first SigV4-signed request throws `Bedrock AWS authentication requires optional AWS dependencies. Run \`npm install @aws-sdk/credential-provider-node @smithy/hash-node @smithy/signature-v4\` and try again.` bearer auth does not load these packages, and the missing-dependency paths are covered by the dependency-mocking tests.

package.json
HAYDEN-OAI: agreed, updated all four new AWS and Smithy dev dependencies to caret ranges while keeping the existing resolved versions in the lockfile.

RowanAldean:
When can we expect this in Pi (such that we can start using OpenAI Models via Bedrock?)

tests/fixtures/bedrock/v1/sigv4.json
```diff
@@ -0,0 +1,22 @@
...
+  "request": {
+    "method": "POST",
+    "url": "https://bedrock-mantle.us-east-1.api.aws/openai/v1/responses",
+    "body": "{\"model\":\"gpt-4o\",\"input\":\"hello\"}",
+    "contentType": "application/json"
+  },
+  "expected": {
+    "date": "20250102T030405Z",
+    "payloadHash": "50329e51ad520f21b77bad0b01999930ff556cd1bf18434701251ba6c9f877bc",
+    "canonicalRequestHash": "1b69b17ef7548a7bf16a6ee749acfd44b7793e04216345dddd6cfaf4c01bfde5",
```
apcha-oai, line 20: would be nice if we could comment / mark these as test creds more specifically lol, found these are the aws documented test creds

examples/bedrock/responses.ts
```diff
@@ -1,15 +1,20 @@
...
 
 // For refreshed Bedrock bearer tokens:
-// const client = new BedrockOpenAI({ awsRegion: 'us-west-2', bedrockTokenProvider: getBedrockToken });
+// const client = new OpenAI({
+//   provider: bedrock({ region: 'us-west-2', tokenProvider: getBedrockToken }),
+// });
 
 async function main() {
   const response = await client.responses.create({
-    model: 'openai.gpt-5.4',
```
apcha-oai, line 12: nit: put back to 5.4/5.5

README.md
```diff
@@ -403,35 +403,44 @@ console.log(result.choices[0]!.message?.content);
...
 
-// gets the bearer token from AWS_BEARER_TOKEN_BEDROCK and the region from AWS_REGION/AWS_DEFAULT_REGION
-const client = new BedrockOpenAI();
+const client = new OpenAI({
+  provider: bedrock({ region: 'us-west-2' }),
+});
 
 const response = await client.responses.create({
-  model: 'openai.gpt-5.4',
+  model: 'openai.gpt-oss-120b',
```
apcha-oai: same here

src/lib/ResponsesParser.ts
```diff
@@ -55,6 +55,10 @@ export function maybeParseResponse<
         }
       }),
     };
+    if (!Object.getOwnPropertyDescriptor(response, 'output_text')) {
```
apcha-oai: Ok I see what this is doing, this was an ambient bug to cover this path:
```
Raw API/SSE response
  → ResponseStream accumulates response
  → maybeParseResponse()
  → finalResponse()
```

Ideally would like us to somehow cover with properly typed data but we can punt for now, the issue is we added `output_text` to the wire types when that isn't actually how `output_text` works; it is a synthetic field added by the SDKs at runtime for convenience. Codex notes for posterity below. 
 
> I traced why the existing types did not catch this.
>
> [`Response.output_text` is declared as a required `string`](https://github.com/openai/openai-node/blob/24cebd4f50be0f3f95da45a345cdc2aa65c9a8b5/src/resources/responses/responses.ts#L964), but streaming event types use that same enriched `Response` type for raw wire data—for example, [`[ResponseCompletedEvent.response: Response](https://github.com/openai/openai-node/blob/24cebd4f50be0f3f95da45a345cdc2aa65c9a8b5/src/resources/responses/responses.ts#L1851-L1855)`](https://github.com/openai/openai-node/blob/24cebd4f50be0f3f95da45a345cdc2aa65c9a8b5/src/resources/responses/responses.ts#L1851-L1855).
>
> SSE data crosses an unchecked `any` boundary at [`[JSON.parse(sse.data) as any](https://github.com/openai/openai-node/blob/24cebd4f50be0f3f95da45a345cdc2aa65c9a8b5/src/core/streaming.ts#L58-L73)`](https://github.com/openai/openai-node/blob/24cebd4f50be0f3f95da45a345cdc2aa65c9a8b5/src/core/streaming.ts#L58-L73). `ResponseStream` then trusts the event type and stores `event.response` directly as a `Response`, both when initializing the snapshot and upon `response.completed`:
>
> - [[Initial snapshot assignment](https://github.com/openai/openai-node/blob/24cebd4f50be0f3f95da45a345cdc2aa65c9a8b5/src/lib/responses/ResponseStream.ts#L203-L212)](https://github.com/openai/openai-node/blob/24cebd4f50be0f3f95da45a345cdc2aa65c9a8b5/src/lib/responses/ResponseStream.ts#L203-L212)
> - [`response.completed` assignment](https://github.com/openai/openai-node/blob/24cebd4f50be0f3f95da45a345cdc2aa65c9a8b5/src/lib/responses/ResponseStream.ts#L281-L283)
>
> Effectively:
>
> ```ts
> const raw = JSON.parse(data) as any;
>
> // The compiler trusts this because `raw` is `any`.
> const event: ResponseCompletedEvent = raw;
>
> // Typed as string, but actually undefined when omitted from the wire payload.
> event.response.output_text;
> ```

scripts/build
```diff
@@ -15,7 +15,7 @@ rm -rf dist; mkdir dist
 # Copy src to dist/src and build from dist/src into dist, so that
 # the source map for index.js.map will refer to ./src/index.ts etc
 cp -rp src README.md dist
-for file in LICENSE CHANGELOG.md; do
+for file in LICENSE CHANGELOG.md bedrock.md; do
```
apcha-oai: Would prefer we either do this consistently for other standalone docs or leave out for now? There are a bunch like `azure.md`/`api.md` that are not included here either.

scripts/build
```diff
@@ -36,6 +36,9 @@ node scripts/utils/postprocess-files.cjs
 # import the output ESM
 (cd dist && node -e 'require("openai")')
 (cd dist && node -e 'import("openai")' --input-type=module)
+(cd dist && node -e 'require("openai/providers/bedrock")')
+(cd dist && node -e 'import("openai/providers/bedrock")' --input-type=module)
+(cd dist && node -e '(async () => { const { OpenAI } = require("openai"); const { bedrock } = await import("openai/providers/bedrock"); process.chdir(require("os").tmpdir()); const client = new OpenAI({ provider: bedrock({ region: "us-east-1", baseURL: null, accessKeyId: "test", secretAccessKey: "test" }), fetch: async () => new Response("{}", { headers: { "content-type": "application/json" } }) }); await client.models.list(); })()')
```
apcha-oai: nit: fine for back compat tests but lets leave this out of `build`

scripts/build-deno
```diff
@@ -7,8 +7,9 @@ cd "$(dirname "$0")/.."
 rm -rf dist-deno; mkdir dist-deno
 cp -rp src/* jsr.json dist-deno
 
-for file in README.md LICENSE CHANGELOG.md; do
+for file in README.md LICENSE CHANGELOG.md bedrock.md; do
```
apcha-oai: similar here re: markdown including markdown in build dists

src/internal/provider.ts
```diff
@@ -0,0 +1,48 @@
...
+}
+
+export interface ProviderDefinition {
+  configure(): ProviderRuntime;
+}
+
+const providerDefinitionsKey = Symbol.for('openai.node.providerDefinitions.v1');
+const providerGlobal = globalThis as any;
+const existingProviderDefinitions = providerGlobal[providerDefinitionsKey] as
+  | WeakMap<Provider, ProviderDefinition>
```
apcha-oai, line 41: I see what this is doing after reviewing more. Would be good nice to document this from first principles are what requires this. IIUC this is essentially:
1. bedrock(options) to capture provider configuration.
2. Fresh provider runtime state for each OpenAI client.
3. Provider-specific dependencies isolated from the core SDK.
4. Providers to work across duplicate CJS/ESM module instances.
5. A runtime guarantee that even JavaScript or any objects are rejected unless issued by createProvider().

Specifically (5) is what makes the internal provider registry necessary + weakmap/globalthis usage necessary. Otherwise we could simplify this quite a bit. If we want to keep (5) this seems reasonable though.

src/client.ts
```diff
@@ -501,7 +529,21 @@ export class OpenAI {
...
+      delete inheritedOptions.organization;
+      delete inheritedOptions.project;
+      delete inheritedOptions.defaultHeaders;
```
apcha-oai: from codex: think this is slightly too much:

```
const client = new OpenAI({
  provider,
  defaultHeaders: {
    'x-provider-custom': 'preserve-me',
  },
});

const cloned = client.withOptions({ timeout: 1 });
// x-provider-custom is now missing
```

src/providers/bedrock.ts
```diff
@@ -0,0 +1,480 @@
...
+let signingDependencies: Promise<SigningDependencies> | undefined;
+let defaultProviderDependency: Promise<DefaultProvider> | undefined;
+
+/**
+ * Keep optional AWS imports lazy and opaque to browser bundlers. TypeScript emits
+ * these as module-relative requires for CommonJS and preserves native dynamic
+ * imports for ESM, so resolution does not depend on the process working directory.
+ */
+async function importOptionalDependency(specifier: string): Promise<Record<string, any>> {
+  return import(/* webpackIgnore: true */ specifier);
```
apcha-oai: This might have issues with serverless deployments, tried a quick dummy vercel deployment and it ran into module_not_found even with the dep

apcha-oai: Checked this can also be pretty awkward with vite / force `noExternal` usage

apcha-oai: recommendation: if we change import paths to something like:
```
import { bedrock } from 'openai/providers/bedrock/aws';
```
and use standard imports there instead of runtime imports I think that will work a lot better across ecosystems. Would be good to sanity check vite/webpack/some serverless

tests/fixtures/bedrock/v1/sigv4.json
HAYDEN-OAI: yep, good call lol. i added a top-level `$comment` identifying these as the public aws sigv4 documentation credentials and explicitly saying they are not real credentials. fixed in `69c51656`.

examples/bedrock/responses.ts
HAYDEN-OAI: done, the public example is back on `openai.gpt-5.4` in `69c51656`. i kept the gpt-oss model only in the opt-in live-test command because that is the model we validated there.

README.md
HAYDEN-OAI: done here too, the readme now uses `openai.gpt-5.4`. thanks!

src/lib/ResponsesParser.ts
HAYDEN-OAI: thanks for tracing this, that distinction makes sense. i left the runtime repair scoped as-is for this pr, since separating raw wire responses from the sdk-enriched `Response` type needs broader schema work. i also captured the lesson in the On Releasing Features doc so it does not get lost.

scripts/build
HAYDEN-OAI: agreed. i removed `bedrock.md` from the npm build copy list, so it stays a repository guide like the other standalone docs.

scripts/build
HAYDEN-OAI: done, i removed the signed-request smoke test from `build`. that behavior stays covered by the focused tests and packed-artifact checks instead.

scripts/build-deno
HAYDEN-OAI: done here too. `bedrock.md` is no longer copied or converted by the deno build.

src/internal/provider.ts
HAYDEN-OAI: thanks, this framing was really helpful. i added a first-principles comment over the registry covering captured configuration, fresh per-client runtime state, dependency isolation, duplicate cjs/esm instances, the unforgeability guarantee, and why the global weakmap is used.

src/client.ts
HAYDEN-OAI: good catch. same-provider clones now preserve provider-owned headers, organization, and project values. those inherited values are dropped only when crossing from the core openai route to a provider or switching provider objects. i added the exact `x-provider-custom` regression test too.

src/providers/bedrock.ts
HAYDEN-OAI: thanks, this was the right direction. i split out `openai/providers/bedrock/aws` with normal static imports and kept `openai/providers/bedrock` bearer-only and dependency-free. i verified the packed package with the browser webpack build, vite 7 ssr using `noExternal`, and vercel's node file tracer. the full github ci run is green as well.

README.md
HAYDEN-OAI: quick follow-up after the packaging review: this behavior changed in `69c51656`. the new `/bedrock/aws` entrypoint uses static imports, so a missing peer now fails immediately with the runtime's normal module-not-found error. the base `/bedrock` bearer entrypoint remains dependency-free.

apcha-oai (dismissed):

HAYDEN-OAI:
Reverified during the upstream audit: this PR merged in [84838199](https://github.com/openai/openai-node/commit/848381998fe3d0d7132903ee0d802981939ed8fc). This PR is already closed; audit result: done.

PR #1941, opened 2026-06-15
fix: locale-sensitive header normalization

### LINKED ISSUE #1928: Bug: TypeError: Header name must be a valid HTTP token ["openaı-organization"] on Turkish Windows/Locale

jbeckwith-oai:
## Summary

Fixes #1928.

This updates `buildHeaders` to normalize header names to lowercase ASCII before passing them into `Headers.delete()` and `Headers.append()`.

## Root cause

The original issue reports `OpenAI-Organization` and `OpenAI-Project` becoming invalid header names like `openaı-organization` under a Turkish locale. Plain JavaScript `String.prototype.toLowerCase()` is not locale-sensitive in Node, so changing only the internal comparison key is not enough. The issue workaround points at the real failure mode: a bundled/runtime `Headers` implementation may normalize the mixed-case header name with Turkish locale rules when we call `delete()` or `append()` with the original name.

By handing `Headers` the SDK's already-normalized ASCII header name, we avoid depending on the runtime's header-name casing behavior.

## Validation

- `pnpm exec jest tests/buildHeaders.test.ts --runInBand`
- `env LANG=tr_TR.UTF-8 LC_ALL=tr_TR.UTF-8 pnpm exec jest tests/buildHeaders.test.ts --runInBand`
- `pnpm exec jest tests/index.test.ts tests/lib/azure.test.ts tests/log.test.ts tests/lib/workload-identity.test.ts tests/lib/bedrock.test.ts tests/api-resources/webhooks.test.ts --runInBand`
- `pnpm exec jest tests/buildHeaders.test.ts tests/api-resources/chat/completions/completions.test.ts --runInBand`
- `pnpm exec jest tests/api-resources/beta/threads/threads.test.ts tests/api-resources/videos.test.ts --runInBand`
- `pnpm exec prettier --check src/internal/headers.ts tests/buildHeaders.test.ts`
- `pnpm exec eslint src/internal/headers.ts tests/buildHeaders.test.ts`
- `pnpm exec tsc --noEmit`
- `pnpm build`

### DIFF
--- a/src/internal/headers.ts
+++ b/src/internal/headers.ts
@@ -12,6 +12,7 @@ export type HeadersLike =
   | NullableHeaders;
 
 const brand_privateNullableHeaders = /* @__PURE__ */ Symbol('brand.privateNullableHeaders');
+const httpTokenHeaderName = /^[!#$%&'*+\-.^_`|~0-9A-Za-z]+$/;
 
 /**
  * @internal
@@ -74,16 +75,19 @@ export const buildHeaders = (newHeaders: HeadersLike[]): NullableHeaders => {
   for (const headers of newHeaders) {
     const seenHeaders = new Set<string>();
     for (const [name, value] of iterateHeaders(headers)) {
+      if (!httpTokenHeaderName.test(name)) {
+        throw new TypeError(`Header name must be a valid HTTP token ["${name}"]`);
+      }
       const lowerName = name.toLowerCase();
       if (!seenHeaders.has(lowerName)) {
-        targetHeaders.delete(name);
+        targetHeaders.delete(lowerName);
         seenHeaders.add(lowerName);
       }
       if (value === null) {
-        targetHeaders.delete(name);
+        targetHeaders.delete(lowerName);
         nullHeaders.add(lowerName);
       } else {
-        targetHeaders.append(name, value);
+        targetHeaders.append(lowerName, value);
         nullHeaders.delete(lowerName);
       }
     }
--- a/tests/buildHeaders.test.ts
+++ b/tests/buildHeaders.test.ts
@@ -85,4 +85,76 @@ describe('buildHeaders', () => {
       expect(inspectNullableHeaders(buildHeaders(input))).toEqual(expected);
     });
   }
+
+  test('normalizes OpenAI headers with locale-invariant ASCII casing', () => {
+    expect('OpenAI-Organization'.toLocaleLowerCase('tr-TR')).toBe('openaı-organization');
+    expect('OpenAI-Project'.toLocaleLowerCase('tr-TR')).toBe('openaı-project');
+
+    const result = buildHeaders([
+      {
+        'OpenAI-Organization': 'org_test',
+        'OpenAI-Project': 'proj_test',
+      },
+    ]);
+
+    const keys = [...result.values.keys()];
+    expect(keys).toEqual(['openai-organization', 'openai-project']);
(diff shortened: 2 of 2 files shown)


src/internal/headers.ts
```diff
@@ -76,14 +76,14 @@ export const buildHeaders = (newHeaders: HeadersLike[]): NullableHeaders => {
     for (const [name, value] of iterateHeaders(headers)) {
       const lowerName = name.toLowerCase();
       if (!seenHeaders.has(lowerName)) {
-        targetHeaders.delete(name);
+        targetHeaders.delete(lowerName);
```
HAYDEN-OAI, line 83: nice fix, and the regression coverage for locale-sensitive `Headers` implementations is convincing. one edge case worth following up on: `toLowerCase()` performs full unicode case conversion, so an invalid name like `cooKie` becomes the valid `cookie` before `Headers` can reject it. validating the original `name` against the ascii http-token grammar before lowercasing would preserve the existing rejection behavior without reintroducing the turkish locale issue.

HAYDEN-OAI (approved):
looks good overall. thanks for the careful root-cause analysis and targeted coverage!

HAYDEN-OAI:
Reverified during the upstream audit: this PR merged in [18e8c290](https://github.com/openai/openai-node/commit/18e8c290d422ef1dfb09a4940f2a0b7703691cc1), and current `buildHeaders` uses validated ASCII lowercase names ([source](https://github.com/openai/openai-node/blob/main/src/internal/headers.ts#L81-L93)). This PR is already closed; audit result: done.

PR #1948, opened 2026-06-22
build: vendor tsc-multi build tool

jbeckwith-oai:
## Summary

- Vendor the `tsc-multi` v1.1.11 TypeScript source under `scripts/_vendor/tsc-multi`.
- Remove the GitHub release tarball dependency from `package.json` and `pnpm-lock.yaml`.
- Invoke the vendored CLI from `scripts/build` via `ts-node/register/transpile-only`.
- Add only the vendored CLI runtime imports that remain necessary as direct dev dependencies.
- Replace low-risk vendored imports with Node built-ins/local code: `debug`, `get-stdin`, `picocolors`, and `string-to-stream`.
- Remove stale `tsc-multi` exceptions from `pnpm-workspace.yaml` trust/release-age policy.
- Keep forked worker `ts-node` compilation independent of target `--cwd` configs by skipping project discovery and pinning explicit CommonJS compiler options.

Linear: SDK-75

## Validation

- `pnpm test` - 101 suites / 1031 tests passed, 1 suite intentionally skipped
- `pnpm lint` - includes Prettier, ESLint, build, Deno dist generation, `tsc`, ATTW, publint, and JSR publish dry-run
- Vendored CLI dry run: `node -r ts-node/register/transpile-only scripts/_vendor/tsc-multi/src/cli.ts --dry --config tsc-multi.json`
- Isolated `/tmp` fixture covering `--cwd`, `--config`, dual `.js`/`.mjs` targets, CJS require, ESM import, and `--clean`
- Review regression fixture: dry-ran vendored CLI against `ecosystem-tests/bun` with `module: "esnext"` and `moduleResolution: "bundler"` while preserving `--cwd ecosystem-tests/bun`
- Additional isolated bundler fixture: verified a target project using `module: "esnext"` and `moduleResolution: "bundler"` no longer affects the forked worker
- Original-vs-vendored build output comparison: built this repo once with the vendored CLI and once with the original v1.1.11 release CLI, then `diff -qr` compared all 2521 files with no differences; reran after the worker config fix and current vendored output remains byte-for-byte identical to the tarball baseline


scripts/_vendor/tsc-multi/src/build.ts
```diff
@@ -0,0 +1,123 @@
...
+}
+
+async function runWorker({
+  stdout,
+  stderr,
+  ...options
+}: WorkerOptions & Pick<BuildOptions, 'stdout' | 'stderr'>): Promise<number> {
+  const worker = fork(WORKER_PATH, [], {
+    cwd: options.cwd,
+    execArgv: ['--require', TS_NODE_REGISTER],
```
HAYDEN-OAI, line 59: could we keep the worker ts-node config independent of `--cwd` here? because the child runs with `cwd: options.cwd`, `ts-node/register/transpile-only` discovers that project tsconfig when compiling this vendored `.ts` worker. for projects using `module: "esnext"` and `moduleResolution: "bundler"`, the worker is emitted as esm and its extensionless imports fail with `ERR_MODULE_NOT_FOUND` before the build starts. i reproduced this with `ecosystem-tests/bun/tsconfig.json`. pinning ts-node to a vendor-owned commonjs config, or skipping project discovery with explicit compiler options, should preserve `--cwd` across target configs. thanks!

jbeckwith-oai: Very good catch

HAYDEN-OAI (approved):

PR #1132, opened 2024-10-15
Run tool calls in parallel

jmoseley:
- [x] I understand that this repository is auto-generated and my pull request may not be merged

## Changes being requested

Run tool calls in parallel, since they have no dependency on one another.

## Additional context & links

https://github.com/openai/openai-node/issues/1131

PR #1949, opened 2026-06-23
Run tool calls in parallel safely

### LINKED ISSUE #1131: runTools function isn't running tools in parallel

HAYDEN-OAI:
## Summary

- preserve @jmoseley's original contribution and commit authorship from #1132
- run groups of function tool calls concurrently while preserving tool result order
- retain sequential execution when `parallel_tool_calls` is `false` or a single function is forced
- wait for all concurrent calls to settle before surfacing an error, preventing background side effects after the runner ends
- document the behavior and add a focused concurrency/order regression test

## History

#1132 was first merged into this integration branch, then current `main` was merged in and the review findings were fixed here. The branch therefore keeps Jeremy's original commit as a distinct authored commit instead of replacing it with a clean-room implementation.

## Validation

- `./scripts/test tests/lib/ChatCompletionRunFunctions.test.ts --runInBand` (17 tests)
- Prettier check on changed files
- ESLint on changed TypeScript files
- `tsc --noEmit`

Closes #1131

### DIFF
--- a/src/lib/AbstractChatCompletionRunner.ts
+++ b/src/lib/AbstractChatCompletionRunner.ts
@@ -356,8 +356,8 @@ export class AbstractChatCompletionRunner<
         return;
       }
 
-      for (const tool_call of message.tool_calls) {
-        if (tool_call.type !== 'function') continue;
+      await Promise.all(message.tool_calls.map(async (tool_call) => {
+        if (tool_call.type !== 'function') return;
         const tool_call_id = tool_call.id;
         const { name, arguments: args } = tool_call.function;
         const fn = functionsByName[name];
@@ -370,14 +370,14 @@ export class AbstractChatCompletionRunner<
             .join(', ')}. Please try again`;
 
           this._addMessage({ role, tool_call_id, content });
-          continue;
+          return;
         } else if (singleFunctionToCall && singleFunctionToCall !== name) {
           const content = `Invalid tool_call: ${JSON.stringify(name)}. ${JSON.stringify(
             singleFunctionToCall,
           )} requested. Please try again`;
 
           this._addMessage({ role, tool_call_id, content });
-          continue;
+          return;
         }
 
         let parsed;
@@ -386,7 +386,7 @@ export class AbstractChatCompletionRunner<
         } catch (error) {
           const content = error instanceof Error ? error.message : String(error);
           this._addMessage({ role, tool_call_id, content });
-          continue;
+          return;
         }
 
         // @ts-expect-error it can't rule out `never` type.
@@ -398,7 +398,7 @@ export class AbstractChatCompletionRunner<
           await afterCompletion?.(chatCompletion, this);
           return;
         }
-      }
+      });
 
       await afterCompletion?.(chatCompletion, this);
     }
--- a/src/lib/AbstractChatCompletionRunner.ts
+++ b/src/lib/AbstractChatCompletionRunner.ts
@@ -14,8 +14,11 @@ import type {
   ParsedChatCompletion,
 } from '../resources/chat/completions';
 import type { CompletionUsage } from '../resources/completions';
-import type { ChatCompletionToolRunnerParams } from './ChatCompletionRunner';
-import type { ChatCompletionStreamingToolRunnerParams } from './ChatCompletionStreamingRunner';
+import type { ChatCompletionRunner, ChatCompletionToolRunnerParams } from './ChatCompletionRunner';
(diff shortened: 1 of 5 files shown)


jbeckwith-oai (approved):
Reviewed by Codex. Approved; detailed feedback has been moved into inline code comments.

src/lib/AbstractChatCompletionRunner.ts
```diff
@@ -336,6 +335,50 @@ export class AbstractChatCompletionRunner<
...
+
+      let parsed;
+      try {
+        parsed = isRunnableFunctionWithParse(fn) ? await fn.parse(args) : args;
+      } catch (error) {
+        const content = error instanceof Error ? error.message : String(error);
+        return { message: { role, tool_call_id, content }, functionCalled: false };
+      }
+
+      // @ts-expect-error it can't rule out `never` type.
```
jbeckwith-oai: Codex: This pre-existing suppression moved into the new helper, but the refactor is a good chance to eliminate it. Could we split the parse/no-parse branches and call `fn.function(...)` inside each branch so TypeScript keeps the argument relationship instead of suppressing it?

src/lib/AbstractChatCompletionRunner.ts
```diff
@@ -356,47 +399,31 @@ export class AbstractChatCompletionRunner<
...
-          continue;
-        } else if (singleFunctionToCall && singleFunctionToCall !== name) {
-          const content = `Invalid tool_call: ${JSON.stringify(name)}. ${JSON.stringify(
-            singleFunctionToCall,
-          )} requested. Please try again`;
-
-          this._addMessage({ role, tool_call_id, content });
-          continue;
-        }
+      if (singleFunctionToCall || params.parallel_tool_calls === false) {
```
jbeckwith-oai, line 410: Codex: Since this branch is the documented sequential opt-out for `parallel_tool_calls: false`, please add a regression test that blocks the first tool and asserts the second tool has not started until the first resolves. The current new test only pins the default parallel path.

tests/lib/ChatCompletionRunFunctions.test.ts
```diff
@@ -220,7 +220,7 @@ class RunnerListener {
     );
     expect(await this.runner.finalFunctionToolCallResult()).toEqual(this.finalFunctionCallResult);
     expect(this.chatCompletions).toEqual(this.runner.allChatCompletions());
-    expect(this.messages).toEqual(this.runner.messages.slice(-this.messages.length));
+    expect(this.messages).toEqual(this.runner.messages.filter((message) => this.messages.includes(message)));
```
jbeckwith-oai: Codex: This identity filter works around `afterCompletion` injecting a non-emitted message, but it also makes the helper ignore any other non-emitted messages that accidentally appear. I would make the allowed ignored message(s) explicit instead of broadly filtering `runner.messages`.

tests/lib/ChatCompletionRunFunctions.test.ts
```diff
@@ -324,7 +324,9 @@ class StreamingRunnerListener {
     );
     expect(await this.runner.finalFunctionToolCallResult()).toEqual(this.finalFunctionCallResult);
     expect(this.eventChatCompletions).toEqual(this.runner.allChatCompletions());
-    expect(this.eventMessages).toEqual(this.runner.messages.slice(-this.eventMessages.length));
+    expect(this.eventMessages).toEqual(
```
jbeckwith-oai: Codex: Same concern as the non-streaming listener: this broad identity filter weakens the invariant that emitted messages match runner state. Please make the `afterCompletion` injected-message exception explicit.

jbeckwith-oai (commented):
Reviewed by Codex. Moving the detailed feedback into inline review comments as requested.

src/lib/AbstractChatCompletionRunner.ts
HAYDEN-OAI: thanks for catching this! i split the parse and no-parse paths so the argument types stay correlated, and passed the concrete runner through without a suppression. fixed in deb566b1.

PR #1951, opened 2026-06-23
Preserve MIME type in toFile

fallintoplace:
## Summary

- Preserve MIME type and `lastModified` when `toFile()` clones File-like objects.
- Default Response-derived files to the blob MIME type when the caller did not pass one.
- Fix Blob part MIME type inference by reading `part.type` instead of checking the matched part object.

## Why

Some upload inputs were converted into new `File` objects without carrying their content type forward. That could leave files with an empty MIME type even when the original Blob, File-like object, or Response body had one.

## Validation

- `./node_modules/.bin/jest tests/uploads.test.ts --runInBand`
- `./node_modules/.bin/prettier --check src/internal/to-file.ts tests/uploads.test.ts`
- `./node_modules/.bin/eslint src/internal/to-file.ts tests/uploads.test.ts`
- `./scripts/build`

### DIFF
--- a/src/internal/to-file.ts
+++ b/src/internal/to-file.ts
@@ -95,12 +95,19 @@ export async function toFile(
     if (value instanceof File) {
       return value;
     }
-    return makeFile([await value.arrayBuffer()], value.name);
+    return makeFile([await value.arrayBuffer()], value.name, {
+      type: value.type,
+      lastModified: value.lastModified,
+      ...options,
+    });
   }
 
   if (isResponseLike(value)) {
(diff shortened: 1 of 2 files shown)


HAYDEN-OAI:
Thanks for the contribution! File-like metadata preservation is already tracked in #1883. A complete fix should also carry forward the additional cases covered here: Response-derived MIME types, Blob-part MIME inference, and explicit metadata overrides. I’m closing this as a duplicate so those requirements can be addressed together in #1883.

PR #1953, opened 2026-06-23
Validate maxRetries

fallintoplace:
## Summary

- Validate client-level and per-request `maxRetries` as non-negative integers.
- Change retry gates to check `retriesRemaining > 0` instead of truthiness.
- Add coverage for invalid retry counts and for `maxRetries: 0` disabling retries.

## Why

Negative retry counts are truthy, so retryable failures could keep decrementing forever instead of stopping. Invalid retry counts now fail before a request is sent.

## Validation

- `./node_modules/.bin/jest tests/index.test.ts --runInBand`
- `./node_modules/.bin/prettier --check src/client.ts src/internal/utils/values.ts tests/index.test.ts`
- `./node_modules/.bin/eslint src/client.ts src/internal/utils/values.ts tests/index.test.ts`
- `./scripts/build`

### DIFF
--- a/src/client.ts
+++ b/src/client.ts
@@ -3,7 +3,12 @@
 import type { RequestInit, RequestInfo, BodyInit } from './internal/builtin-types';
 import type { HTTPMethod, PromiseOrValue, MergedRequestInit, FinalizedRequestInit } from './internal/types';
 import { uuid4 } from './internal/utils/uuid';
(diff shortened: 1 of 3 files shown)


HAYDEN-OAI:
Thanks for the contribution. We are not planning to reject invalid maxRetries values because throwing would introduce a backward-incompatible behavior change for existing applications. If we revisit this, we would prefer falling back to the default retry count instead of rejecting requests or failing client construction. Closing as not planned.

PR #1954, opened 2026-06-23
Reject emitted promises on error

fallintoplace:
## Summary

- Make `EventEmitter.emitted(event)` reject when `error` is emitted before the requested event.
- Keep `emitted('error')` resolving with the error value.
- Remove the opposite one-time listener after either resolution or rejection.
- Add coverage for both `core` and `lib` EventEmitter implementations.

## Why

The docstring promised that `emitted()` rejects on `error`, but the implementation only listened for the requested event. Code awaiting a message-like event could hang forever if an error arrived first.

## Validation

- `./node_modules/.bin/jest tests/EventEmitter.test.ts --runInBand`
- `./node_modules/.bin/prettier --check src/lib/EventEmitter.ts src/core/EventEmitter.ts tests/EventEmitter.test.ts`
- `./node_modules/.bin/eslint src/lib/EventEmitter.ts src/core/EventEmitter.ts tests/EventEmitter.test.ts`
- `./scripts/build`

### DIFF
--- a/src/core/EventEmitter.ts
+++ b/src/core/EventEmitter.ts
@@ -74,8 +74,18 @@ export class EventEmitter<EventTypes extends Record<string, (...args: any) => an
     : EventParameters<EventTypes, Event>
   > {
     return new Promise((resolve, reject) => {
-      // TODO: handle errors
-      this.once(event, resolve as any);
+      const errorEvent = 'error' as keyof EventTypes;
+      const onEvent = ((value: unknown) => {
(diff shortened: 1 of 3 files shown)


HAYDEN-OAI:
Thanks for the contribution! The public library EventEmitter already rejects pending emitted-event promises when an error occurs, as implemented in https://github.com/openai/openai-node/commit/e0b09da782f0d6ce69a383fd3f3d02a0c7608d6d. The separate core emitter intentionally has different event-resolution semantics, so we’re closing this proposed change as already addressed.

PR #1956, opened 2026-06-23
Clean up abort signal listeners

fallintoplace:
## Summary
- keep caller-provided abort signals connected after response headers arrive
- remove the abort listener after SDK-owned body consumption finishes or is cancelled
- add regression coverage for normal body parsing, aborting during a pending body read, and streamed response iteration

## Why
`fetchWithTimeout()` previously cleaned up the abort listener as soon as `fetch()` resolved. That avoided the listener leak, but it also disconnected caller aborts while `APIPromise` was still parsing the response body or while a stream was being consumed.

This keeps the listener attached through body consumption, then removes it from the parse and stream finalization paths.

## Tests
- `./node_modules/.bin/jest tests/index.test.ts --runInBand`
- `./node_modules/.bin/prettier --check src/client.ts src/internal/parse.ts src/core/streaming.ts tests/index.test.ts`
- `./node_modules/.bin/eslint src/client.ts src/internal/parse.ts src/core/streaming.ts tests/index.test.ts`
- `./scripts/build`

### DIFF
--- a/src/client.ts
+++ b/src/client.ts
@@ -246,6 +246,11 @@ import { isEmptyObj } from './internal/utils/values';
 
 const WORKLOAD_IDENTITY_API_KEY_PLACEHOLDER = 'workload-identity-auth';
 
+type FetchWithTimeoutResponse = {
+  response: Response;
+  cleanupAbortSignal: () => void;
+};
+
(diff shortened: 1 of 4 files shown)


HAYDEN-OAI:

PR #1952, opened 2026-06-23
Fix playAudio for Response streams

fallintoplace:
## Summary

- Convert Response Web ReadableStream bodies with `Readable.fromWeb()` before piping them to `ffplay`.
- Add a regression test using a native `Response` and mocked `ffplay` stdin.

## Why

`playAudio(response)` treated `Response.body` as a Node stream and called `.pipe()` directly. Native fetch responses expose a Web `ReadableStream`, so `.pipe()` is not available and playback fails at runtime.

## Validation

- `./node_modules/.bin/jest tests/helpers/audio.test.ts --runInBand`
- `./node_modules/.bin/prettier --check src/helpers/audio.ts tests/helpers/audio.test.ts`
- `./node_modules/.bin/eslint src/helpers/audio.ts tests/helpers/audio.test.ts`
- `./node_modules/typescript/bin/tsc`
- `./scripts/build`

PR #1960, opened 2026-06-25
fix(audio): support Response streams safely

HAYDEN-OAI:
## Summary

- carry forward @fallintoplace's two original commits from #1952 verbatim
- convert native `Response.body` Web streams with `Readable.fromWeb()` while preserving Node-readable bodies from custom fetch implementations
- use `stream.pipeline()` so playback rejects cleanly if the response stream fails and terminates `ffplay` instead of leaving the child hanging
- add focused coverage for native Web streams, Node-backed response streams, and mid-stream failures

## Why

`playAudio(response)` currently assumes every response body has Node's `.pipe()` method. Native fetch responses expose a Web `ReadableStream`, so playback fails immediately. The original PR fixed that incompatibility and preserved the existing Node-stream path after review feedback.

During deeper review, we found that an error from the adapted Web stream would be emitted without a listener, potentially terminating the process instead of rejecting `playAudio()`. Routing the selected source through `pipeline()` handles that failure and closes the playback pipe safely.

## Attribution

This integration branch contains the original #1952 commits `fb2dc044` and `c9e3996a` unchanged. A direct maintainer push to the external fork was blocked by local repository policy, so the original history was carried into this upstream branch before adding the review fix.

## Validation

- `pnpm exec jest tests/helpers/audio.test.ts --runInBand`
- `pnpm exec tsc --noEmit`
- `pnpm exec eslint src/helpers/audio.ts tests/helpers/audio.test.ts`
- `pnpm exec prettier --check src/helpers/audio.ts tests/helpers/audio.test.ts`
- `pnpm build`

Supersedes #1952.


HAYDEN-OAI:
closing this out since the fix landed through the original author’s PR in #1952. thanks!

PR #1961, opened 2026-06-26
fix(client): allow explicit empty apiKey string

### LINKED ISSUE #1957: apiKey became mandatory in v6.36.0

### Describe the bug

Since this commit:
https://github.com/openai/openai-node/commit/359afb9e4320f2205009dbc32123b87a7faaf967

An empty apiKey ("") doesn't work anymore:
```

     42|     const client = new OpenAI({
       |                    ^
Error: Error: Missing credentials. Please pass an `apiKey`, `workloadIdentity`, `adminAPIKey`, or set the `OPENAI_API_KEY` or `OPENAI_ADMIN_KEY` environment variable.
     43|         apiKey: "",
```

### To Reproduce

Create an OpenAI client with an empty apiKey

### Code snippets

```JavaScript

```

### OS

any

### Node version

any

### Library version

>=6.36

mralonsocorona:
## Summary

Fixes a credential validation regression introduced when admin API key auth support added a missing-credentials guard.

The guard used `!apiKey`, which treats an explicitly provided empty string as missing. Since `apiKey` is typed as `string | ApiKeySetter | null | undefined`, this updates the check to treat only `null` and `undefined` as missing.

Fixes #1957.

## Changes

- Use `apiKey == null` in the missing-credentials guard.
- Add a regression test for `new OpenAI({ apiKey: "" })`.

## Testing

- `./node_modules/.bin/jest tests/index.test.ts -t "apiKey empty string|at least one"`
- `./scripts/format`
- `./scripts/lint`

`tests/index.test.ts`: 64/64 passing.

### DIFF
--- a/src/client.ts
+++ b/src/client.ts
@@ -456,7 +456,7 @@ export class OpenAI {
       throw new Errors.OpenAIError('The `apiKey` and `workloadIdentity` options are mutually exclusive');
     }
 
-    if (!providerRuntime && !apiKey && !adminAPIKey && !workloadIdentity) {
+    if (!providerRuntime && apiKey == null && !adminAPIKey && !workloadIdentity) {
       throw new Errors.OpenAIError(
         'Missing credentials. Please pass an `apiKey`, `workloadIdentity`, `adminAPIKey`, or set the `OPENAI_API_KEY` or `OPENAI_ADMIN_KEY` environment variable.',
       );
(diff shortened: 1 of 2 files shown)


HAYDEN-OAI:
Thanks for reporting this! The OpenAI client requires valid credentials, and an empty API key is not a supported credential. For keyless compatible servers, a placeholder API key can be used as discussed in #1957. We’re not planning to accept an empty API key.

PR #1962, opened 2026-06-28
Codex/add agent skill

rhampd550-jpg:


HAYDEN-OAI:
thanks for putting this together. there are several unrelated changes bundled into this PR, so we are not going to merge it as-is. the small ResponseStream error-message fix looks useful though, and we would be happy to consider that as a focused PR. appreciate the effort!

PR #1963, opened 2026-07-02
fix(assistants): place array delta entries by index instead of appending

spokodev:
When accumulating streamed array deltas, `AssistantStream.accumulateDelta`
reads and writes entries by their `index` field. The branch that handles a
not-yet-seen entry appended it with `push` instead of assigning it at
`accValue[index]`. These only agree when the index equals the current array
length, so any non-contiguous or out-of-order `index` lands the entry at the
wrong position.

The consequence is data corruption: because the entry is not at `accValue[index]`,
a later fragment for the same index finds an empty slot and appends again. The
same tool call (or content block) ends up as two separate array entries with
split, invalid JSON arguments in the accumulated snapshot.

This assigns the entry at its `index`, matching the sibling branch and the
Python SDK, which uses `insert(index, entry)` for the same case. The existing
contiguous path is unchanged.

Added a unit test that seeds an entry at index 0, then streams two fragments
for index 2. Before the fix the array holds two index-2 entries with split
arguments; after the fix it holds one merged entry.

### DIFF
--- a/src/lib/AssistantStream.ts
+++ b/src/lib/AssistantStream.ts
@@ -686,7 +686,7 @@ export class AssistantStream
 
           const accEntry = accValue[index];
           if (accEntry == null) {
-            accValue.push(deltaEntry);
+            accValue[index] = deltaEntry;
           } else {
             accValue[index] = this.accumulateDelta(accEntry, deltaEntry);
           }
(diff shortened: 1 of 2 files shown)


HAYDEN-OAI (approved):
thanks for the clear explanation and the focused regression test. using the delta entry's index here matches the surrounding snapshot logic and fixes the out-of-order accumulation case cleanly. great work!

PR #1965, opened 2026-07-03
Handle keepalive events in Responses streams

### LINKED ISSUE #1964: v6.45.0 can't handle stream keepalive events

### Describe the bug

On a long-running request, I see:

```
Unhandled response stream event: {"type":"keepalive","sequence_number":466}
```

I don't think OpenAI ever *documented* keepalive events but they definitely exist.

### To Reproduce

1. Send a request that will take a long time to the Responses API, such as using gpt-5.5-pro on xhigh or something. Must be a streaming request. Disable reasoning summaries to increase chance of repro.
2. Wait
3. SDK emits an error as soon as a `keepalive` event is emitted because there's no case for it.

https://github.com/openai/openai-node/blob/71d2347c59007124e7b317db0205b98a69d747ee/src/lib/responses/ResponseAccumulator.ts#L410-L412

### Code snippets

```JavaScript

```

### OS

Windows

### Node version

Node v24.14.0

### Library version

openai v6.45.0

koriyoshi2041:
Problem
Responses streaming can receive `keepalive` SSE events on long-running requests. They do not carry Response state, but the accumulator currently falls through to the exhaustive error path and ends the stream. Fixes #1964.

Fix
Treat `keepalive` as a no-op accumulator event, alongside other stream events that do not map onto the Response snapshot. Unknown event types still go through the existing error path.

Test
- `pnpm test tests/lib/ResponseAccumulator.test.ts`
- `pnpm exec prettier --check src/lib/responses/ResponseAccumulator.ts tests/lib/ResponseAccumulator.test.ts`
- `pnpm build`
- `git diff --check`

Risk
Low. This only ignores the explicit `keepalive` event; it does not change handling for normal response deltas or undocumented unknown event types.

### DIFF
--- a/src/lib/responses/ResponseAccumulator.ts
+++ b/src/lib/responses/ResponseAccumulator.ts
@@ -6,6 +6,11 @@ import {
 import { OpenAIError } from '../../error';
 import { addOutputText } from '../ResponsesParser';
 
+type ResponseKeepAliveEvent = {
+  type: 'keepalive';
+  sequence_number: number;
+};
+
 /**
  * Applies a streaming event to a response snapshot.
  *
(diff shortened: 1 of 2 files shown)


HAYDEN-OAI (approved):
thanks for the thoughtful fix and the focused regression test. this is a nice, narrowly scoped way to handle keepalive events while preserving the error path for genuinely unknown events. great work!

PR #1968, opened 2026-07-06
export runner options type

### LINKED ISSUE #593: RunnerOptions is not exposed for runTools and runFunctions

### Describe the bug

The `maxChatCompletions` option is very helpful in managing costs. However, it is not exposed through the interface due to a typing bug.

I think the fix is to simply change the Core.RequestOptions type to RunnerOptions which only contains maxCompletions at the moment.

### To Reproduce

Try passing `maxCompletions` to the runTools/runFunctions options

Screenshot 2023-12-20 at 4 10 47 PM
Screenshot 2023-12-20 at 4 10 53 PM

### Code snippets

_No response_

### OS

macOS

### Node version

v18.16.0

### Library version

v4.24.0

sarptandoven:
summary
- export `RunnerOptions` from the public package and chat completions resource surfaces
- expose the type on the `OpenAI` namespace for parity with other chat completion types
- add a type regression covering root and resource imports

change
- `openai` now supports `import type { RunnerOptions } from "openai"`
- `openai/resources/chat/completions` now supports `import type { RunnerOptions } ...`
- `OpenAI.RunnerOptions` is available in the root namespace

why
`runTools` already accepts `RunnerOptions` internally, but users could not name the options type from public exports when configuring `maxChatCompletions`.

fixes #593

tests
- `pnpm exec tsc --noEmit --pretty false`
- `pnpm exec jest tests/runner-options-exports.test.ts --runInBand`
- `pnpm lint`

### DIFF
--- a/src/client.ts
+++ b/src/client.ts
@@ -229,6 +229,7 @@ import {
   ChatCompletionUpdateParams,
   ChatCompletionUserMessageParam,
   ChatCompletionsPage,
+  RunnerOptions,
 } from './resources/chat/completions/completions';
 import { type Fetch } from './internal/builtin-types';
 import { isRunningInBrowser } from './internal/detect-platform';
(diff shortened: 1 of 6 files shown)


HAYDEN-OAI:
Thanks for the contribution! The underlying runTools options support requested in #593 has already shipped in https://github.com/openai/openai-node/commit/4c622f9f55529e3aab30c834349d341038499db1, and RunnerOptions can already be imported from its existing openai/lib/AbstractChatCompletionRunner subpath. We’re not planning to expand the root and resource namespaces solely to duplicate that type export.

PR #1969, opened 2026-07-06
export stream types from root

### LINKED ISSUE #558: Stream nor ChatCompletionStream types not exported

### Describe the bug

One of **openai.chat.completions.create** return types is **ApiPromise<Stream<OpenAI.Chat.ChatCompletionChunk>>**
yet the Stream is not the intrinsic Nodejs Stream, and not exported from openai.

### To Reproduce

try to set type of awaited or in Promise return value from **openai.chat.completions.create()**

### Code snippets

```JavaScript
import OpenAI from 'openai'

import { config } from 'server/config'
import { errorHandler } from 'server/helpers'

const openai = new OpenAI({
  apiKey: config?.OPENAI_API_KEY as string,
})

export const createText = async(prompt: string): Promise<OpenAI.Chat.ChatCompletion | null> => {
  try {
    const params: OpenAI.Chat.ChatCompletionCreateParams = {
      messages: [{ role: 'user', content: prompt }],
      model: 'gpt-3.5-turbo',
      stream: true,
    }

    const options: OpenAI.RequestOptions = {}

    const chatCompletion: OpenAI.Stream<OpenAI.Chat.ChatCompletionChunk> = await openai.chat.completions.create(params, options)

    return chatCompletion
  } catch (error: unknown) {
    console.error(error)
    errorHandler(error)
    return null
  }
}
```

### OS

any

### Node version

18.13.0

### Library version

4.20.1

sarptandoven:
summary
- exports Stream and ChatCompletionStream from the root package entrypoint
- exposes the same stream types through the OpenAI namespace
- adds a regression test covering root imports and namespace type access

why
- fixes #558
- users can access these stream classes internally, but cannot import the types directly from openai without reaching into internal paths

validation
- pnpm exec jest tests/stream-exports.test.ts --runInBand
- pnpm exec tsc --noEmit --pretty false
- pnpm lint
- git diff --check

### DIFF
--- a/src/client.ts
+++ b/src/client.ts
@@ -15,6 +15,8 @@ import { stringifyQuery } from './internal/utils/query';
 import { VERSION } from './version';
 import * as Errors from './core/error';
 import * as Pagination from './core/pagination';
+import * as Streaming from './streaming';
+import * as ChatCompletionStreams from './lib/ChatCompletionStream';
 import type { WorkloadIdentity } from './auth/types';
 import { WorkloadIdentityAuth } from './auth/workload-identity-auth';
 import { OAuthError, SubjectTokenProviderError } from './core/error';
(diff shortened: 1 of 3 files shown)


HAYDEN-OAI:
Thanks for the contribution! The stream classes can already be imported from openai/streaming and openai/lib/ChatCompletionStream, as described in #558. This implementation also declares OpenAI.Stream/OpenAI.ChatCompletionStream without defining those runtime statics, so we’re closing this approach.

PR #1973, opened 2026-07-06
feat: pass context to runTools callbacks

### LINKED ISSUE #597: Passing global context into runTools helper

I currently have a pattern where I need to pass context to my tools to allow them to act on my app. For example:

```
function updateEvent(context: { eventId: string}, args: ArgsFromOpenAi) {
   const event = await fetchEvent(eventId);
}
```

It'd be great if there were some way to pass a global context to the runner since the runner is passed into each function call. Then, I could do something like this:

```
function updateEvent( args: ArgsFromOpenAi, runner: ChatCompletionStreamingRunner<EventContext>) {
  const { eventId } = runner.context;
   const event = await fetchEvent(eventId);
}
```

### Additional context

A workaround is to build my own runner that leverages the existing helpers. However, this is complicated because of the types integration.

HAYDEN-OAI:
## Summary

- add a typed `toolContext` option to both streaming and non-streaming `runTools` parameters
- pass the same context value as the third argument to raw and parsed tool callbacks
- remove `toolContext` before creating Chat Completions requests so it remains local to the runner
- add runtime and compile-time coverage for context delivery, identity preservation, request omission, and type inference

## Motivation

`runTools` users currently need closures or custom wrapper types to share per-run application state across tools defined in separate modules. This adds a first-class context channel while preserving existing callback signatures and keeping the context scoped to one runner invocation.

Resolves #597

## Validation

- `pnpm run format`
- `pnpm exec tsc --noEmit`
- `pnpm exec jest tests/lib/ChatCompletionRunFunctions.test.ts --runInBand`
- `git diff --check`

### DIFF
--- a/src/lib/ChatCompletionRunner.ts
+++ b/src/lib/ChatCompletionRunner.ts
@@ -16,16 +16,40 @@ export interface ChatCompletionRunnerEvents extends AbstractChatCompletionRunner
   content: (content: string) => void;
 }
 
-export type ChatCompletionToolRunnerParams<
+type ChatCompletionToolRunnerParamsBase = Omit<ChatCompletionCreateParamsNonStreaming, 'tools'>;
+
+/**
+ * Parameters for tools that do not require a context value.
+ */
+export type ChatCompletionToolRunnerParamsWithoutContext<FunctionsArgs extends BaseFunctionsArgs> =
+  ChatCompletionToolRunnerParamsBase & {
+    tools: RunnableTools<FunctionsArgs> | AutoParseableTool<any, true>[];
+    toolContext?: never;
+  };
+
+/**
+ * Parameters for tools that require a context value.
+ */
+export type ChatCompletionToolRunnerParamsWithContext<
   FunctionsArgs extends BaseFunctionsArgs,
-  ToolContext = unknown,
-> = Omit<ChatCompletionCreateParamsNonStreaming, 'tools'> & {
+  ToolContext,
+> = ChatCompletionToolRunnerParamsBase & {
   tools: RunnableTools<FunctionsArgs, ToolContext> | AutoParseableTool<any, true>[];
   /**
    * Context to pass to each tool callback during this run.
    */
-  toolContext?: ToolContext;
+  toolContext: ToolContext;
 };
+
+/**
+ * Parameters for running tools. Supplying a context type makes `toolContext`
+ * required; omitting it preserves the existing no-context form.
+ */
+export type ChatCompletionToolRunnerParams<FunctionsArgs extends BaseFunctionsArgs, ToolContext = never> = [
+  ToolContext,
+] extends [never] ?
+  ChatCompletionToolRunnerParamsWithoutContext<FunctionsArgs>
+: ChatCompletionToolRunnerParamsWithContext<FunctionsArgs, ToolContext>;
 
 export class ChatCompletionRunner<ParsedT = null> extends AbstractChatCompletionRunner<
   ChatCompletionRunnerEvents,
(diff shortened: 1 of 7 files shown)


src/lib/ChatCompletionRunner.ts
```diff
@@ -16,20 +16,24 @@ export interface ChatCompletionRunnerEvents extends AbstractChatCompletionRunner
...
 }
 
-export type ChatCompletionToolRunnerParams<FunctionsArgs extends BaseFunctionsArgs> = Omit<
-  ChatCompletionCreateParamsNonStreaming,
-  'tools'
-> & {
-  tools: RunnableTools<FunctionsArgs> | AutoParseableTool<any, true>[];
+export type ChatCompletionToolRunnerParams<
+  FunctionsArgs extends BaseFunctionsArgs,
+  ToolContext = unknown,
```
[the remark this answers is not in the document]
HAYDEN-OAI: fixed in `312facbf`. i added dedicated public overloads for context-bearing runnable tools so `toolContext` flows into the callback's third parameter in the emitted package declarations, plus a broad contextual tool shape that preserves inference across parsed and raw tools. i also normalized optional context inside the runner, which fixes the package-build TS2345. verified with `pnpm run lint` and a strict typecheck against the built `dist` package that accepts `context.eventId` and rejects `context.missing`. thanks for the careful catch!

jbeckwith-oai (approved):

src/lib/ChatCompletionRunner.ts
```diff
@@ -16,20 +16,24 @@ export interface ChatCompletionRunnerEvents extends AbstractChatCompletionRunner
...
+> = Omit<ChatCompletionCreateParamsNonStreaming, 'tools'> & {
+  tools: RunnableTools<FunctionsArgs, ToolContext> | AutoParseableTool<any, true>[];
+  /**
+   * Context to pass to each tool callback during this run.
+   */
+  toolContext?: ToolContext;
```
jbeckwith-oai: **[P1] Require context when callbacks require it**

Because this property stays optional for a concrete `ToolContext`, the following compiles under `strict` (including against the built package):

```ts
const tool: RunnableToolFunction<string, { eventId: string }> = {
  type: 'function',
  function: { parameters: {}, description: 'x',
    function: (_args, _runner, context) => context.eventId },
};
ChatCompletionRunner.runTools(openai, { model, messages, tools: [tool] });
```

At runtime `_runTools` casts the omitted value to `ToolContext` and invokes the callback with `undefined`; the example throws while reading `eventId`. The resource helper also accepts this typed-tool form without `toolContext`, and the streaming params have the same issue. Please expose a context-bearing form/overload where `toolContext` is required, while retaining a separate optional/no-context form for existing callbacks.

jbeckwith-oai (changes requested):
I verified the linked #597 use case against the built declarations and both runner modes. The base revision rejects a three-argument callback/context-bearing call, while this head correctly infers `context.eventId`; the focused runner suite (21 tests), `tsc --noEmit`, and an additional streaming harness all pass. The same context object is preserved and `toolContext` is absent from every API request.

There is one blocking soundness issue: concrete context-bearing tools can still be invoked without a context, which compiles and then supplies `undefined` to a callback typed as receiving the concrete context. Details and a minimal reproduction are inline. Existing two-argument callbacks remain source/runtime compatible, CI's breaking-change check and SDK tests are green, and I found no other behavioral regressions.

src/lib/ChatCompletionRunner.ts
HAYDEN-OAI: good catch, i split the required-context and no-context params and overloads across the resource, static, and streaming entry points, with compile-time coverage that rejects a typed context tool when `toolContext` is omitted.

jbeckwith-oai (approved):

PR #1974, opened 2026-07-06
support realtime transcription intent

### LINKED ISSUE #1511: Azure: Support realtime transcription

Hi,

According to the Python example [here](https://techcommunity.microsoft.com/blog/azure-ai-services-blog/real-time-speech-transcription-with-gpt-4o-transcribe-and-gpt-4o-mini-transcribe/4410353), an additional query parameter should be passed:
```
intent=transcription
```

Currently, the API of both `OpenAIRealtimeWebSocket.azure()` and `OpenAIRealtimeWS.azure()` does not support passing this parameter.

As a workaround, one can manually call `OpenAIRealtimeWebSocket` passing their own `onURL` implementation.

Can it be added?

Thanks,

Yehoshua

### Additional context

_No response_

sarptandoven:
## Summary
- Add transcription-only connections to the supported stable Realtime browser and Node.js helpers.
- Use the OpenAI GA endpoint with `intent=transcription` and no connection-level model.
- Migrate all stable Azure targets to `/openai/v1/realtime`: model sessions use `model=<deployment>`, transcription uses `intent=transcription`, and sideband calls use `call_id`.
- Remove obsolete preview `api-version` and `deployment` query parameters from stable Azure connections while preserving authentication and credential redaction.
- Leave removed beta helpers untouched and document the supported GA client.

## Why
Fixes #1511. The backend no longer supports Realtime beta, and Microsoft deprecated its preview Realtime API on April 30, 2026.

## Verification
- All 1,621 handwritten SDK tests pass.
- Browser and Node transports, API keys and bearer tokens, deployment defaults and overrides, sideband sessions, transcription, and unchanged beta behavior.
- Lint, source and published-package TypeScript checks, and full CJS/ESM build.
- Built-package URLs match current Azure GA documentation.

Azure WebSocket documentation: https://learn.microsoft.com/en-us/azure/foundry/openai/how-to/realtime-audio-websockets
Azure migration guide: https://learn.microsoft.com/en-us/azure/foundry/openai/how-to/realtime-audio-preview-api-migration-guide


src/realtime/internal-base.ts
```diff
@@ -151,12 +160,18 @@ export function buildRealtimeURL(
...
+        url.searchParams.set('intent', intent);
+      }
     }
   } else {
     if (hasCallID) {
       url.searchParams.set('call_id', config.callID!);
     } else {
       url.searchParams.set('model', config.model!);
+      if (intent) {
+        url.searchParams.set('intent', intent);
```
HAYDEN-OAI: i think this still makes transcription connections fail for the standard OpenAI path. intent only exists on the model variant, so this branch always emits both model and intent=transcription. the realtime API rejects transcription sessions when a model query parameter is present (invalid_model); the documented URL is just ...?intent=transcription, with the transcription model selected in session.update. could we model transcription as a third connection variant that emits intent without model or deployment? thanks!

sarptandoven:
thanks, that was right. i changed this to model transcription as a separate connection variant instead of adding `intent` to the model/deployment variant.

updated behavior:
- standard OpenAI transcription now builds `...?intent=transcription`, with no `model` query param
- Azure transcription builds with `intent=transcription` and no `deployment` query param
- model/deployment and sideband `callID` connections keep their existing URL shapes
- `model + intent`, `deploymentName + intent`, and `callID + intent` are rejected instead of silently producing an invalid mixed target

I also updated the realtime tests to cover the intent-only URLs and the invalid mixed-target cases.

HAYDEN-OAI (dismissed):
Reviewed and verified transcription-only connections across stable/beta browser and Node transports, Azure static/token authentication and credential redaction, mutually exclusive target validation, current documentation, the full SDK test suite, both published TypeScript versions, and the packed npm artifact.

HAYDEN-OAI (approved):
Verified against production Realtime backend handlers and current Microsoft GA documentation. Stable OpenAI and Azure transcription use supported endpoints; all stable Azure model/call/transcription targets normalize to /openai/v1/realtime without preview parameters or beta markers. Browser/Node transports, auth/redaction, deployment overrides, full 1,635-test suite, lint, source/published typechecks, and production builds pass. Removed beta API remains untouched.

PR #1978, opened 2026-07-06
fix streaming accept header

### LINKED ISSUE #375: Inquiry about the 'Accept' header

sarptandoven:
summary
- set the default `Accept` header to `text/event-stream` when request options enable streaming
- keep `application/json` for non-streaming requests
- preserve caller-provided `Accept` headers for custom streaming transports

why
- closes #375
- streaming responses are returned as `text/event-stream`, and some OpenAI-compatible gateways validate the request `Accept` header before returning an SSE response
- callers can already work around this with a manual header override; this makes the default client behavior match the expected response type

validation
- `pnpm exec jest tests/index.test.ts -t 'accept header' --runInBand`
- `pnpm exec jest tests/index.test.ts --runInBand`
- `pnpm exec tsc --noEmit --pretty false`
- `git diff --check`
- `pnpm lint`
- `pnpm test`

### DIFF
--- a/src/client.ts
+++ b/src/client.ts
@@ -1157,7 +1157,7 @@ export class OpenAI {
     const headers = buildHeaders([
       idempotencyHeaders,
       {
-        Accept: 'application/json',
+        Accept: options.stream ? 'text/event-stream' : 'application/json',
         'User-Agent': this.getUserAgent(),
         'X-Stainless-Retry-Count': String(retryCount),
         ...(options.timeout ? { 'X-Stainless-Timeout': String(Math.trunc(options.timeout / 1000)) } : {}),
--- a/tests/index.test.ts
+++ b/tests/index.test.ts
@@ -27,6 +27,26 @@ describe('instantiate client', () => {
       adminAPIKey: 'My Admin API Key',
     });
 
+    test('uses an event-stream accept header for streaming requests', async () => {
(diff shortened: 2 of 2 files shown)


HAYDEN-OAI:
Thanks for the suggestion. The SDK Accept: application/json default is established behavior, and the OpenAI API does not require Accept: text/event-stream to stream responses. Changing the default can break compatible gateways, proxies, caches, and custom servers. Callers that require SSE negotiation can explicitly supply headers: { Accept: "text/event-stream" }, so we are keeping the existing default.

PR #1979, opened 2026-07-06
docs: document assistant stream failures

### LINKED ISSUE #959: Getting `Unhandled Rejection` event, randomly.

### Describe the bug

A Node.js library is throwing an unhandled promise rejection error due to improper error handling.

### To Reproduce

How to reproduce this issue:

- run and stream, then add an event listener to the stream.
- Attempt to run and stream again on the same thread ID. This will trigger an unhandled rejection, causing your Node server to crash if there is no global error handler in place.

Ideally, this error should be handled with stream.on('error', myCustomErrorHandler).

### Code snippets

_No response_

### OS

macOs

### Node version

v18.19.0

### Library version

openai 4.52.7

HAYDEN-OAI:
## Summary

- document the Assistants stream `abort` event alongside `error` and `end`
- explain that cancellation through an `AbortSignal` or `.abort()` emits `abort`
- add a safe example that registers both listeners and catches the `finalRun()` promise

## Why

Issue #959 reports unhandled rejections while retrying or cancelling Assistant streams. The current helper already separates API and SDK failures from cancellations, but the Assistants event docs only showed `error` and did not explain that promise-returning helpers still need their own `await` or `catch`.

## Impact

Users now have an explicit pattern for handling stream failures before handing a stream to another consumer, including client disconnect and cancellation paths. This clarifies the existing runtime behavior without changing stream semantics.

## Validation

- `pnpm run format`
- `pnpm run lint`

Resolves #959

### DIFF
--- a/helpers.md
+++ b/helpers.md
@@ -286,12 +286,31 @@ More information on tools can be found here [Tools](https://platform.openai.com/
 
 ```ts
 .on('error', (error: OpenAIError) => ...)
+.on('abort', (error: APIUserAbortError) => ...)
 .on('end', () => ...)
 ```
 
-The `error` event is emitted when the stream encounters an API or SDK error. The `end` event is the last SDK
-event emitted when the stream finishes, including after an error or abort. The raw `[DONE]` marker is consumed
-by the SDK rather than emitted through the `event` listener.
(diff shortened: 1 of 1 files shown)


apcha-oai (approved):
LGTM, assistants is going to be removed soon though.

PR #1987, opened 2026-07-08
feat: add fromReadableStream to ResponseStream

### LINKED ISSUE #1376: Add fromReadableStream or similar to the ResponseStream class

Both the ChatCompletionStream and AssistantStream classes have a fromReadableStream method that the new ResponseStream class is missing. 

I'm looking for something similar to allow my client side web application to request a response stream from my server and convert that to a ResponseStream instance.

For example:

```ts
const sendMessage = async (input) => {
    controller.current = new AbortController();
    const response = await fetch(`${API_URL}/responses`, {
      method: "POST",
      headers: {
        "Content-Type": "application/json",
      },
      body: JSON.stringify({
        input,
      }),
      signal: controller.current.signal,
    })
    const stream = ResponseStream.fromReadableStream(response.body)
    handleReadableStream(stream)
  };
```

### Additional context

_No response_

morgan-coded:

## Changes being requested

`ResponseStream` does not have the same readable-stream rehydration helper that `ChatCompletionStream` and `AssistantStream` expose, so callers proxying newline-delimited response stream events cannot turn them back into a `ResponseStream`. This adds `ResponseStream.fromReadableStream()` and a protected `_fromReadableStream()` path that reads `ResponseStreamEvent` values through `Stream.fromReadableStream()` and feeds them through the existing response-stream event handling. Generated response resources stay unchanged.

## Additional context & links

Fixes #1376


HAYDEN-OAI:
pushed a follow-up fix for cancellation. Stream.fromReadableStream now holds the source reader and cancels it when the runner aborts, so ResponseStream.fromReadableStream will not hang on a stalled proxy body. i also added a regression test that verifies the source stream is canceled and done() rejects with APIUserAbortError.

HAYDEN-OAI (approved):

PR #1988, opened 2026-07-09
docs: clarify strict Zod function schemas

### LINKED ISSUE #1119: `zodFunction` helper rejects `.default()` and `.min()` zod methods for validation tool call schemas

### Describe the bug

When using the `zodFunction` helper from the official OpenAI Node library's Zod utilities (found [here](https://github.com/openai/openai-node/blob/master/src/helpers/zod.ts#L82)), Zod schemas containing `.default()` and `.min()` methods are rejected with errors like `'default' is not permitted'` and `'minLength' is not permitted'`. These methods work fine when using the example `zodFunction` utility provided [here](https://github.com/openai/openai-node/blob/master/examples/tool-call-helpers-zod.ts#L118), but not with the exported helper function.

### To Reproduce

1. **Working Example:**

```ts
import { z, ZodSchema } from 'zod';
import { JSONSchema } from 'openai/lib/jsonschema';
import { zodToJsonSchema } from 'zod-to-json-schema';
import { RunnableToolFunctionWithParse } from 'openai/lib/RunnableFunction';

export const TimeParams = z.object({
  timeZone: z.string().describe('Time zone of the location').default('Europe/Belgrade'),
});
type TimeParams = z.input<typeof TimeParams>;

const getCurrentTime = async (params: TimeParams) => {
  try {
    const response = await fetch(`https://worldtimeapi.org/api/timezone/${params.timeZone}`);

    const data = await response.json();
    const importantTimeData = {
      abbreviation: data.abbreviation,
      datetime: new Date(data.datetime).toLocaleString('en-US', {
        weekday: 'long',
        year: 'numeric',
        month: 'long',
        day: 'numeric',
        hour: 'numeric',
        minute: 'numeric',
        timeZone: data.timezone,
        timeZoneName: 'short',
      }),
      timezone: data.timezone,
    };

    return {
      current_time: importantTimeData,
    };
  } catch (error) {
    return {
      error: getErrorMessage(error, 'Failed to fetch the current time'),
    };
  }
};

function zodFunction<T extends object>({
  function: fn,
  schema,
  description = '',
  name,
}: {
  function: (args: T) => Promise<object>;
  schema: ZodSchema<T>;
  description?: string;
  name?: string;
}): RunnableToolFunctionWithParse<T> {
  return {
    type: 'function',
    function: {
      function: fn,
      name: name ?? fn.name,
      description: description,
      parameters: zodToJsonSchema(schema) as JSONSchema,
      parse(input: string): T {
        const obj = JSON.parse(input);
        return schema.parse(obj);
      },
    },
  };
}

export const chatCompletionsTools = [
  zodFunction({
    function: getCurrentTime,
    name: 'getCurrentTime',
    schema: TimeParams,
    description: 'Get the current time in a given location',
  }),
];
```

2. **Non-Working Example (Using the Official `zodFunction` Helper):**

```ts
import { z } from 'zod';
import { zodFunction } from 'openai/helpers/zod';

export const TimeParams = z.object({
  timeZone: z.string().describe('Time zone of the location').default('Europe/Belgrade'),
});
type TimeParams = z.input<typeof TimeParams>;

const getCurrentTime = async (params: TimeParams) => {
  try {
    const response = await fetch(`https://worldtimeapi.org/api/timezone/${params.timeZone}`);

    const data = await response.json();
    const importantTimeData = {
      abbreviation: data.abbreviation,
      datetime: new Date(data.datetime).toLocaleString('en-US', {
        weekday: 'long',
        year: 'numeric',
        month: 'long',
        day: 'numeric',
        hour: 'numeric',
        minute: 'numeric',
        timeZone: data.timezone,
        timeZoneName: 'short',
      }),
      timezone: data.timezone,
    };

    return {
      current_time: importantTimeData,
    };
  } catch (error) {
    return {
      error: getErrorMessage(error, 'Failed to fetch the current time'),
    };
  }
};

export const chatCompletionsTools = [
  zodFunction({
    function: getCurrentTime,
    name: 'getCurrentTime',
    parameters: TimeParams, 
    description: 'Get the current time in a given location',
  }),
];
```

### Code snippets

_No response_

### OS

Windows 11

### Node version

v20.16.0

### Library version

4.67.1

HAYDEN-OAI:
## Summary

- document that `zodFunction()` always creates a strict function tool
- call out that `.default()` and string `.min()`/`.max()` emit keywords outside the strict schema subset
- point users who need those Zod behaviors to a regular non-strict tool with local Zod parsing

## Why

The report in #1119 compared `zodFunction()` with a custom helper that appeared to work. The difference is that the custom helper sends a non-strict tool, while `zodFunction()` deliberately opts into strict Structured Outputs. In strict mode, `default`, `minLength`, and `maxLength` can be rejected by the API.

## Impact

Users now get an actionable explanation before reaching the API error and can choose the non-strict tool path when their Zod schema depends on defaults or string length constraints. This avoids silently stripping constraints from the model-facing schema while preserving the existing strict helper behavior.

## Validation

- `pnpm run format`
- `pnpm run lint`
- `git diff --check`

Resolves #1119

### DIFF
--- a/helpers.md
+++ b/helpers.md
@@ -64,6 +64,10 @@ Some Zod behavior cannot be represented in the schema sent to the API:
   Custom refinements and transforms still run when the SDK parses the response, but
   some schemas that cannot be converted, including bare transforms in Zod v4, cause
   the helper to throw before a request is made.
+- `.default()` emits the unsupported `default` keyword, and string `.min()` or
+  `.max()` emit the unsupported `minLength` or `maxLength` keywords. Because
+  `zodFunction()` always creates a strict function tool, use a regular non-strict
+  tool and parse its arguments with Zod yourself when you need these behaviors.
 - TypeScript comments are not available at runtime and are not sent to the API. Use
   `.describe()` for field descriptions or the helper's `description` option for a
   top-level response format or function description.
--- a/helpers.md
+++ b/helpers.md
@@ -64,10 +64,10 @@ Some Zod behavior cannot be represented in the schema sent to the API:
   Custom refinements and transforms still run when the SDK parses the response, but
   some schemas that cannot be converted, including bare transforms in Zod v4, cause
   the helper to throw before a request is made.
-- `.default()` emits the unsupported `default` keyword, and string `.min()` or
-  `.max()` emit the unsupported `minLength` or `maxLength` keywords. Because
-  `zodFunction()` always creates a strict function tool, use a regular non-strict
-  tool and parse its arguments with Zod yourself when you need these behaviors.
+- `zodFunction()` always creates a strict function tool. `.default()` emits
+  `default`, but strict conversion still marks the property as required, so it does
+  not make model-generated tool arguments optional. String `.min()` and `.max()`
+  emit `minLength` and `maxLength`, which are not supported by fine-tuned models.
 - TypeScript comments are not available at runtime and are not sent to the API. Use
   `.describe()` for field descriptions or the helper's `description` option for a
   top-level response format or function description.


jbeckwith-oai (approved):
LGTM

helpers.md
```diff
@@ -64,6 +64,10 @@ Some Zod behavior cannot be represented in the schema sent to the API:
   Custom refinements and transforms still run when the SDK parses the response, but
   some schemas that cannot be converted, including bare transforms in Zod v4, cause
   the helper to throw before a request is made.
+- `.default()` emits the unsupported `default` keyword, and string `.min()` or
```
jbeckwith-oai: This is no longer accurate for the current API. The Structured Outputs guide now documents string length constraints as unsupported only for fine-tuned models, and live calls with the exact schema generated by both Zod v3 and v4 were accepted by `gpt-4.1-mini`, `gpt-4o-2024-08-06`, and `gpt-4o-mini-2024-07-18` (including `default`, `minLength`, and `maxLength`). The original 2024 rejection is therefore not reproducible today. Please avoid categorically labeling these keywords unsupported or recommending a non-strict tool; at minimum this needs to distinguish fine-tuned models and explain that `.default()` does not make the field optional because the strict conversion adds it to `required`. Current reference: https://developers.openai.com/api/docs/guides/structured-outputs#some-type-specific-keywords-are-not-yet-supported

jbeckwith-oai (changes requested):
I reproduced the schema generation and tested the proposed diagnosis against the current API. Both `zod/v3` and `zod/v4` generate strict function schemas containing `default`, `minLength`, and `maxLength`, with defaulted fields added to `required`. However, the original 2024 API rejection is no longer reproducible: live forced-tool calls using that exact generated schema succeeded on `gpt-4.1-mini`, `gpt-4o-2024-08-06`, and `gpt-4o-mini-2024-07-18`. I also tested `default` and `minLength`/`maxLength` separately on `gpt-4.1-mini`; all were accepted.

The current Structured Outputs guide likewise lists `minLength`/`maxLength` as unsupported only for fine-tuned models, not standard models: https://developers.openai.com/api/docs/guides/structured-outputs#some-type-specific-keywords-are-not-yet-supported. As written, this PR would tell users that a working strict schema is unsupported and direct them to give up strict tool calling. It also conflates `.default()` being emitted/accepted with optionality: strict conversion makes the property required, so a default generally cannot be exercised by a conforming model-generated call.

Local validation: `./node_modules/.bin/jest tests/helpers/zod.test.ts --runInBand` (22/22 passed), `./node_modules/.bin/prettier --check helpers.md` (passed), and `git diff --check origin/main...HEAD` (passed). The patch is docs-only and introduces no runtime/type ABI break, but the behavioral guidance is materially incorrect and should be revised before merge. The original issue should likely be closed as no longer reproducible, or any docs should narrowly describe fine-tuned-model limitations and default semantics.

helpers.md
```diff
@@ -64,6 +64,10 @@ Some Zod behavior cannot be represented in the schema sent to the API:
...
+  `.max()` emit the unsupported `minLength` or `maxLength` keywords. Because
+  `zodFunction()` always creates a strict function tool, use a regular non-strict
+  tool and parse its arguments with Zod yourself when you need these behaviors.
```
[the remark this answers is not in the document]
HAYDEN-OAI: good catch, i removed the non-strict fallback entirely, so this no longer suggests a path that would fail under `.parse()`.

helpers.md
HAYDEN-OAI: thanks, updated this to scope `minLength`/`maxLength` to fine-tuned models and clarify that `.default()` still leaves the property required.

jbeckwith-oai (approved):

PR #1990, opened 2026-07-13
fix(runTools): sanitize replayed messages

### LINKED ISSUE #1161: Calling openai.beta.chat.completions.runTools with gpt-4-32k returns error: "Additional properties are not allowed ('parsed', 'refusal' were unexpected)"

### Describe the bug

I am using the example from `https://github.com/openai/openai-node/blob/master/examples/tool-call-helpers.ts` and using OpenAI node SDK v.4.69.0.

When I execute the example using gpt-4-32k model, I receive the following error: `BadRequestError: 400 Additional properties are not allowed ('parsed', 'refusal' were unexpected) - 'messages.2'`

image

When I execute the example using gpt-4o model, the example works correctly.

image

I have been testing with different OpenAI node SDK versions. The last one where runTools actually works with this example is **v.4.54.0**. Once you upgrade to 4.55.x the error starts happening.

It could be related with the addition of Structured outputs that was added in 4.55.x.

I am using GPT models deployed in an Azure OpenAI service.

Maybe I should be doing something different to make it work with gpt-4-32k? (I haven't tested with `gpt-4-1106-preview` that is used in the example, but this should also work with `gpt-4-32k`, right?).

### To Reproduce

- Use the example available at `https://github.com/openai/openai-node/blob/master/examples/tool-call-helpers.ts` 
- Configure the OpenAI instance to use a GPT4 32k model. 
- Execute the example

### Code snippets

_No response_

### OS

macOS

### Node version

v20.18.0

### Library version

openai v.4.69.0

HAYDEN-OAI:

## Changes being requested

- sanitize assistant messages before `runTools` replays them in follow-up requests
- preserve enriched `runner.messages`, events, and parsed completions for callers while removing helper-only `parsed` and `parsed_arguments`, null `refusal`, and other response-only assistant fields from outbound payloads
- update the streaming and non-streaming runner expectations so the regression stays covered

## Additional context & links

Fixes #1161.

`runTools` currently stores parsed completion messages in its history and then sends that history back unchanged on the next tool-loop request. Azure and other strict-compatible endpoints reject the helper-only fields with errors such as `Additional properties are not allowed (parsed, refusal were unexpected)`.

The fix sanitizes only the outbound copy, so existing parsed runner results remain available to users.

Verification:

- `pnpm test --runInBand tests/lib/ChatCompletionRunFunctions.test.ts`
- `pnpm lint`
- `pnpm test --runInBand`


jbeckwith-oai (approved):
Reviewed the request-message sanitization across streaming and non-streaming runTools paths. The patch preserves parsed runner/event data while replaying only request-valid assistant fields, including audio and both function/custom tool calls.\n\nVerified locally:\n- pnpm test --runInBand tests/lib/ChatCompletionRunFunctions.test.ts (25 passed)\n- pnpm test --runInBand (1,195 passed, 1 skipped)\n- pnpm build\n- pnpm lint

PR #1993, opened 2026-07-13
feat(zod): support schema definitions

### LINKED ISSUE #1285: Allow specifying `definitions` as a `zodResponseFormat` param

Allow specifying `schemaDefinitions` as a `zodResponseFormat` param

Proposed signature
```ts
const schemaDefinitions = {
  foo: z.enum('one', 'two', 'three', ....)  // 200 elements
  bar: z.enum('a', 'b', 'c', ...)  // 200 elements
)

const responseFormat = zodResponseFormat(
  z.object({
    foo: schemaDefinitions.foo
    foos: z.array[schemaDefinitions.foo]
    bar: schemaDefinitions.foo
    bars: z.array[schemaDefinitions.bar]
  }),
 { name: 'fooBar', schemaDefinitions }
)
```

### Additional context

❌ Without definitions this schema produces 800 enums, which is above the 500 enums OpenAI limit. 
First occurrence of a def is left in place, starting with the second one it's copied to definitions and a reference is used. 

✅ `zod-to-json-schema` library has a definitions option that is used to supply definitions, and avoid duplication – resulting in 400 enums.

Alternative approach would be to detect shared definitions, move out to definitions, and replace with refs automatically. This would be a significantly more complex, but easier to use solution. Probably not worth the work for now, since everything is in place to just expose the zod-`to-json-schema`'s `definitions` option

HAYDEN-OAI:

## Changes being requested

Fixes #1285.

`zodResponseFormat()` now accepts an optional `schemaDefinitions` map in its existing third argument. Callers can name shared Zod schemas that should be extracted into JSON Schema `definitions`, allowing repeated large schemas such as enums to be referenced instead of copied at every use site.

The helper previously forwarded response-format props but had no way to pass definitions into schema conversion. Reusing two 200-value enums could therefore emit more than the API's 500-enum limit even though the logical schema only contained 400 unique values.

The implementation keeps the same public option for both supported Zod paths:

- Zod v3 forwards the supplied map to the vendored `zod-to-json-schema` definitions option.
- Zod v4 registers the supplied schemas with their definition names before native JSON Schema conversion.

`schemaDefinitions` is removed from the outgoing response-format object after conversion, and callers who do not provide it keep the existing generated schema shape.

The regression test mirrors the issue's two reused 200-value enums for both Zod v3 and v4. It verifies that only 400 enum values are emitted and that every generated reference resolves.

## Additional context & links

- Issue: #1285
- Validation:
  - `pnpm test --runInBand tests/helpers/zod.test.ts`
  - `pnpm lint`
  - `pnpm test --runInBand`


src/helpers/zod.ts
```diff
@@ -37,20 +37,45 @@ type InferZodType<T extends ZodTypeLike> =
...
+  options: { name: string; schemaDefinitions?: ZodSchemaDefinitions | undefined },
+): Record<string, unknown> {
   return _zodToJsonSchema(schema, {
     openaiStrictMode: true,
     name: options.name,
     nameStrategy: 'duplicate-ref',
     $refStrategy: 'extract-to-root',
     nullableStrategy: 'property',
+    ...(options.schemaDefinitions ?
+      { definitions: options.schemaDefinitions as unknown as Record<string, z3.ZodType> }
```
jbeckwith-oai, line 112: **[P2] Prevent definition keys from colliding with the response name**

On the v3 path, the supplied definitions seed the converter's `seen` map while `options.name` is also used as the root definition path. A natural collision produces a schema with no concrete body:

```ts
const Root = z.object({ a: z.string() });
zodResponseFormat(Root, 'root', { schemaDefinitions: { root: Root } });
```

This emits a top-level `#/definitions/root` ref whose `definitions.root` value is another ref to itself. Please reserve/rename/validate this key (or otherwise avoid sharing the root path) and add a v3 regression test.

src/helpers/zod.ts
```diff
@@ -37,20 +37,45 @@ type InferZodType<T extends ZodTypeLike> =
...
 }
 
-function zodV4ToJsonSchema(schema: ZodV4Schema): Record<string, unknown> {
+function zodV4ToJsonSchema(
+  schema: ZodV4Schema,
+  options: { schemaDefinitions?: ZodSchemaDefinitions | undefined } = {},
+): Record<string, unknown> {
+  const metadata = options.schemaDefinitions ? z4.registry<Record<string, unknown>>() : undefined;
+  for (const [name, definition] of Object.entries(options.schemaDefinitions ?? {})) {
+    metadata?.add(definition as unknown as z4.ZodType, { id: name });
```
jbeckwith-oai, line 125: **[P2] Escape or reject JSON Pointer metacharacters in definition names**

Both converter paths use these names directly in `$ref`. For example, `{ schemaDefinitions: { 'foo/bar~baz': Foo } }` emits `#/definitions/foo/bar~baz` while the actual definition key is the literal `foo/bar~baz`; a standards-compliant JSON Pointer resolver interprets that ref as nested path segments. Please either constrain names to a safe character set or escape `~` as `~0` and `/` as `~1`, with a regression test that resolves the pointer rather than slicing the prefix.

jbeckwith-oai (changes requested):
The main implementation is focused and the intended enum-deduplication case works for both Zod v3 and v4. I’m requesting changes for two definition-name edge cases in the new public map that currently generate invalid schemas.

Validated locally at `16b870f0`:
- `pnpm test --runInBand tests/helpers/zod.test.ts` — 28/28 passed
- `pnpm lint` — passed, including build, types, package checks, and JSR dry run
- additional Zod v3/v4/v4-mini runtime probes

The ordinary PR checks are green. OkTest is red for an unrelated generated fixture mismatch (`InputTokensDetails.cache_write_tokens`), before its compatibility suite runs.

src/helpers/zod.ts
jbeckwith-oai, line 112: Blocking: guard collisions between the response-format `name` and supplied definition keys. The v3 converter uses `name` as the root definition key in the same namespace as `definitions`. For example, `zodResponseFormat(z.object({ a: Shared, b: Shared }), 'root', { schemaDefinitions: { root: Shared } })` emits both property refs as `#/definitions/root`, but `definitions.root` is overwritten with the root object. The result is unintended recursion and no longer describes `Shared`; v4 produces the expected schema, so the two supported paths also diverge. Please reject/reserve collisions or choose a non-colliding internal root name, and add a v3/v4 regression test.

src/helpers/zod.ts
jbeckwith-oai, line 125: Blocking: definition names are typed as arbitrary strings, but they become JSON Pointer tokens without escaping. With `{ schemaDefinitions: { 'with/slash': Shared } }`, both paths emit `$ref: '#/definitions/with/slash'` while storing the schema at `definitions['with/slash']`; the pointer resolves as nested keys and is dangling. `~` has the same problem. Please JSON-Pointer-escape tokens (`~0` / `~1`) or validate and reject unsafe names, and make the ref-resolution test follow actual JSON Pointer semantics so it catches this case.

jbeckwith-oai (dismissed):
The core approach is sound and the normal v3/v4 cases are well covered, but I found two correctness gaps in the newly accepted `schemaDefinitions` names that can produce invalid or semantically wrong schemas. Please address these before merge.

Validation performed:
- focused Zod helper suite: 28/28 passed
- full repository suite: 1,189 passed (1 skipped)
- `pnpm lint` and `pnpm build`: passed, including type checks, publint, package loading, and JSR dry-run
- GitHub build/lint/tests/ecosystem/examples/breaking-change checks: passed

The separate OkTest failure is unrelated: its Node bridge fixtures are missing the newly required `cache_write_tokens` field, and the run fails before exercising this change.

src/helpers/zod.ts
HAYDEN-OAI: fixed — the v3 converter now chooses a non-colliding internal root name while preserving both the response-format name and supplied definition key. added v3/v4 regression coverage that resolves every emitted ref using actual JSON Pointer semantics.

HAYDEN-OAI: fixed — the v3 converter now chooses a non-colliding internal root name while preserving both the response-format name and supplied definition key. added v3/v4 regression coverage that resolves every emitted ref using actual JSON Pointer semantics.

src/helpers/zod.ts
HAYDEN-OAI: fixed — supplied definition keys remain literal, while reference tokens now escape ~ and / as ~0 and ~1. the v3/v4 regression uses foo/bar~baz and follows the emitted pointer to verify it resolves correctly.

HAYDEN-OAI: fixed — supplied definition keys remain literal, while reference tokens now escape ~ and / as ~0 and ~1. the v3/v4 regression uses foo/bar~baz and follows the emitted pointer to verify it resolves correctly.

src/helpers/zod.ts
```diff
@@ -37,20 +37,92 @@ type InferZodType<T extends ZodTypeLike> =
...
+type ZodResponseFormatProps = Omit<ResponseFormatJSONSchema.JSONSchema, 'schema' | 'strict' | 'name'> & {
+  /**
+   * Schemas to extract into the generated JSON Schema definitions.
+   * Use this to reuse large shared schemas instead of inlining them at every occurrence.
+   */
+  schemaDefinitions?: ZodSchemaDefinitions | undefined;
+};
+
+function escapeJsonPointerToken(token: string): string {
+  return token.replace(/~/g, '~0').replace(/\//g, '~1');
```
jbeckwith-oai: **[P2] URI-encode definition names used in `$ref` fragments**

This escapes the JSON Pointer token layer, but the pointer is embedded in a URI fragment and must also preserve literal percent sequences through fragment decoding. With `schemaDefinitions: { ["foo%2Fbar"]: Shared }`, both v3 and v4 emit `#/definitions/foo%2Fbar` and store the literal key `foo%2Fbar`; a conforming resolver percent-decodes the fragment first, interprets `%2F` as `/`, and looks for nested `definitions.foo.bar`, so the ref is dangling. Please URI-encode the escaped token (or reject names requiring it) and add v3/v4 coverage that resolves a percent-containing key.

src/helpers/zod.ts
```diff
@@ -37,20 +37,92 @@ type InferZodType<T extends ZodTypeLike> =
...
-function zodV3ToJsonSchema(schema: z3.ZodType, options: { name: string }): Record<string, unknown> {
-  return _zodToJsonSchema(schema, {
+type ZodSchemaDefinitions = Record<string, ZodTypeLike>;
+
+type ZodResponseFormatProps = Omit<ResponseFormatJSONSchema.JSONSchema, 'schema' | 'strict' | 'name'> & {
+  /**
+   * Schemas to extract into the generated JSON Schema definitions.
+   * Use this to reuse large shared schemas instead of inlining them at every occurrence.
+   */
+  schemaDefinitions?: ZodSchemaDefinitions | undefined;
```
jbeckwith-oai, line 47: **[P2] Reject or safely preserve prototype-magic definition keys**

Because this is an arbitrary string-keyed record, callers can supply an own computed `__proto__` key: `{ schemaDefinitions: { ["__proto__"]: Shared } }`. On both v3 and v4 the generated property refs are `#/definitions/__proto__`, but `definitions` has no own `__proto__` entry because the converters assemble it with ordinary objects/assignment; every emitted ref is therefore dangling. Please validate/reject this key or ensure the generated definitions record preserves it safely, with a regression for both paths.

jbeckwith-oai (changes requested):
The two previously reported name issues are fixed, and the normal enum-deduplication path is sound. However, the public map still accepts arbitrary string keys that produce dangling `$ref`s in two reproducible cases on both Zod v3 and v4. Please either encode/preserve these keys correctly or reject them explicitly, with regression coverage for both converter paths.

Validation at `bbd5c857`: focused Zod suite 32/32 passed; all full-suite Jest assertions passed (1,201 passed, 1 skipped), although the command exited nonzero afterward because the existing mock-server cleanup passed multiple newline-separated PIDs to `kill`; full `pnpm lint` passed (format, ESLint, CJS/ESM and Deno builds, TypeScript, ATTW, publint, JSR dry-run). Current CI and OkTest 237/237 are green.

src/helpers/zod.ts
HAYDEN-OAI: fixed in 3305aa5 — definition ref tokens are now JSON Pointer-escaped and then URI-encoded, so a literal `foo%2Fbar` key emits `#/definitions/foo%252Fbar`. added v3/v4 regression coverage that resolves the emitted ref through URI-fragment decoding.

src/helpers/zod.ts
HAYDEN-OAI: fixed in 3305aa5 — `schemaDefinitions` now rejects an own `__proto__` key before either converter runs. added v3/v4 regression coverage that asserts the explicit error.

jbeckwith-oai (approved):
Re-reviewed the exact current head (`3305aa52`) after the prior change request. Both dangling-reference findings are fixed: URI-fragment tokens are now JSON Pointer-escaped and URI-encoded (`foo%2Fbar` emits `foo%252Fbar` and resolves to the literal key), and an own `__proto__` definition key is rejected before either converter runs. The thread-aware review state shows both prior threads resolved.

I verified the fixes on Zod v3, v4, and v4-mini, including adjacent `/`, `~`, `#`, spaces, Unicode, `constructor`, and `prototype` names. Local validation passed: focused Zod suite 36/36 and full `pnpm lint` (Prettier, ESLint, CJS/ESM and Deno builds, TypeScript, ATTW, publint, JSR dry-run). Current CI is green across Node 20/22/24/26.2, ecosystem tests, Agents SDK regressions, breaking-change detection, examples, and CodeQL; OkTest is 237/237. No substantive issues remain.

PR #1998, opened 2026-07-14
fix(client): handle empty JSON response body without content-length

LeSingh1:
## Problem

`defaultParseResponse` (`src/internal/parse.ts`) already special-cases an empty JSON response body:

```ts
const contentLength = response.headers.get('content-length');
if (contentLength === '0') {
  // if there is no content we can't do anything
  return undefined as T;
}

const json = await response.json();
```

That guard only fires when the server sends an explicit `content-length: 0` header. When the header is **absent** — which is common over HTTP/2, with chunked transfer encoding, or behind gateways/proxies and OpenAI-compatible servers — an empty body with a JSON `content-type` falls through to `response.json()` and throws:

```
SyntaxError: Unexpected end of JSON input
```

So two byte-for-byte identical empty responses parse differently depending solely on whether the server happened to include `content-length: 0`. The existing guard shows this case is meant to be handled gracefully; it's just incomplete.

### Reproduction

```ts
const client = new OpenAI({
  apiKey: '...',
  // An empty body with a JSON content-type and no `content-length` header.
  // (`new Response('')` in undici sets no content-length.)
  fetch: async () => new Response('', { headers: { 'Content-Type': 'application/json' } }),
});

await client.request({ path: '/foo', method: 'get' });
// Before: throws `SyntaxError: Unexpected end of JSON input`
// After:  resolves to `undefined`
```

## Fix

Read the body as text first and treat an empty payload the same as an explicit `content-length: 0`. Non-empty bodies are still parsed with `JSON.parse`, so malformed JSON continues to throw as before.

```ts
const bodyText = await response.text();
if (!bodyText) {
  return undefined as T;
}
const json = JSON.parse(bodyText);
```

## Verification

Added a `response parsing` test group in `tests/index.test.ts` covering:
- empty JSON body **without** `content-length` → resolves to `undefined` (the fixed case; fails on `main` with `SyntaxError: Unexpected end of JSON input`)
- empty JSON body **with** `content-length: 0` → resolves to `undefined` (existing behavior preserved)
- non-empty JSON body → parsed correctly
- malformed non-empty JSON body → still throws

```
$ npx jest tests/index.test.ts
Tests:       68 passed, 68 total
```

`npx tsc --noEmit`, `eslint`, and `prettier --check` all pass on the changed files.


HAYDEN-OAI:
Thanks for the contribution! This fixes the SDK’s shared response parser for empty JSON bodies without a content-length header. I’m moving it to SDK triage because the parser is maintained as SDK runtime code.

HAYDEN-OAI (approved):
Validated against the first-party API schema: four supported Realtime call-control endpoints return empty HTTP 200 responses. The parser now handles absent content-length, bodyless and closed-stream responses while preserving malformed JSON errors, JSON primitives, request IDs, +json media types, and HTTP 204. Contributor tests were moved into handwritten Vitest coverage. All 1,651 tests, lint, source and published-package typechecks, and production CJS/ESM builds pass.

PR #2003, opened 2026-07-17
chore: refresh more ecosystem test dependencies

jbeckwith-oai:
## Summary
- refresh the node-ts-cjs toolchain and move Undici from 7.11.0 to 7.28.0
- upgrade the Vercel fixture to Next 15, AI SDK 7, Edge Runtime 4, and Vercel CLI 55
- migrate the AI streaming sample to the current UI message protocol while continuing to stream from the OpenAI SDK
- override exact vulnerable leaf pins in the Vercel CLI graph with patched releases

These two fixture manifests currently account for 105 of the 115 open Dependabot alerts on main. Both refreshed lockfiles now report zero vulnerabilities. The published openai package dependency graph is unchanged, and the ecosystem fixtures remain independent npm consumers of the packed SDK.

## Testing
- pnpm lint
- npm audit --package-lock-only --ignore-scripts in both fixtures
- Node 24 packed-SDK harness for node-ts-cjs and vercel-edge
- Vercel CLI 55.0.0 startup smoke test

### DIFF
--- a/ecosystem-tests/node-ts-cjs/package-lock.json
+++ b/ecosystem-tests/node-ts-cjs/package-lock.json
@@ -10,141 +10,84 @@
       "dependencies": {
         "formdata-node": "^4.4.1",
         "tsconfig-paths": "^4.0.0",
-        "typescript": "^5.7.3",
-        "undici": "^7.10.0"
+        "typescript": "^5.9.3",
+        "undici": "^7.28.0"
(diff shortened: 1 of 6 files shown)


HAYDEN-OAI (approved):
looks good, thanks for refreshing these fixtures and keeping the streaming sample current!

PR #2007, opened 2026-07-17
fix: address code scanning findings

jbeckwith-oai:
## Summary

- restrict the affected GitHub Actions jobs to explicit read-only repository permissions
- stop Cloudflare Worker, Vercel Edge, and Vercel Node test endpoints from returning stack traces in HTTP 500 responses while retaining server-side error logging
- add focused regressions for each error boundary and a Cloudflare health-check control

## Root cause

Six jobs inherited the repository's default `GITHUB_TOKEN` permissions instead of declaring the minimum required scope. Three ecosystem-test endpoints also reused their server-side stack-trace diagnostics as public error-response bodies.

## Validation

- `./scripts/lint` on Node 24: Prettier, ESLint, build, TypeScript, ATTW, publint, and JSR dry-run all passed
- `./scripts/test` on Node 24: 117 suites and 1,193 tests passed; one test intentionally skipped
- `pnpm tsn ecosystem-tests/cli.ts cloudflare-worker vercel-edge --jobs=1 --verbose`: both packages passed against the locally built `openai-6.48.0` tarball
- focused Cloudflare regression/control tests: 2 passed
- focused Vercel Edge/Node regression tests: 2 passed
- `actionlint` on both changed workflows: passed

### DIFF
--- a/ecosystem-tests/cloudflare-worker/src/worker.ts
+++ b/ecosystem-tests/cloudflare-worker/src/worker.ts
@@ -79,24 +79,18 @@ export default {
 				expectSimilar,
 			});
 
-			let allPassed = true;
-			const results = [];
-
 			for (const { description, handler } of tests) {
 				console.error('running', description);
-				let result;
 				try {
-					result = await handler();
+					await handler();
 					console.error('passed ', description);
 				} catch (error) {
 					console.error('failed ', description, error);
-					allPassed = false;
-					result = error instanceof Error ? error.stack : String(error);
+					return new Response('Internal Server Error', { status: 500 });
 				}
-				results.push(`${description}\n\n${String(result)}`);
 			}
 
-			return new Response(allPassed ? 'Passed!' : results.join('\n\n'));
+			return new Response('Passed!');
 		} catch (error) {
 			console.error(error instanceof Error ? error.stack : String(error));
 			return new Response('Internal Server Error', { status: 500 });
--- a/ecosystem-tests/cloudflare-worker/src/worker.ts
+++ b/ecosystem-tests/cloudflare-worker/src/worker.ts
@@ -99,7 +99,7 @@ export default {
 			return new Response(allPassed ? 'Passed!' : results.join('\n\n'));
 		} catch (error) {
 			console.error(error instanceof Error ? error.stack : String(error));
-			return new Response(error instanceof Error ? error.stack : String(error), { status: 500 });
+			return new Response('Internal Server Error', { status: 500 });
 		}
 	},
 };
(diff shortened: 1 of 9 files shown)


ecosystem-tests/cloudflare-worker/src/worker.ts
```diff
@@ -99,7 +99,7 @@ export default {
 			return new Response(allPassed ? 'Passed!' : results.join('\n\n'));
 		} catch (error) {
 			console.error(error instanceof Error ? error.stack : String(error));
-			return new Response(error instanceof Error ? error.stack : String(error), { status: 500 });
+			return new Response('Internal Server Error', { status: 500 });
```
HAYDEN-OAI, line 96: this protects only the outer catch. when one of the registered test handlers rejects, the inner catch above still assigns error.stack to result, then results.join(...) is returned to the caller with a 200. the same path remains in both vercel handlers, so a deployed test endpoint can still expose stack traces even though codeql no longer reports the three original sinks. could we keep the details in the existing console.error logs and return only a generic failure body?

ecosystem-tests/vercel-edge/tests/edge-error-response.test.ts
```diff
@@ -0,0 +1,22 @@
+import type { NextRequest } from 'next/server';
+import handler from '../src/pages/api/edge-test';
+
+it('does not expose stack traces from unexpected edge runtime errors', async () => {
+  const apiKey = process.env.OPENAI_API_KEY;
+  const consoleError = jest.spyOn(console, 'error').mockImplementation(() => {});
+  delete process.env.OPENAI_API_KEY;
```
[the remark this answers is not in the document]
jbeckwith-oai: Addressed in ff6e8d93. Each error-response regression now saves, clears, and restores both `OPENAI_API_KEY` and `OPENAI_ADMIN_KEY` before invoking the handler, so ambient credentials cannot affect the result or permit live requests. The targeted suites and full CI are green.

ecosystem-tests/cloudflare-worker/src/worker.ts
jbeckwith-oai: Addressed in ae091b49. The inner catches in Cloudflare, Vercel Edge, and Vercel Node keep the full exception in `console.error` but now stop with a generic `500 Internal Server Error`. Regression tests inject a secret-bearing fetch failure without making a live request. Targeted tests, full CI, CodeQL, and OkTest are green.

HAYDEN-OAI (approved):
re-reviewed the latest changes. the earlier credential-isolation and inner error-response issues are both addressed, and i did not find any remaining blockers. looks good to me, thanks!

PR #2013, opened 2026-07-24
ci: add stable Node test matrix gate

jbeckwith-oai:
## Summary

- add a stable `test matrix` check that depends on every Node.js matrix variant
- fail the summary unless the complete matrix succeeds
- give the intentionally skipped internal-PR instance a distinct name so it cannot satisfy branch protection

This lets the `main` ruleset replace version-specific checks such as `test (20)` and `test (26.2.0)` with one stable context after this change lands.

## Validation

- `git diff --check`
- YAML syntax parse
- `actionlint` v1.7.12

## Rollout

Keep the current version-specific required checks until this PR merges and `test matrix` has run successfully on the default workflow. Then replace the four Node-version contexts in the ruleset with `test matrix`.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -127,17 +127,8 @@ jobs:
         run: ./scripts/test
 
   test_matrix:
-    name: >-
-      ${{
-        (github.event_name == 'push' || github.event.pull_request.head.repo.fork)
-        && 'test matrix'
-        || 'test matrix (not run)'
-      }}
-    if: >-
-      ${{
-        always() &&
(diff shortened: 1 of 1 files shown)


HAYDEN-OAI (approved):
Reviewed the stable Node.js matrix gate against the live push and pull-request runs. All four Node versions complete before the test matrix gate, and the skipped internal-PR job cannot satisfy the required check. No blocking findings. Non-blocking: GitHub displays the skipped job naming expression literally. After merging, replace the four version-specific required checks with test matrix while retaining the GitHub Actions integration.

PR #2015, opened 2026-07-25
fix(parser): recognize raw json_schema response format in auto-parsing checks

hsusul:
### Problem

Raw JSON Schema Structured Outputs are supported by both active APIs, but SDK stream finalization only recognized formats carrying SDK-helper branding. Non-streaming parsers already decoded raw schemas, creating inconsistent runtime and TypeScript results.

The wire shapes are distinct:

- Chat Completions: `response_format: { type: 'json_schema', json_schema: { name, schema } }`
- Responses API: `text: { format: { type: 'json_schema', name, schema } }`

### Changes

- Use one shared parseability decision and content parser for raw JSON Schema formats and SDK-branded helpers.
- Parse raw Chat streaming partial/completed events and final Chat/Responses results consistently.
- Preserve existing branded and unbranded custom `$parseRaw` callbacks.
- Infer `unknown` for raw schemas and broad public request unions while retaining precise branded-helper types, including optional/union formats.
- Preserve refusal handling, dedicated length/content-filter errors, incomplete response behavior, strict tool parsing, and existing wire-event contracts.
- Keep all changes in handwritten SDK implementation and Vitest tests.

### Validation

- 1,690 handwritten Vitest tests passed.
- Repository lint and root TypeScript checking passed.
- Complete ESM/CommonJS package build and published-source TypeScript checking passed.
- Built-package probes confirmed both raw JSON parsing and existing unbranded custom parsers.


jbeckwith-oai (dismissed):
Raw-schema recognition is currently added only to the coarse `hasAutoParseableInput()` gate, while the type extractors and streaming parser still use the old brand-only definition. That split leaves two blocking inconsistencies:

1. Both `ExtractParsedContentFromParams` aliases still resolve an unbranded `{ type: "json_schema", ... }` format to `null`. The methods now return parsed JSON at runtime while `message.parsed`, `content[].parsed`, and `output_parsed` are declared as `null`. Please represent raw-schema output with a usable type (for example `unknown` or a JSON value type) in both Chat Completions and Responses, with compile-time coverage.

2. `ChatCompletionStream` still gates partial/final event parsing through `#getAutoParseableResponseFormat()`, which accepts only branded helper formats. I reproduced a raw-schema stream where both `content.delta.parsed` values and `content.done.parsed` were `null`, but `finalChatCompletion().choices[0].message.parsed` was `{ city: "SF" }`. Please unify the parsing decision so raw schemas are handled consistently in stream events too, and add regression coverage.

This is also a maintainability issue: adding the same special-case recognition at only two entry gates creates competing definitions of “auto-parseable.” A canonical predicate/parser path should own the branded-callback versus generic-JSON behavior.

Validation at `0c57565d03bd75c029b3f658b42e6e9adde60588`: the three focused parser/stream suites passed (28 tests, 26 snapshots); changed-file Prettier and ESLint passed; TypeScript passed; package build passed; and `git diff --check` passed. The streaming inconsistency above was reproduced against the built package.

HAYDEN-OAI (approved):
Reviewed current head 0236ed1a7e0b7c8f7cfc5b5dbd3c839282ae98c8 against the actual supported Chat Completions and Responses Structured Outputs schemas, backend handlers, model gates, and streaming event contracts. All three prior review findings are fixed. Raw schemas now parse consistently while preserving branded and unbranded custom parsers, distributive public request typing, refusal/unfinished response behavior, and existing wire event shapes. Verified 1,690 Vitest tests, repository lint, both TypeScript configurations, full ESM/CommonJS build, and built-package raw/custom parser smoke tests.

PR #2017, opened 2026-07-25
fix(responses): add nullish item safety to toResponseInputItem helper

hsusul:
### Problem
When `toResponseInputItem` or `toResponseInputItems` is passed `null`, `undefined`, a primitive, or an item object with a nullish `type` property (for instance, when mapping over sparse arrays or optional fields in stored response history), `toResponseInputItem` attempts to access `item.type` or execute `'created_by' in item` directly. This causes an uncaught `TypeError: Cannot read properties of null (reading 'type')` or `TypeError: Cannot use 'in' operator to search for 'created_by' in null`. Additionally, `ResponseStream.finalResponse()` threw an error message referring to `ChatCompletion` instead of `Response` when called prior to receiving a response snapshot.

### Root Cause
1. `ResponseInputItems.ts` lacked a top-level nullish/object check before evaluating `item.type` and before invoking `stripCreatedBy(item)`.
2. `ResponseStream.ts` line 297 contained a copy-paste error in its exception message string, referencing `ChatCompletion` instead of `Response`.

### Solution
1. Added top-level nullish and non-object guards to `toResponseInputItem` and `stripCreatedBy` in `src/lib/responses/ResponseInputItems.ts`. Items that are nullish, primitives, or have `type: null` / `type: undefined` now safely evaluate to `null` so `toResponseInputItems` can filter them out without crashing.
2. Corrected the exception string in `ResponseStream.finalResponse()` from `'stream ended without producing a ChatCompletion'` to `'stream ended without producing a Response'`.

### Regression Coverage
1. Added unit test in `tests/lib/ResponseInputItems.test.ts` verifying that `null`, `undefined`, numbers, strings, and objects with nullish `type` properties return `null` safely, and `toResponseInputItems` returns `[]`.
2. Added unit test in `tests/lib/ResponseStream.test.ts` verifying that `finalResponse()` throws `'stream ended without producing a Response'` when invoked on an empty stream.

### Validation
- `pnpm test` (120 passed, 0 failed, 1464 tests passed)
- `pnpm build` (Passed)
- `pnpm lint` (Passed formatting, ESLint, TypeScript types, and publint)
- `git diff --check` (Passed cleanly)

### Generated Code Impact
None. Changes are strictly confined to manually maintained files in `src/lib/responses/` and `tests/lib/`.

### DIFF
--- a/src/lib/responses/ResponseInputItems.ts
+++ b/src/lib/responses/ResponseInputItems.ts
@@ -37,6 +37,10 @@ export function toResponseInputItems(items: Iterable<ResponseInputItemLike>): Re
  * @throws {TypeError} If the item type is not supported by the installed SDK.
  */
 export function toResponseInputItem(item: ResponseInputItemLike): ResponseInputItem | null {
+  if (!item || typeof item !== 'object') {
+    return null;
+  }
+
   switch (item.type) {
     case 'additional_tools': {
       if (item.role !== 'developer') {
@@ -102,10 +106,13 @@ export function toResponseInputItem(item: ResponseInputItemLike): ResponseInputI
     case 'shell_call':
     case 'tool_search_call':
     case 'tool_search_output':
-    case 'web_search_call':
+    case 'web_search_call': {
+      return stripCreatedBy(item) as ResponseInputItem;
+    }
+
     case null:
     case undefined: {
-      return stripCreatedBy(item) as ResponseInputItem;
+      return null;
     }
 
     default: {
@@ -115,7 +122,7 @@ export function toResponseInputItem(item: ResponseInputItemLike): ResponseInputI
 }
 
 function stripCreatedBy<T extends object>(item: T): T {
-  if (!('created_by' in item)) {
+  if (!item || typeof item !== 'object' || !('created_by' in item)) {
     return item;
   }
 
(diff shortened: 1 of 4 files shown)


jbeckwith-oai (changes requested):
`src/lib/responses/ResponseInputItems.ts:113` — Please preserve the `undefined` discriminator path for valid typeless messages. `ResponseInputItem` includes `EasyInputMessage`, where `type?: "message"` is intentionally optional, so a valid input such as `{ role: "user", content: "hello" }` reaches `case undefined`. Returning `null` here silently drops that message from both normalization helpers; the previous implementation preserved it. Please only discard invalid nullish values, keep type-omitted `EasyInputMessage` objects, and add a regression test for that public input shape.

HAYDEN-OAI:
Thanks for the contribution. We checked this against the actual Responses API schema, backend input validator, current SDK, and existing review feedback. Messages intentionally support an omitted type; the backend defaults typeless user, developer, system, and assistant messages to message, and shorthand item references with only an id are also supported. Returning null for undefined discriminators would silently discard valid conversation history and item references. The null, undefined, and primitive values being guarded are outside the documented helper input and actual API output contracts. Separately, the finalResponse wording has already been corrected on main, and an empty stream fails earlier with request ended without sending any events, so the proposed regression cannot reach that wording branch. Since the only supported fix is already present and the remaining change breaks valid API inputs, we are closing this. If there is a real supported-input reproduction beyond malformed caller data, please share it and we can investigate.

PR #2018, opened 2026-07-26
fix(responses): safely initialize reasoning summary array during event accumulation

hsusul:

### Problem
When processing streaming events in `accumulateResponse`, accumulating reasoning summary events (`response.reasoning_summary_part.added`, `response.reasoning_summary_part.done`, `response.reasoning_summary_text.delta`, and `response.reasoning_summary_text.done`) on reasoning output items where `summary` was omitted or uninitialized (e.g. when an output item of type `reasoning` is added without a pre-initialized `summary` array) causes an uncaught JavaScript runtime error: `TypeError: Cannot read properties of undefined (reading 'push')` or `TypeError: Cannot read properties of undefined (reading '0')`.

### Root Cause
1. `ResponseAccumulator.ts` attempted to call `output.summary.push(...)` in `response.reasoning_summary_part.added` without checking if `output.summary` was defined, unlike `response.content_part.added` which explicitly initializes `output.content = []` when nullish.
2. `response.reasoning_summary_part.done`, `response.reasoning_summary_text.delta`, and `response.reasoning_summary_text.done` passed `output.summary` directly to `getContent(output.summary, event.summary_index)`. When `output.summary` was undefined, `getContent` accessed `content[contentIndex]`, throwing an unhandled `TypeError`.

### Solution
1. Added nullish array initialization (`if (!output.summary) output.summary = [];`) to `response.reasoning_summary_part.added`, `response.reasoning_summary_part.done`, `response.reasoning_summary_text.delta`, and `response.reasoning_summary_text.done` in `src/lib/responses/ResponseAccumulator.ts`.
2. Updated `getContent` to check for nullish `content` arrays and throw a clear `OpenAIError` instead of throwing a raw JavaScript `TypeError`.

### Regression Coverage
Added a unit test in `tests/lib/ResponseAccumulator.test.ts` (`'handles reasoning summary events on output items with uninitialized summary'`) verifying that reasoning output items with uninitialized `summary` arrays safely accumulate `summary_text` parts and deltas.

### Validation
- `npx jest tests/lib/ResponseAccumulator.test.ts` (Passed 8/8)
- `npx jest tests/lib` (Passed 16/16 test suites, 400/400 tests)
- `pnpm build` (Passed with zero errors)
- `pnpm lint` (Passed formatting, ESLint, and TypeScript type checks)
- `git diff --check` (Passed cleanly)

### Generated Code Impact
None. Changes are strictly confined to manually maintained files under `src/lib/responses/` and `tests/lib/`.

### DIFF
--- a/src/lib/responses/ResponseAccumulator.ts
+++ b/src/lib/responses/ResponseAccumulator.ts
@@ -33,7 +33,7 @@ export function accumulateResponse(
 
   switch (event.type) {
     case 'response.output_item.added': {
-      snapshot.output.push(structuredClone(event.item));
+      snapshot.output.push(normalizeOutputItem(structuredClone(event.item)));
       if (event.item.type === 'message') {
         addOutputText(snapshot);
       }
@@ -41,7 +41,7 @@ export function accumulateResponse(
     }
     case 'response.output_item.done': {
       getOutput(snapshot, event.output_index);
-      snapshot.output[event.output_index] = structuredClone(event.item);
+      snapshot.output[event.output_index] = normalizeOutputItem(structuredClone(event.item));
       if (event.item.type === 'message') {
         addOutputText(snapshot);
       }
@@ -394,10 +394,25 @@ export function accumulateResponse(
 
 function cloneResponse(response: Response): Response {
   const snapshot = structuredClone(response);
+  for (const output of snapshot.output) {
+    normalizeOutputItem(output);
+  }
   if (!Object.getOwnPropertyDescriptor(snapshot, 'output_text') || snapshot.output_text == null) {
     addOutputText(snapshot);
   }
   return snapshot;
+}
+
+/**
+ * Reasoning items are required to carry a `summary` array, but the wire can omit it.
+ * Normalizing items as they enter a snapshot keeps that invariant for every snapshot
+ * we return, so consumers and the summary event handlers can rely on the array.
+ */
+function normalizeOutputItem(output: Response['output'][number]): Response['output'][number] {
+  if (output.type === 'reasoning' && !output.summary) {
+    output.summary = [];
+  }
+  return output;
 }
 
 function getOutput(snapshot: Response, outputIndex: number): Response['output'][number] {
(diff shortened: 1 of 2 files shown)


jbeckwith-oai (changes requested):
Requesting a structural correction before merge:

**`src/lib/responses/ResponseAccumulator.ts:35-44` / `:187-225` — establish the required `summary` invariant at ingestion instead of patching four consumers.** The regression fixture demonstrates that the wire can provide a reasoning item without its required `summary`, but this patch still pushes that malformed item into the snapshot unchanged. If no summary event follows, `accumulateResponse()` returns a value typed as `Response` with `summary === undefined`; lifecycle and `output_item.done` replacements can preserve the same mismatch. Meanwhile, the new initialization is repeated in four handlers, and three of them assign `[]` only to call `getContent(..., summary_index)` and immediately throw, leaving partially mutated state.

Please normalize cloned reasoning items once at the `output_item.added`/`done` and response-cloning boundaries (or via one canonical helper), so every returned snapshot satisfies the `ResponseReasoningItem.summary` contract. Then the summary handlers can stay direct: `part.added` appends to a known array, while `getContent` validates event ordering for delta/done cases. Add a no-summary-events regression proving the returned snapshot exposes `summary: []`; use a narrow wire-boundary assertion rather than broad `as any` in the fixture.

Validation at exact head `9546deb0d1c7532e927529709a9d2e7079f56bbf`: focused ResponseAccumulator suite passed (8 tests); changed-file ESLint and Prettier checks passed; `git diff --check` passed. No hosted checks are attached to the PR.

HAYDEN-OAI:
Thanks for the contribution and for addressing the earlier review. We traced this against the actual Responses reasoning-item renderer, first-party emitted-event integration tests, generated SDK/OpenAPI types, and Azure's documented contract. Every reasoning item, including response.output_item.added and response.output_item.done, includes the required summary array; the backend explicitly renders summary as an empty array until summary content is available. Only reasoning content is optional, which is why its accumulator has different initialization behavior. The fixtures in this PR omit a required field and do not represent a supported OpenAI or Azure event, and we found no Bedrock/provider reproduction. Normalizing malformed completed or lifecycle snapshots would also hide producer failures and add work on every response. We are therefore closing this rather than adding unsupported wire behavior. If you have a captured event from a supported provider that actually omits summary, please share it so we can investigate the producer/protocol.

PR #2019, opened 2026-07-26
test: support hermetic local API reference

apcha-oai:
Allow the SDK mock server to use a locally available API reference when present, while preserving the existing network-based fallback.

### DIFF
--- a/.prettierignore
+++ b/.prettierignore
@@ -4,6 +4,7 @@ pnpm-lock.yaml
 /ecosystem-tests/*/**
 /node_modules
 /deno
+/api_reference/openapi.transformed.yml
 
 # don't format tsc output, will break source maps
 /dist
--- a/scripts/mock
+++ b/scripts/mock
@@ -7,6 +7,15 @@ cd "$(dirname "$0")/.."
 if [[ -n "$1" && "$1" != '--'* ]]; then
   URL="$1"
   shift
+elif [[ -f .castiron.stats.yml && -f api_reference/openapi.transformed.yml ]]; then
+  URL="api_reference/openapi.transformed.yml"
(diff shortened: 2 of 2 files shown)


jbeckwith-oai (approved):
Thermo-nuclear code-quality review: no high-confidence maintainability findings. The change is a small, cohesive extension of the existing spec-resolution flow; `scripts/mock` remains 61 lines and no file approaches the 1,000-line threshold. Approved.

.prettierignore
```diff
@@ -4,6 +4,7 @@ pnpm-lock.yaml
 /ecosystem-tests/*/**
 /node_modules
 /deno
+/api_reference/openapi.transformed.yml
```
HAYDEN-OAI, line 7: [P2] Preserve the transformed spec across Git checkout

Ignoring Prettier protects the file after checkout, but not during checkout. The companion Castiron flow emits `api_reference/.gitattributes` with only `openapi.transformed.yml -whitespace`; that leaves Git text conversion unspecified. In a checkout with `core.autocrlf=true`, the committed LF bytes become CRLF while `.castiron.stats.yml` retains the LF MD5, so `scripts/mock` exits with 'does not match generation metadata' on an otherwise valid generated SDK. Please mark this path `-text` (here or in the producer's generated attributes) so the exact-byte contract survives checkout.

HAYDEN-OAI (approved):
Reviewed the exact head against the Castiron hermetic-reference contract. Local, explicit-override, legacy-fallback, and MD5-rejection paths behave as intended, and npm/JSR build outputs exclude the repository-level reference. One low-priority cross-platform checkout edge is noted inline; approving because the core change is sound.

PR #2021, opened 2026-07-26
fix(responses): reject with OpenAIError when the stream emits an error event

### LINKED ISSUE #2020: ResponseStream rejects with a plain object when the API emits an `error` event
Streaming a Responses request that surfaces an API `error` frame rejects
`finalResponse()` with a plain object instead of an `OpenAIError`, so
`instanceof`-based error handling misses it.

Repro against the shipped class — no network, replaying a captured frame
through `ResponseStream.fromReadableStream`:

    finalResponse REJECTED: isError=false isOpenAIError=false
    raw= {"type":"error","code":"server_error","message":"boom","sequence_number":1}

Cause: `error` is a legitimate member of `ResponseStreamEvent`, so
`#addEvent`'s `default:` arm re-emits it as an ordinary listener payload, and
`EventStream._emit` then rejects with the raw frame. The Assistants helper
handles the same class explicitly (AssistantStream.ts:381, "errors should be
processed earlier"), and the realtime emitter converts properly — Responses is
the only stream helper that doesn't.

Sending a PR alongside this: convert the frame in ResponseStream's error path,
mirroring the Assistants handling. The alternative is a transport-layer fix in
core/streaming.ts (which carries an unresolved TODO asking exactly this
question) — happy to move it there instead if you'd rather weigh the
Chat-Completions blast radius.

morgan-coded:

## Summary

Fixes #2020.

- Convert streamed Responses API failures into `APIError` before response accumulation or generic event emission, including errors received before `response.created`.
- Preserve the error message, code, parameter, classification, and embedded retry metadata for both actual nested OpenAI/Azure error envelopes and the flat shape described by the current generated schema.
- Reuse `EventStream.events('event')` so asynchronous iteration retains failures that occur while an application is processing an earlier event, drains buffered events in order, and cleans up correctly on cancellation.
- Keep lifecycle `error` listeners typed as `OpenAIError` and preserve normal `response.failed` snapshot handling.

## Verification

- Added coverage for nested and flat errors, initial errors, serialized replay, actual SSE transport, request IDs, delayed iteration, buffered events, iteration after completion, and early cancellation.
- Full handwritten suite: **1,701 tests passing**.
- Repository lint and Node.js version-policy validation passed.
- Source and published-package type checks passed with TypeScript 6 and TypeScript 4.9.
- Complete CommonJS/ESM package build and Bedrock import smoke tests passed.


jbeckwith-oai (dismissed):
The conversion is still incomplete in two valid consumption paths:

1. `#addEvent` calls `accumulateResponse()` before checking `event.type`, so an `error` frame sent before `response.created` never reaches the new branch. It still rejects with the accumulator's generic `OpenAIError` and loses the server message/code/param. Please convert error frames before snapshot accumulation and add coverage for an initial error frame.

2. The raw error frame is emitted through the generic `event` channel before the `APIError` is thrown, while `ResponseStream[Symbol.asyncIterator]` only rejects reads that are pending at the instant the error fires and does not retain the failure. A normal `for await` consumer therefore observes `[response.created, error]` and then completes cleanly; only a separate `finalResponse()` call sees the `APIError`. Please make async iteration reject consistently (the failure-retention pattern in `EventStream.events()` is a useful model) and add a regression test.

Validation at `c34bfea5d94941e37038bb6ac8834240f84590c4`: the focused ResponseStream suite passed (6/6), Prettier and ESLint passed on changed files, TypeScript checking passed, the package build passed, and `git diff --check` passed. I also reproduced both cases against the built package.

morgan-coded:
Reproduced both stream-error paths against the built package and fixed them in the same order.

1. Moved the `error` conversion before snapshot accumulation and generic `event` emission, so an initial error frame now rejects with an `APIError` carrying the server's message, code, and param. The regression failed at `c34bfea5` with `Expected constructor: APIError` / `Received constructor: OpenAIError`, then passed on the revision.
2. Retained async-iterator failures using `EventStream.events()`'s queued-event-first, one-time rejection pattern, so `for await` rejects even when no read was pending when the error fired. The regression failed at `c34bfea5` with `Received promise resolved instead of rejected` / `Resolved to value: undefined`, then passed on the revision.

The focused ResponseStream file passes 8/8; `./scripts/lint` and `./scripts/build` pass; the full `./scripts/test` passes 120 suites, 1465 tests, and 64 snapshots, with one suite and one test skipped.

HAYDEN-OAI (approved):
Validated against the actual first-party/Azure nested SSE error contract and the documented flat compatibility shape. Errors are normalized before accumulation, API metadata and request IDs are preserved, and async iteration reuses EventStream.events to retain delayed failures. Verified 1,701 unit tests, focused live-SSE/replay/cancellation cases, lint, TypeScript 4.9/6 published checks, and full CJS/ESM build.

PR #2022, opened 2026-07-27
fix(responses): correct error message in finalResponse and safely accumulate string deltas

hsusul:

## Changes being requested

- Fix error message thrown by `ResponseStream.finalResponse()` when a stream ends prematurely without producing a response (`stream ended without producing a Response` instead of `stream ended without producing a ChatCompletion`).
- Fix typo in `ResponseStream.finalResponse()` docstring (`REsponse` -> `Response`).
- Safely handle nullish initial values when accumulating string deltas in `ResponseAccumulator.ts` across `output_text`, `refusal`, `function_call_arguments`, `reasoning_text`, `reasoning_summary_text`, `custom_tool_call_input`, and `mcp_call_arguments` events using `(target ?? '') + event.delta`.
- Add unit tests for `finalResponse()` error message and `ResponseAccumulator` string delta accumulation.

### Problem
1. `ResponseStream` belongs to the Responses API. When `finalResponse()` is called on a stream that ended prematurely without producing a `Response`, it previously threw `OpenAIError('stream ended without producing a ChatCompletion')`.
2. When stream delta events (`response.function_call_arguments.delta`, `response.refusal.delta`, etc.) arrive on uninitialized string properties, `+= event.delta` resulted in `"undefined"` prefixing the delta content.

### Root cause
1. Copy-paste typo from `ChatCompletionStream.ts` in `ResponseStream.ts`.
2. Missing nullish fallback `(str ?? '')` on string property delta accumulation in `ResponseAccumulator.ts`.

### Verification
- `pnpm test tests/lib/ResponseStream.test.ts` (Pass)
- `pnpm build` (Pass)
- `pnpm lint` (Pass)

### DIFF
--- a/src/lib/responses/ResponseAccumulator.ts
+++ b/src/lib/responses/ResponseAccumulator.ts
@@ -90,8 +90,8 @@ export function accumulateResponse(
         if (content.type !== 'output_text') {
           throw new OpenAIError(`expected content to be 'output_text', got ${content.type}`);
         }
-        content.text += event.delta;
-        snapshot.output_text += event.delta;
+        content.text = (content.text ?? '') + event.delta;
+        snapshot.output_text = (snapshot.output_text ?? '') + event.delta;
       }
       break;
     }
@@ -127,7 +127,7 @@ export function accumulateResponse(
         if (content.type !== 'refusal') {
           throw new OpenAIError(`expected content to be 'refusal', got ${content.type}`);
         }
-        content.refusal += event.delta;
+        content.refusal = (content.refusal ?? '') + event.delta;
       }
       break;
     }
@@ -145,7 +145,7 @@ export function accumulateResponse(
     case 'response.function_call_arguments.delta': {
       const output = getOutput(snapshot, event.output_index);
       if (output.type === 'function_call') {
-        output.arguments += event.delta;
+        output.arguments = (output.arguments ?? '') + event.delta;
       }
       break;
     }
@@ -166,7 +166,7 @@ export function accumulateResponse(
         if (content.type !== 'reasoning_text') {
           throw new OpenAIError(`expected content to be 'reasoning_text', got ${content.type}`);
         }
-        content.text += event.delta;
+        content.text = (content.text ?? '') + event.delta;
       }
       break;
     }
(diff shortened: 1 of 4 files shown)


masomelelemgqamqo-rgb (approved):

jbeckwith-oai (changes requested):
Requesting one test correction before merge:

**`tests/lib/ResponseStream.test.ts:276` — the regression test never exercises the nullish case.** The fixture initializes `arguments` to `''`, so the pre-change implementation (`output.arguments += event.delta`) already produces `'{"loc'` and passes this assertion. The test therefore cannot catch a regression in the new `?? ''` behavior. Please model the reported wire payload by actually omitting/setting the field to `undefined` at a narrow cast boundary. Since this PR changes eight accumulator targets, table-driven coverage in the existing focused `ResponseAccumulator.test.ts` for the affected event families would make the contract explicit without duplicating full stream setup.

The runtime changes themselves are small, direct, and otherwise look safe.

Validation at exact head `5daeffe0febbdcb0d59672e8ee3eae3eab2fc79a`: focused ResponseStream and ResponseAccumulator tests passed (14 total); changed-file ESLint and Prettier checks passed; `git diff --check` passed. Full repository type-check was blocked by missing optional example-only dependencies (`@azure/identity`, `express`, and `next`) in the available dependency tree. No hosted checks are attached to the PR.

HAYDEN-OAI:
Thanks for the contribution and for following up on the earlier review. The finalResponse() wording and documentation have already been fixed on main. We also traced every proposed nullish accumulator case against the actual Responses service, canonical schemas, emitted-event integration tests, and supported-provider contracts: output/refusal/reasoning text, function and MCP arguments, custom tool input, and reasoning-summary text are required and initialized before delta events. The one SDK-only field, output_text, is also reconstructed by cloneResponse()/addOutputText() before accumulation. The added fixtures delete required fields from otherwise valid snapshots and therefore do not reproduce a supported OpenAI, Azure, or Bedrock stream; this duplicates the unsupported premise already addressed in #2016. We are closing this as superseded and out of scope. If you have a captured event from a supported provider that actually omits one of these initialized fields, please share it and we can investigate the producer/protocol.

PR #2026, opened 2026-07-27
feat!: require Node.js 22 and codify version support

jbeckwith-oai:
## Summary

- require Node.js 22 or later, with Node.js 22 and 24 as supported LTS lines
- forward-test Node.js 26 without making the Current line a production-support promise
- add a concise lifecycle policy, type-checked drift and packed-package checks, and a monthly Codex review that opens draft proposals
- document the purpose of the affected GitHub Actions workflows

## User impact

Node.js 20 reached upstream end of life on April 30, 2026 and is no longer supported. Previously published SDK releases remain available, but receive no guaranteed fixes or security backports for unsupported Node.js versions. Repository tooling continues to use Node.js 24 independently of the consumer runtime floor.

## Validation

- Node.js policy checker
- TypeScript typecheck
- actionlint
- Prettier and diff hygiene
- full lint/build/package validation
- full test suite on Node.js 22 and 24: 1,462 passed, 1 skipped
- packed npm artifact CommonJS, ESM, and engine-metadata checks on Node.js 22 and 24
- forward-compatibility suite on Node.js 26

## Release note

**Breaking change:** The SDK now requires Node.js 22 or later. The exact final Node.js 20-compatible SDK version must be named in the release notes before this pull request is marked ready.


.github/workflows/ci.yml
```diff
@@ -10,6 +13,7 @@ on:
       - '!codegen/**'
       - 'codegen/stl/**'
   pull_request:
+    types: [opened, synchronize, reopened, ready_for_review]
```
HAYDEN-OAI, line 16: P2: The generated draft’s ready-for-review path only reaches this workflow. `.github/workflows/detect-breaking-changes.yml` still uses the default `pull_request` activity types, so the documented “mark ready to start checks” flow skips both breaking-change detection and the Agents SDK regression job. Add `ready_for_review` there as well, or trigger those checks through another authenticated path.

.github/workflows/node-version-review.yml
```diff
@@ -0,0 +1,84 @@
...
+      - name: Validate the repository
+        run: |
+          node --experimental-strip-types scripts/check-node-version-policy.ts
+          pnpm lint
+          pnpm test
+
+      - name: Open or update the draft pull request
+        uses: peter-evans/create-pull-request@22a9089034f40e5a961c8808d113e2c98fb63676 # v7
+        with:
+          token: ${{ github.token }}
```
HAYDEN-OAI, line 66: P2: The prompt requires updates to `.github/workflows/ci.yml`, and that file is included in `add-paths`, but this step publishes with `github.token`. The built-in workflow token has `contents: write` here but cannot be granted the separate workflow-file write permission needed to push a changed file under `.github/workflows`, so the monthly job cannot open its draft PR for the main matrix-drift case. Use a token with workflow-write capability or keep workflow edits out of this automated push.

HAYDEN-OAI (approved):
Reviewed at `b35ec6c6f25f`. The Node.js 22 floor, published engine metadata path, compatibility docs, and policy checker look sound. Approving with two low-priority automation follow-ups inline.

.github/workflows/node-version-review.yml
jbeckwith-oai: Fixed in 09160300 without adding another credential. CI now derives its matrix from the canonical policy table, and `.nvmrc` owns the repository toolchain across workflows. The monthly Codex publisher no longer edits or includes workflow files, so `github.token` only pushes ordinary repository files. The policy checker, emitted matrix, targeted typecheck, and `actionlint` all pass.

.github/workflows/ci.yml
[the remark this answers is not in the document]
jbeckwith-oai: Fixed in 09160300. `detect-breaking-changes.yml` now explicitly includes `ready_for_review`, so both breaking-change detection and the Agents SDK regression job run when a generated draft is marked ready. `actionlint` passes with the new trigger.

.github/workflows/node-version-review.yml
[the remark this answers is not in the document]
jbeckwith-oai: Fixed in 09160300 without adding another credential. CI now derives its matrix from the canonical policy table, and `.nvmrc` owns the repository toolchain across workflows. The monthly Codex publisher no longer edits or includes workflow files, so `github.token` only pushes ordinary repository files. The policy checker, emitted matrix, targeted typecheck, and `actionlint` all pass.

PR #2040, opened 2026-07-31
[SDK-199] Migrate handwritten Node SDK tests from Jest to Vitest

HAYDEN-OAI:
## improves test perf by ~ 50%

The equivalent handwritten suite drops from **13.58 seconds with Jest to 6.60 seconds with Vitest**: approximately **51% less wall-clock time**, or **2.1× faster**. All Stainless-generated tests remain on Jest.

## Changes being requested

- Migrate handwritten SDK tests, mocks, timers, module isolation, and snapshot compatibility from Jest to Vitest.
- Keep all Stainless-generated tests, live tests, ecosystem fixtures, and existing snapshots unchanged on their existing Jest runners.
- Route handwritten and generated test filters to the correct runner and preserve Steady mock-server lifecycle.
- Run handwritten Vitest files in parallel with four isolated workers; preserve process isolation for mocks, globals, environment variables, and fake timers.
- Keep `./scripts/test` as the canonical complete-suite entrypoint while retaining `pnpm test` as an alias.
- Update CI, contributor documentation, dependency lockfiles, runner regression coverage, and focused handwritten SDK behavior tests.

## Additional context & links

Resolves [SDK-199](https://linear.app/openai/issue/SDK-199/migrate-the-openai-node-test-suite-from-jest-to-vitest).

Validation:

- `CI=1 ./scripts/test`
  - Vitest: 61 files and 1,453 tests passed.
  - Generated Jest: 557 tests passed; 1 expected skip; all 12 snapshots passed.
  - 2,010 tests passed in total.
- `./node_modules/.bin/prettier --check .`
- `./node_modules/.bin/eslint .`
- `./node_modules/.bin/tsc --noEmit`
- `./scripts/build`
- `pnpm exec attw --pack .`
- `pnpm exec publint`

Handwritten-only runner benchmark (same 59 files / 1,415 tests, Node 26.5.0, three warm runs):

| Runner | Median |
| --- | ---: |
| Jest, single process | 13.58 seconds |
| Vitest, four isolated workers | 6.60 seconds |

Vitest runs the equivalent handwritten suite approximately 2.1 times faster while leaving Stainless-generated tests on Jest.


scripts/vitest-coverage-provider.mjs
```diff
@@ -0,0 +1,38 @@
+import { readFileSync } from 'node:fs';
+import { fileURLToPath } from 'node:url';
+import v8Coverage from '@vitest/coverage-v8';
+import { V8CoverageProvider } from '@vitest/coverage-v8/dist/provider.js';
```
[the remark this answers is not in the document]
HAYDEN-OAI: This is a false positive for the exact dependency pinned on this head. `@vitest/coverage-v8@4.1.10` exports `"./*": "./*"`, so `@vitest/coverage-v8/dist/provider.js` is an exported subpath. The fresh exact-head coverage jobs also exercised this import and passed (for example run 30652880464 / job 91230156400: 82 suites, 559 tests, coverage thresholds met). No code change is needed.

scripts/merge-coverage.cjs
```diff
@@ -0,0 +1,37 @@
...
+const { createContext } = require('istanbul-lib-report');
+const reports = require('istanbul-reports');
+
+const coverageDirectory = join(__dirname, '..', 'coverage');
+const coverageMap = createCoverageMap({});
+
+for (const suite of ['unit', 'generated']) {
+  const coveragePath = join(coverageDirectory, suite, 'coverage-final.json');
+  const coverage = JSON.parse(readFileSync(coveragePath, 'utf8'));
+  coverageMap.merge(coverage);
```
jbeckwith-oai: The two inputs still do not share branch/function identities, so this merges incompatible coverage ranges. On this exact head, unit coverage has 1,071 functions / 4,567 branches and generated coverage has 835 / 1,258, but the merged map becomes 1,209 / 5,252; the pre-PR all-Jest run over the identical `src/` tree is 1,140 / 5,032. The `fnMap`/`branchMap` locations differ across the Vite and SWC transforms (for example, `src/client.ts` produces 291 vs. 245 branch entries), so the thresholds no longer measure one coherent source map. Please make both runners emit the same canonical function/branch map (or collect through one instrumenter), and add a regression that checks map-shape compatibility before merging.

scripts/test
```diff
@@ -40,6 +40,74 @@ function has_jest_worker_flag() {
...
+      ;;
+    tests/*|./tests/*|*/tests/*)
+      has_unit_filter=true
+      vitest_args+=("$arg")
+      ;;
+    --runInBand|--runInBand=*|--run-in-band|--run-in-band=*|-i|--showConfig)
+      jest_args+=("$arg")
+      ;;
+    --maxWorkers|--max-workers|-w)
+      jest_args+=("$arg")
```
jbeckwith-oai, line 89: `--maxWorkers` is supported by both Jest and Vitest, but this consumes it only into `jest_args`. As a result, `pnpm test -- --maxWorkers=1` still starts four Vitest workers, regressing the existing resource-control contract and risking failures in constrained jobs. Please forward the camel-case option and its value to both runners (translating aliases if needed), while keeping genuinely Jest-only flags such as `--runInBand` local to Jest, and cover that routing in `test-script.test.ts`.

jest.config.ts
```diff
@@ -8,6 +8,7 @@ const generatedTestPatterns = [
 ];
 
 const testSuite = process.env['OPENAI_TEST_SUITE'] ?? 'all';
+const partialCoverage = process.env['OPENAI_COVERAGE_PARTIAL'] === '1';
 
 if (!['all', 'generated', 'unit'].includes(testSuite)) {
```
jbeckwith-oai: The suite boundary is now duplicated across this config, `vitest.config.mts`, and the shell router, while Jest still defaults to `all` and accepts `unit`. Both of those modes now load handwritten files that import `vitest`, so direct `jest` / `OPENAI_TEST_SUITE=unit jest` fails before running tests. Please make Jest generated-only and move the generated patterns to one canonical source consumed by both configs and the router; otherwise a future generated top-level file can silently land in the wrong runner.

jbeckwith-oai (changes requested):
Requesting changes for combined-coverage correctness, worker-limit preservation, and a single generated/handwritten suite boundary. The runner migration itself is well tested, but these are blocking harness regressions.

scripts/merge-coverage.cjs
HAYDEN-OAI: Fixed in `d87c27925a8905e7ecfb819c5c728f227fe81e10`. Coverage now executes every handwritten and generated suite through one Jest/V8 instrumenter, so there is one coherent function/branch identity rather than two incompatible maps. The full coverage run passed 2,017 tests (one expected skip) and the unchanged thresholds: 98.94% statements/lines, 90.65% branches, and 94.47% functions. Added regressions also reject count-equal but location-different `fnMap`/`branchMap` inputs before any future merge.

scripts/test
HAYDEN-OAI: Fixed in `d87c27925a8905e7ecfb819c5c728f227fe81e10`. All supported `--maxWorkers`/`--max-workers`/`-w` value spellings now route to both runners with Vitest aliases normalized, while `--runInBand` and other Jest-only flags remain local to Jest. The real `pnpm test -- --maxWorkers=1` run passed 1,460 Vitest tests and 557 Jest tests; focused routing/config regressions passed 34 tests.

jest.config.ts
HAYDEN-OAI: Fixed in `d87c27925a8905e7ecfb819c5c728f227fe81e10`. Generated-suite patterns now live in one canonical JSON manifest consumed by ordinary Jest, Vitest exclusions, and the shell router. Ordinary/default Jest is generated-only; the coverage-only config is the sole all-suite Jest mode. Regressions verify direct suite boundaries, relative/absolute filters, and that every current Stainless-generated top-level test routes only to Jest. Full lint/build/types/ATTW/publint also passed.

scripts/jest-coverage-transformer.cjs
```diff
@@ -0,0 +1,23 @@
+const { createTransformer } = require('@swc/jest');
+
+const transformer = createTransformer({ sourceMaps: 'inline' });
+
+function normalizeVitestMocks(source, filename) {
+  if (!filename.includes('/tests/') || !/from\s*['"]vitest['"]/.test(source)) {
```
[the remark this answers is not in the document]
HAYDEN-OAI: Fixed in `5daa613ef6ee13134819f287871a7a3ac703b270`. The transformer now normalizes `\\` to `/` before applying the `/tests/` guard, retaining the non-test early return. The regression exercises both a POSIX path and `C:\\workspace\\openai-node\\tests\\...` and verifies `vi.mock` is rewritten/hoisted ahead of the static import. Focused coverage/config tests passed 35 tests; the exact coverage path passed 144 suites, 2,018 tests, 58 snapshots, and the unchanged thresholds, with full lint/build/types/ATTW/publint clean.

scripts/merge-coverage.cjs
```diff
@@ -0,0 +1,80 @@
...
+          );
+        }
+      }
+    }
+  }
+}
+
+function mergeCoverage() {
+  const coverageDirectory = join(__dirname, '..', 'coverage');
+  const suiteCoverage = ['unit', 'generated'].map((suite) => {
```
jbeckwith-oai: The new `test:coverage` correctly uses one Jest/V8 pass and never produces `coverage/unit` or `coverage/generated`, so this entire merge path is now unreachable. The only caller of its export is its own test; invoking `node scripts/merge-coverage.cjs` after the real coverage command fails with `ENOENT` for `coverage/unit/coverage-final.json`. Please take the simpler design all the way: delete this script, `tests/merge-coverage.test.ts`, and the now-private `istanbul-lib-coverage` / `istanbul-lib-report` / `istanbul-reports` dependencies. Keeping ~150 lines of tested but unusable parallel coverage infrastructure is exactly the kind of incidental complexity this refactor should remove.

jbeckwith-oai (dismissed):
The three prior blockers are resolved. One new maintainability blocker remains: the new single-instrumenter design makes the old merge subsystem dead and broken, so it should be deleted rather than retained as parallel infrastructure.

scripts/merge-coverage.cjs
HAYDEN-OAI: Fixed in `5890eee25907af30896f2884c4584ee30ea3eb68`. Removed `scripts/merge-coverage.cjs`, its self-only test, and the three direct Istanbul declarations/importer entries; the packages remain only where transitively required. I also removed the stale coverage-merging wording from the contributor guide. The exact `pnpm test:coverage` path passed 143 suites / 2,013 tests at the unchanged thresholds (98.94% statements/lines, 90.65% branches, 94.47% functions), and frozen-lockfile install, formatting, ESLint, build, TypeScript, ATTW, publint, and the Node policy check all passed.

.github/workflows/ci.yml
```diff
@@ -144,7 +144,7 @@ jobs:
           node-version: '${{ matrix.node-version }}'
 
       - name: Run tests
-        run: ./scripts/test
+        run: pnpm test
```
apcha-oai: `./scripts/test` is the canonical test contract across our SDKs, so CI should exercise it directly. `pnpm test` can remain a convenient alias to that script, but making the workflow call the package-manager alias reverses that ownership and makes it possible for package metadata to drift from the cross-SDK CI contract. I think this should remain `./scripts/test`.

jest.coverage.config.ts
```diff
@@ -0,0 +1,17 @@
+import type { JestConfigWithTsJest } from 'ts-jest';
+import config from './jest.config';
+
+const coverageConfig: JestConfigWithTsJest = {
+  ...config,
+  testMatch: ['<rootDir>/tests/**/*.test.ts'],
+  transform: {
+    '^.+\\.(t|j)sx?$': '<rootDir>/scripts/jest-coverage-transformer.cjs',
```
apcha-oai: I think the aggregate coverage requirement may be driving substantially more complexity than the runner migration itself. This path runs Vitest-authored tests under Jest by rewriting `vi.mock(...)` and providing a partial Jest-backed `vi` shim, so the handwritten suite has to work under both its real runner and a compatibility runtime.

Could we simplify by reconsidering whether one combined coverage report is necessary here? A few options seem materially simpler:

- Drop coverage enforcement for now. It was only added recently, so this would not give up a longstanding guarantee.
- Keep separate runner-native coverage for handwritten Vitest tests and generated Jest tests, with separate reports or thresholds.
- Keep coverage for only one suite initially and revisit the combined story separately.

Any of these would let each suite run only under its actual runner and avoid the coverage-only transformer and shim. My preference would be whichever of these gives us the simplest migration while preserving the signal we care about.

apcha-oai (commented):
Two suggestions to simplify the test-runner boundary and keep the cross-SDK test contract explicit.

.github/workflows/ci.yml
HAYDEN-OAI: Fixed in `07ac5f7a582509e13e5248c0cac1f2b48b2c1675`. The Node matrix now invokes the repository contract directly with `./scripts/test`; `pnpm test` remains a package-manager alias. The workflow dependency/aggregate structure was validated, and the canonical full run passed 2,010 tests (one expected skip) with all 12 snapshots.

jest.coverage.config.ts
HAYDEN-OAI: Addressed in `07ac5f7a582509e13e5248c0cac1f2b48b2c1675` with the simplest option you proposed: the recent aggregate coverage gate is removed so each suite runs only under its native runner. This deletes the coverage job/script, all-suite Jest config, rewrite transformer, partial `vi` shim, compatibility-only tests, direct coverage provider dependency/lock entries, and stale coverage docs; the ordinary generated Jest and handwritten Vitest boundaries remain canonical. Validation passed 2,010 tests (one expected skip), 38 focused Node 22 routing/behavior checks, frozen install, format, ESLint, build, TypeScript, ATTW, publint, packed CJS/ESM verification, workflow validation, and policy checks.

apcha-oai (approved):
nit: think description is out of date re: coverage otherwise lgtm!

PR #2043, opened 2026-07-31
Reject recordAudio when ffmpeg exits unsuccessfully

fallintoplace:
`recordAudio()` currently resolves with whatever bytes ffmpeg wrote even when the process exits with a non-zero status. This can turn device, permission, or encoder failures into apparently successful audio files.

Reject explicit non-zero exit codes while preserving the existing behavior for signal-terminated recordings.

Tests cover a failed exit after partial output and the signal-termination path.

Checks:
- `pnpm exec jest --runInBand tests/helpers/audio-recording.test.ts`
- `pnpm test:unit`
- `pnpm exec prettier --check .`
- `pnpm exec eslint .`
- `pnpm build`

### DIFF
--- a/src/helpers/audio.ts
+++ b/src/helpers/audio.ts
@@ -90,6 +90,7 @@ function nodejsRecordAudio({ signal, device, timeout }: RecordAudioOptions = {})
   checkFileSupport();
   return new Promise((resolve, reject) => {
     const data: any[] = [];
+    let terminationRequested = false;
     const provider = recordingProviders[platform];
     try {
       const ffmpeg = spawn(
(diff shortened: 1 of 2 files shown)


HAYDEN-OAI:
Thanks for identifying this. The recordAudio failure-handling behavior is already implemented on main in #2281: unexpected FFmpeg exits are rejected, intentionally stopped recordings retain their captured audio, and failed termination attempts are handled correctly. Existing tests also cover external aborts, timeouts, pre-aborted signals, and unsuccessful exits. The current implementation additionally rejects unexpected signal-terminated processes, which this older patch would accidentally accept. Closing as superseded.

PR #2044, opened 2026-08-01
[SDK-198] Migrate ESLint and Prettier to Oxlint and Oxfmt

HAYDEN-OAI:

## Changes being requested

Migrate the existing ESLint/Prettier toolchain to standalone Oxlint and Oxfmt.

- Pin `oxlint@1.75.0` and `oxfmt@0.60.0`, and remove the root ESLint, TypeScript-ESLint, and Prettier dependencies.
- Preserve the existing formatting options, exclusions, restricted self-import rule, fixture exceptions, and historical breaking-change compatibility.
- Add a small dependency-free Oxlint plugin to preserve the current unused-import-only behavior without introducing broad unused-variable errors.
- Update the existing lint/format/fast-format scripts, contributor documentation, and VS Code development-container integration.
- Apply the one-time mechanical formatting migration required because Oxfmt does not support Prettier's `experimentalTernaries` option.

## Additional context & links

- Linear: https://linear.app/openai/issue/SDK-198/migrate-eslint-and-prettier-to-oxlint-and-oxfmt

### Validation

- `pnpm lint`
- `pnpm test:unit` — 61 test files and 1,453 tests passing
- 20 focused Oxlint import-rule/autofix cases passing
- `pnpm format` and selective/empty `scripts/fast-format` behavior verified

### Tooling performance

Median wall-clock time for the exact repository checks:

| Check | ESLint / Prettier | Oxlint / Oxfmt | Improvement |
| --- | ---: | ---: | ---: |
| Formatting | Prettier 3.1.1: **6.44s** | Oxfmt 0.60.0: **0.87s** | **7.4× faster** (86% less time) |
| Linting | ESLint 9.39.4: **4.22s** | Oxlint 1.75.0: **0.89s** | **4.7× faster** (79% less time) |
| Combined | **10.72s** | **1.79s** | **6.0× faster** (83% less time) |

Measured on Linux x64 with Node 26.5.0 and 32 available logical CPUs. Each result is the median of **seven interleaved runs after two warmup runs**, using isolated tracked-file snapshots of base `aeab0549` and PR head `b014a1f`; repository ignore rules were preserved and the local untracked `coverage/` directory was excluded from both snapshots. No persistent formatter or linter caches were enabled.

```sh
# Before
prettier --check .
eslint .

# After
oxfmt --check .
oxlint .
```

Combined timings were measured by running the formatter and linter sequentially; unchanged build, TypeScript, and package-validation stages were excluded.


tests/uploads.test.ts
```diff
@@ -66,7 +66,7 @@ describe('toFile', () => {
     const result = await toFile(input);
     const file: File = result;
     const blob: Blob = result;
-    void file, blob;
+    (void file, blob);
```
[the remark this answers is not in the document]
HAYDEN-OAI: Fixed in `b014a1f580d97fc2fb3745b31db2ba58c37cb348`. The assignability-only check now uses separate `void file;` and `void blob;` statements, eliminating the effectless comma expression without changing behavior. The focused upload test, Oxfmt, Oxlint, TypeScript, Node-version policy, and complete `./scripts/lint` pipeline (including build, package-type verification, and publint) passed.

scripts/fast-format
```diff
@@ -24,14 +24,13 @@ if [ ! -f "$FILE_LIST" ]; then
...
+   echo "$LINT_FILES" | xargs ./node_modules/.bin/oxlint --fix
 fi
 
-echo "==> Running prettier --write"
-if ! [ -z "$FILE_LIST" ]; then
-   cat "$FILE_LIST" | xargs ./node_modules/.bin/prettier \
-      --write --cache --cache-strategy metadata --no-error-on-unmatched-pattern --ignore-unknown
+echo "==> Running oxfmt --write"
+if [ -s "$FILE_LIST" ]; then
+   xargs ./node_modules/.bin/oxfmt --no-error-on-unmatched-pattern < "$FILE_LIST"
```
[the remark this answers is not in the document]
HAYDEN-OAI: Addressed in `56223b9ec558fd6cfea77e1a11154040448839af`. `scripts/fast-format` now passes `oxfmt --write`, making the incremental helper's mutation explicit and consistent with `scripts/format`. The pinned Oxfmt 0.60.0 did write implicitly in reproduction, but the explicit flag removes that behavioral ambiguity. An end-to-end JSON formatting regression, `bash -n`, and the complete `./scripts/lint` pipeline passed.

scripts/oxlint-plugin.cjs
```diff
@@ -0,0 +1,130 @@
...
+          );
+          if (!specifier) {
+            continue;
+          }
+
+          context.report({
+            node: specifier.local,
+            messageId: 'unused',
+            data: { name: variable.name },
+            fix: (fixer) => fixUnusedImport(sourceCode, declaration, specifier, fixer),
```
jbeckwith-oai: **P1 — make this a single declaration-level fix.** Each unused binding currently reports its own overlapping edit. With `import d, { a } from './x'`, `oxlint --fix` removes `d` but leaves `{ a }`, reports an error, and exits 1; because `scripts/format` uses `set -e`, formatting stops before Oxfmt. Please collect the unused specifiers for an import and emit one atomic, non-overlapping fix, with regression cases for mixed default/named/namespace and multiple named imports.

scripts/oxlint-plugin.cjs
```diff
@@ -0,0 +1,130 @@
...
+      unused: "'{{name}}' is imported but never used.",
+    },
+  },
+
+  create(context) {
+    const sourceCode = context.sourceCode;
+
+    return {
+      ImportDeclaration(declaration) {
+        for (const variable of sourceCode.getDeclaredVariables(declaration)) {
```
jbeckwith-oai, line 201: **P2 — preserve JSDoc type references before deciding this binding is unused.** In a `// @ts-check` JavaScript file, an import referenced only by `/** @type {DocumentedType} */` has no scope reference here, so `oxlint --fix` deletes it and leaves an unresolved JSDoc type. The previous pinned `eslint-plugin-unused-imports` preserves JSDoc references. Please restore equivalent handling and cover `@type`, `@typedef`, `@param`, and `@returns` cases.

scripts/oxlint-plugin.cjs
```diff
@@ -0,0 +1,130 @@
...
+          if (
+            variable.references.length > 0 ||
+            isUsedThroughSiblingNamespace(sourceCode, declaration, variable.name)
```
jbeckwith-oai: **P2 — this sibling-namespace escape conflates separate bindings and weakens the unused-import gate.** `import * as API from './x'; import { Foo } from './x'; API.Foo` exits 0 here, while the previous ESLint rule correctly reports the separate `Foo` binding as unused. It also preserves unrelated default/aliased imports when their local name happens to match a namespace property. Please remove this exception or narrow it to a demonstrated parser defect without treating namespace property access as use of another binding; add differential alias/default/named tests.

scripts/fast-format
```diff
@@ -24,14 +24,13 @@ if [ ! -f "$FILE_LIST" ]; then
     exit 1
 fi
 
-echo "==> Running eslint --fix"
-ESLINT_FILES="$(grep '\.ts$' "$FILE_LIST" || true)"
-if ! [ -z "$ESLINT_FILES" ]; then
-   echo "$ESLINT_FILES" | xargs ./node_modules/.bin/eslint --cache --fix
+echo "==> Running oxlint --fix"
+LINT_FILES="$(grep -E '\.(ts|mts|cts|js|mjs|cjs)$' "$FILE_LIST" || true)"
```
jbeckwith-oai: **P2 — keep incremental and full lint scopes aligned.** `oxlint .` now checks `.tsx`/`.jsx`, but this filter omits both. I reproduced an unused import in a `.tsx` file surviving `scripts/fast-format`, then failing the full Oxlint run. Please include `tsx|jsx` here (or let Oxlint perform file discovery consistently) and add an incremental regression case.

scripts/fast-format
```diff
@@ -24,14 +24,13 @@ if [ ! -f "$FILE_LIST" ]; then
...
+if [ -n "$LINT_FILES" ]; then
+   echo "$LINT_FILES" | xargs ./node_modules/.bin/oxlint --fix
```
jbeckwith-oai: **P2 — allow ignored-only incremental file lists.** A list containing only `dist/tmp.ts` or `coverage/tmp.ts` reaches this invocation, Oxlint excludes every path via `.oxlintrc.json`, then exits 1 with “No files found to lint”; the prior ESLint helper exited successfully. Oxfmt already uses `--no-error-on-unmatched-pattern` below. Please pass the equivalent flag here and cover an ignored-only list.

jbeckwith-oai (changes requested):
Comprehensive review of the exact current head. I found five reproducible correctness/parity regressions in the new lint/format path. CI and the existing suites are green, but they do not exercise these fixer and incremental-workflow cases; please add committed regression coverage with the fixes.

scripts/oxlint-plugin.cjs
HAYDEN-OAI: Fixed in `ae713c2ea1feb1ff7feccc9438f45329d22c6739`. The rule now collects every unused specifier for an import and emits one declaration-level atomic fix, avoiding overlapping edits. Focused regressions cover mixed default/named, multiple named, mixed default/namespace, and all-unused default/namespace imports; all 5 regression groups and the complete lint/build/type/package pipeline passed.

scripts/oxlint-plugin.cjs
HAYDEN-OAI: Fixed in `ae713c2ea1feb1ff7feccc9438f45329d22c6739`. The plugin now recognizes imported identifiers inside JSDoc type expressions before classifying them as unused. Regression coverage preserves bindings referenced through `@type`, `@typedef`, `@param`, and `@returns`; the focused suite and complete lint/build/type/package pipeline passed.

scripts/fast-format
HAYDEN-OAI: Fixed in `ae713c2ea1feb1ff7feccc9438f45329d22c6739`. `scripts/fast-format` now includes `.tsx` and `.jsx` alongside the existing TS/JS extensions. The end-to-end incremental regression verifies unused imports are removed from both TSX and JSX inputs; the focused suite and complete lint/build/type/package pipeline passed.

scripts/fast-format
HAYDEN-OAI: Fixed in `ae713c2ea1feb1ff7feccc9438f45329d22c6739`. The incremental Oxlint invocation now passes `--no-error-on-unmatched-pattern`, and an ignored-only `dist/**` file-list regression verifies the helper exits successfully. The focused suite and complete lint/build/type/package pipeline passed.

scripts/oxlint-plugin.cjs
HAYDEN-OAI: Fully fixed in follow-up `b43f59d257812878d6ee27193a48786f5a2a7235`. The sibling-namespace exemption is now removed entirely, so `import * as API from './x'; import { Foo } from './x'; API.Foo` removes/reports the separate `Foo` binding just like the prior rule. Differential default, aliased, unrelated named, and exact matching named cases now all require separate unused bindings to be removed. The focused 5/5 regression suite and complete format/lint/build/type/ATTW/publint pipeline passed.

scripts/oxlint-plugin.cjs
```diff
@@ -0,0 +1,113 @@
+/** Import-only unused-binding checks for Oxlint, without ESLint dependencies. */
+
+const JSDOC_TYPE_TAG = /@(?:type|typedef|param|returns?)\s*\{([^}]*)\}/g;
```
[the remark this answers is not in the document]
HAYDEN-OAI: Fixed in `89b8eb5af28ca4f46996a33f94ccd68dbbc85f2c`. The narrow tag whitelist is replaced with balanced JSDoc type-expression extraction: braced type-bearing forms are recognized generally, including `@property`, `@throws`, `@implements`, and `@satisfies`, with nested, multiline, and quoted braces; supported bare structural type tags are handled too. Regressions preserve every requested type-only import while confirming a name mentioned only as prose is still removed. The focused suite, full format/Oxlint/build/type/ATTW/publint pipeline, and all 61 files / 1,453 tests passed.

scripts/oxlint-plugin.cjs
```diff
@@ -0,0 +1,113 @@
...
+
+  if (declaration.importKind === 'type' && bindingStart?.value === 'type') {
+    bindingStart = sourceCode.getTokenAfter(bindingStart);
+  }
+
+  if (!bindingStart || fromToken?.value !== 'from') {
+    return null;
+  }
+
+  return fixer.replaceTextRange([bindingStart.range[0], fromToken.range[0]], `${bindings.join(', ')} `);
```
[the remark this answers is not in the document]
HAYDEN-OAI: Fixed in `89b8eb5af28ca4f46996a33f94ccd68dbbc85f2c`. Commented imports now use surgical specifier/comma removals instead of rebuilding the binding clause, preserving surrounding block and line comments. When a comment belongs to the binding being removed (or a fully unused commented import cannot be preserved safely), the rule deliberately declines the autofix. Regressions cover comments before/after retained bindings, line comments, consecutive unused bindings, mixed default/named imports, and unsafe transformations. The focused suite, full format/Oxlint/build/type/ATTW/publint pipeline, and all 61 files / 1,453 tests passed.

scripts/oxlint-plugin.cjs
```diff
@@ -0,0 +1,196 @@
+/** Import-only unused-binding checks for Oxlint, without ESLint dependencies. */
+
+const JSDOC_BRACED_TYPE_TAG = /@[A-Za-z][\w-]*(?:(?:\s|\*(?=\s))+[^\s{}*]+)?(?:\s|\*(?=\s))*\{/g;
```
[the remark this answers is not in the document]
HAYDEN-OAI: Fully fixed in follow-up `5315c234a8df9c3a8a6f3ff1d117b78713372e9e`. Braced expressions are now accepted only for an audited allowlist of 27 tags whose payload is genuinely a type under JSDoc, TypeScript JSDoc, or Closure syntax. `@external`/`@host` were removed after verifying they take a symbol name, not a type; they and `@lends` are covered as negatives. Module/namespace positives include their required names. All eight focused regressions, the complete format/Oxlint/build/type/ATTW/publint pipeline, and 61 files / 1,453 tests passed.

scripts/oxlint-plugin.cjs
```diff
@@ -0,0 +1,233 @@
...
+  'throws',
+  'type',
+  'typedef',
+  'var',
+  'yield',
+  'yields',
+]);
+const JSDOC_NAMED_TYPE_TAGS = new Set(['arg', 'argument', 'param', 'prop', 'property']);
+const JSDOC_BRACED_TYPE_TAG =
+  /(?:^|[\s*])@([A-Za-z][\w-]*)(?:(?:\s|\*(?=\s))+([^\s{}*@]+))?(?:\s|\*(?=\s))*\{/g;
```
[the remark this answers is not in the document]
HAYDEN-OAI: Fixed in `5315c234a8df9c3a8a6f3ff1d117b78713372e9e`. Both braced and bare tags now match only at actual JSDoc line boundaries, so tag-looking source inside `@example` bodies—including the exact multiline `// @type {ExampleOnly}` case—does not preserve stale imports. Regressions also cover inline/unquoted examples, bare `@implements`, Unicode line separators, and real typed tags following an example block. All eight focused regressions, the complete format/Oxlint/build/type/ATTW/publint pipeline, and 61 files / 1,453 tests passed.

scripts/oxlint-plugin.cjs
```diff
@@ -0,0 +1,233 @@
...
+        if (character === '\\') {
+          end++;
+        } else if (character === quote) {
+          quote = undefined;
+        }
+        continue;
+      }
+
+      if (character === '"' || character === "'") {
+        quote = character;
```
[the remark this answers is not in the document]
HAYDEN-OAI: Fixed in `5315c234a8df9c3a8a6f3ff1d117b78713372e9e`. The balanced type-expression scanner now treats backticks as quoted text and correctly enters/exits `${...}` interpolation, including escaped backticks/backslashes and nested template literals. The exact `Record<`}`, Foo>` case now preserves `Foo`, with additional opening-brace, escaped, interpolated, and nested regressions. All eight focused regressions, the complete format/Oxlint/build/type/ATTW/publint pipeline, and 61 files / 1,453 tests passed.

scripts/oxlint-plugin.cjs
```diff
@@ -0,0 +1,242 @@
...
+  'throws',
+  'type',
+  'typedef',
+  'var',
+  'yield',
+  'yields',
+]);
+const JSDOC_NAMED_TYPE_TAGS = new Set(['arg', 'argument', 'param', 'prop', 'property']);
+const JSDOC_BRACED_TYPE_TAG =
+  /(?:^|[\r\n\u2028\u2029])[\t ]*\*?[\t ]*@([A-Za-z][\w-]*)(?:(?:\s|\*(?=\s))+([^\s{}*@]+))?(?:\s|\*(?=\s))*\{/g;
```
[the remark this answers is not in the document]
HAYDEN-OAI: Fixed in `722062a8177ba8891fbaf0a325b6a39b64156d3d`. The regex-only boundary logic is replaced with a stateful JSDoc scanner that recognizes genuine same-line sibling tags, including the exact `@param {Foo} x @returns {Bar}` case, while ignoring tag-like text inside `@example`, quoted/backticked prose, email-like text, and nested braces/templates. Focused coverage exercises 44 genuine inline type references plus 31 example/prose negatives, aliases, bare tags, nested generics/arrows, and Unicode whitespace. The 10 regression groups, full format/Oxlint/build/type/ATTW/publint pipeline, Node-version policy, and all 61 files / 1,453 unit tests passed.

scripts/oxlint-plugin.cjs
```diff
@@ -0,0 +1,392 @@
...
+      if (templateInterpolationDepths.at(-1) === depth) {
+        templateInterpolationDepths.pop();
+        quote = '`';
+      }
+    }
+  }
+}
+
+function readJSDocBareType(comment, start) {
+  const identifier = /^[A-Za-z_$][\w$.]*/u.exec(comment.slice(start));
```
jbeckwith-oai: **P1 — parse the whole bare JSDoc type before deciding which imports are used.** This starts with one identifier and stops after it unless a generic follows. I reproduced `/** @type Left | Right */`: the new Oxlint rule preserves `Left` but reports and autofixes away `Right`, while the previous pinned ESLint rule preserves both. Modifier/parenthesized forms such as `?Foo` and `(Foo | null)` lose `Foo` entirely. TypeScript accepts these as complete bare type expressions, so `scripts/format` can leave unresolved JSDoc. Please parse the complete expression (unions/intersections/modifiers/parentheses/tuples), or restore the prior compatibility matcher, and add differential cases.

scripts/oxlint-regression.test.cjs
```diff
@@ -0,0 +1,632 @@
...
+    'yields',
+  ];
+  const documentationTags = [
+    'example',
+    'external',
+    'host',
+    'deprecated',
+    'description',
+    'desc',
+    'see',
```
jbeckwith-oai: **P2 — preserve the documentation references the removed rule deliberately treated as uses.** These tests currently classify `see`, `link`, `linkcode`, and `linkplain` as non-uses, but `eslint-plugin-unused-imports@4.4.1` explicitly preserved `@see Foo`, `@link Foo`, and inline `{@link Foo}` / `{@linkcode Foo}` / `{@linkplain Foo}`. I reproduced all of those being removed by the new fixer while the base ESLint stack exits cleanly. That breaks symbol-resolved API documentation links; please restore the old behavior and cover both block and inline forms.

scripts/oxlint-plugin.cjs
```diff
@@ -0,0 +1,392 @@
...
+        index = getNextJSDocLine(comment, type?.end ?? afterTag);
+        break;
+      }
+      index = sibling;
+    }
+  }
+}
+
+function isUsedThroughJSDoc(sourceCode, name) {
+  const identifier = new RegExp(`(^|[^A-Za-z0-9_$])${escapeRegExp(name)}(?=$|[^A-Za-z0-9_$])`);
```
jbeckwith-oai: **P2 — use JavaScript identifier boundaries, not ASCII-only boundaries.** With `import { Foo as Café } ...` and `/** @type {CaféÉ} */`, this considers `Café` used because `É` is outside `[A-Za-z0-9_$]`; the previous rule correctly reports `Café` unused. Valid Unicode identifier-prefix collisions can therefore hide stale imports. Please use Unicode `ID_Continue`-aware boundaries (or structural parsing) and add a regression.

jbeckwith-oai (changes requested):
Re-review of exact head `722062a8177ba8891fbaf0a325b6a39b64156d3d`. Four of the five previously raised blockers are fixed: atomic multi-binding edits, sibling-namespace handling, TSX/JSX incremental linting, and ignored-only file lists. Braced JSDoc references are substantially improved, but JSDoc parity remains incomplete and the fixer still deletes valid references in the cases below. The compatibility shim has now grown into a hand-written partial JSDoc parser with hundreds of lines of parser-specific tests; the misses below show why this is too brittle. Please prefer the removed rule's compact compatibility predicate for exact behavior parity or a maintained JSDoc/type parser instead of extending this grammar case-by-case. I also confirmed the existing open P2 about string literals/property keys in type text, so I have not duplicated that inline. Validation on this exact head: `pnpm lint` passed end-to-end, all 10 lint regressions passed, 61 unit files / 1,453 tests passed, shell syntax and diff checks passed, and current CI is green.

scripts/oxlint-plugin.cjs
HAYDEN-OAI: Fixed in `51c76c4df4d15d5d98f1e210c243b0944ff7864c`. Bare JSDoc types are now scanned as complete balanced expressions instead of stopping after the first identifier/generic, preserving every real reference in unions, intersections, `?`/`!` modifiers, parentheses, tuples/arrays, `keyof`/`typeof`, functions/constructors, conditionals, and nested generics while still stopping before prose or the next tag. The exact `Left | Right`, `?Foo`, and `(Foo | null)` cases are covered. All 15 regression groups pass on Node 22/24/26, with the full lint/build/type/package pipeline and unit suites green.

scripts/oxlint-plugin.cjs
```diff
@@ -0,0 +1,392 @@
...
+
+  return sourceCode.getAllComments().some((comment) => {
+    if (comment.type !== 'Block' || !comment.value.startsWith('*')) {
+      return false;
+    }
+
+    return Array.from(getJSDocTypeExpressions(comment.value)).some((typeExpression) =>
+      identifier.test(typeExpression),
```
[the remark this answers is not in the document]
HAYDEN-OAI: Fixed in `51c76c4df4d15d5d98f1e210c243b0944ff7864c`. JSDoc type text is now tokenized structurally before testing an imported binding: string/numeric literals, object and quoted property keys, qualified member names, parameter/property binders, mapped/infer/generic binders, and shadowed occurrences no longer count as references, while genuine value/type positions do. Focused positives and negatives cover objects, signatures, mapped/conditional/indexed/template types, binders, and namespace access. All 15 regression groups pass on Node 22/24/26; the full Node 24 format/Oxlint/build/TypeScript/package/publint pipeline and 61 files / 1,453 unit tests on Node 22 and 24 also pass.

scripts/oxlint-regression.test.cjs
HAYDEN-OAI: Fixed in `51c76c4df4d15d5d98f1e210c243b0944ff7864c`. The compatibility behavior is restored for genuine block `@see`/`@link`/`@linkcode`/`@linkplain` targets and inline `{@link ...}` variants. The scanner preserves qualified/hash/tilde symbol targets but rejects labels, member-only matches, URLs/mail/module targets, quoted/backticked text, escaped lookalikes, and example bodies. All 15 focused regression groups pass on Node 22/24/26; the full Node 24 lint/build/type/package pipeline and 61 files / 1,453 tests on Node 22 and 24 pass.

scripts/oxlint-plugin.cjs
HAYDEN-OAI: Fixed in `51c76c4df4d15d5d98f1e210c243b0944ff7864c`. Identifier scanning now uses Unicode `ID_Start`/`ID_Continue` semantics, including combining marks, ZWNJ/ZWJ, astral code points, and `\uXXXX`/`\u{...}` escapes. The exact `Café` versus `CaféÉ` prefix collision now removes the stale import, while exact Unicode and escaped references are preserved. Focused coverage includes combining, astral, escaped-prefix, and joiner boundaries. All 15 regression groups pass on Node 22/24/26, with the full lint/build/type/package pipeline and unit suites green.

scripts/oxlint-plugin.cjs
```diff
@@ -0,0 +1,898 @@
...
+
+  create(context) {
+    const sourceCode = context.sourceCode;
+
+    return {
+      ImportDeclaration(declaration) {
+        const unusedSpecifiers = [];
+
+        for (const variable of sourceCode.getDeclaredVariables(declaration)) {
+          if (variable.references.length > 0 || isUsedThroughJSDoc(sourceCode, variable.name)) {
```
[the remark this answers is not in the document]
HAYDEN-OAI: Fixed in `253fee9d091cb9fa89c3036329bbcde40142a3b6`. The rule now preserves the scope-resolved import used by classic JSX elements/fragments, honoring configured factories plus leading `@jsx`, `@jsxFrag`, `@jsxRuntime`, and `@jsxImportSource` pragmas. Automatic-runtime files do not retain classic factories; fragment factories are retained only for fragments; lexical shadowing is handled at each JSX site. All 19 regression groups pass on Node 22/24/26; 61 files / 1,453 tests pass on Node 22 and 24; the full format/Oxlint/build/type/package/publint pipeline and packed CJS/ESM/runtime policy checks pass.

scripts/oxlint-plugin.cjs
```diff
@@ -0,0 +1,898 @@
...
+  return sourceCode.getAllComments().some((comment) => {
+    if (comment.type !== 'Block' || !comment.value.startsWith('*')) {
+      return false;
+    }
+
+    const expressions = [
+      ...getJSDocTypeExpressions(comment.value),
+      ...getJSDocInlineLinkExpressions(comment.value),
+    ];
+    return expressions.some((expression) => isJSDocTypeReference(expression, name));
```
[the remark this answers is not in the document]
HAYDEN-OAI: Fixed in `253fee9d091cb9fa89c3036329bbcde40142a3b6`. JSDoc `@template` declarations are now parsed as comment-local binders, so the exact `import { Foo as T } ... /** @template T @param {T} value */` case removes the unused import. Constraint/default expressions still retain genuine imported types, with coverage for multiple and multiline binders, TypeScript modifiers, defaults, Unicode names, and cross-comment isolation. All 19 regression groups pass on Node 22/24/26; 61 files / 1,453 tests pass on Node 22 and 24; the full format/Oxlint/build/type/package/publint pipeline and packed CJS/ESM/runtime policy checks pass.

scripts/oxlint-plugin.cjs
```diff
@@ -0,0 +1,1152 @@
...
+
+        while (index < lineEnd && comment[index] !== '}') index++;
+      }
+    }
+
+    lineStart = getNextJSDocLine(comment, lineEnd);
+  }
+}
+
+function tokenizeJSDocType(expression) {
```
jbeckwith-oai: **P1 — replace this bespoke compiler instead of extending it further.** The shim is now 1,152 lines and its regression file is 1,620 lines; both crossed the thermo-nuclear 1k-file bar. It now owns JSDoc tag parsing, Unicode escape lexing, TypeScript-type tokenization/delimiter matching/scope resolution, template-parameter semantics, JSX pragma/runtime parsing, AST traversal, and JSX scope lookup. This follow-up added another 760 implementation lines and 988 test lines since the last reviewed `722062a8`, yet the exact head still corrupts valid `keyof` types, generic member scopes, quoted links, and JSX pragmas in the cases below. Please use the existing TypeScript/JSDoc AST (TypeScript is already a dev dependency), a maintained parser, or the removed rule's compact compatibility predicate. Merely splitting this partial grammar across files would not fix the design.

scripts/oxlint-plugin.cjs
```diff
@@ -0,0 +1,1152 @@
...
+        '=',
+        '.',
+        '?.',
+        '#',
+        '~',
+      ].includes(previous.value)
+    );
+  }
+
+  if (comment[next] === '(') return ['function', 'new'].includes(previous.value);
```
jbeckwith-oai: **P1 — this still truncates valid grouped/object operands after `keyof`.** For `import { Foo, Bar } ...; /** @type keyof (Foo & Bar) */`, the extracted expression is only `keyof`, so `oxlint --fix` deletes both required imports. `keyof { key: Foo }` fails the same way. TypeScript 5.8 accepts the complete types and the previous ESLint rule preserves the imports. Please stop heuristically delimiting TypeScript syntax here or cover the complete grammar with a canonical parser.

scripts/oxlint-plugin.cjs
```diff
@@ -0,0 +1,1152 @@
...
+        quote = undefined;
+        index += 2;
+      } else {
+        if (character === quote) quote = undefined;
+        index++;
+      }
+      continue;
+    }
+
+    if (/\s/u.test(character)) {
```
jbeckwith-oai: **P1 — preserve line boundaries when resolving generic binder scope.** Discarding every newline here means the later scope scan can stop only at `,`/`;`. In `@type {{ method<Ref>(): Ref\n other: Ref }}`, TypeScript treats the newline as a member separator, so the method's `Ref` binder does not shadow `other: Ref`. This rule extends that binder through the next member, reports/deletes the imported `Ref`, and `tsc --checkJs` then emits `TS2304` at `other: Ref`. Use the canonical AST/source ranges or retain line-separator tokens and add this differential case.

scripts/oxlint-plugin.cjs
```diff
@@ -0,0 +1,1152 @@
...
+  for (const comment of sourceCode.getAllComments()) {
+    if (comment.range[1] > firstStatementStart) break;
+    if (comment.type !== 'Block') continue;
+
+    let inExample = false;
+    for (const rawLine of comment.value.split(/\r\n|[\n\r\u2028\u2029]/u)) {
+      let line = rawLine.trimStart();
+      if (line.startsWith('*')) line = line.slice(1).trimStart();
+
+      const tag = /^@([A-Za-z][\w-]*)(?=\s|$)/u.exec(line);
```
jbeckwith-oai: **P1 — parse classic JSX pragmas using the compiler's actual rules.** Requiring `@jsx` at the trimmed line start misses an ordinary leading comment such as `/* prose @jsx h */`. TypeScript recognizes that pragma and emits `<div />` as `h("div", null)`, but this rule does not mark `h` used and `oxlint --fix` removes its import. Please derive pragma semantics from TypeScript (or otherwise match its scanner) and add this exact emit/fix regression.

scripts/oxlint-regression.test.cjs
```diff
@@ -0,0 +1,1620 @@
...
+    },
+    {
+      name: 'InlineUrlOnly',
+      comment: '/** Description {@link https://example.com/InlineUrlOnly}. */',
+      used: false,
+    },
+    { name: 'InlineModuleOnly', comment: '/** Description {@link module:InlineModuleOnly}. */', used: false },
+    { name: 'ExampleInlineOnly', comment: '/** @example {@link ExampleInlineOnly} */', used: false },
+    { name: 'ExampleBlockOnly', comment: '/** @example @see ExampleBlockOnly */', used: false },
+    { name: 'QuotedInlineOnly', comment: '/** Description "{@link QuotedInlineOnly}". */', used: false },
```
jbeckwith-oai: **P2 — this is a valid inline documentation link, not a non-use.** Quotes in JSDoc prose do not disable structured `{@link ...}` tags. The current scanner skips everything inside quotes/backticks, so both `/** See "{@link Foo}". */` and a quoted `@deprecated` link cause the imported `Foo` to be autofixed away; the previous pinned rule preserves it. Mark these cases used and parse the inline tag independently of surrounding prose punctuation.

jbeckwith-oai (changes requested):
Re-review of exact current head `76e515fa824cd6efa8c51ec2307fcc58d13a51f5` after the base merge. The earlier atomic-fix, sibling-namespace, incremental-file, direct bare-type/link/Unicode, template-binder, and basic JSX-factory cases are covered, but the replacement rule is still unsafe and has become a structural blocker. The relevant plugin/test/config blobs are identical to `253fee9d`; the findings below reproduce against that exact code. Exact-head validation: 19/19 focused regressions passed in 22.2s, shell syntax and diff checks passed. Full lint and 61 files / 1,453 unit tests passed on the identical pre-merge tooling tree; current exact-head CI is still running after the base merge.

masomelelemgqamqo-rgb:
True all

Sent from Outlook for Android<https://aka.ms/AAb9ysg>
________________________________

scripts/oxlint-plugin.cjs
HAYDEN-OAI: Fixed in `22a138b0ab51854471cd6141dbf241232422fd4c`. The 1,152-line hand-built JSDoc/type/JSX parser was replaced with TypeScript’s maintained compiler JSDoc AST and pragma scanner; the plugin is now 605 lines and the regression suite 881 lines (down from 1,620), without splitting the bespoke grammar. All 21 import-rule groups pass on Node 22/24/26, all 61 files / 1,453 unit tests pass on each runtime, and the full format/Oxlint/build/type/package/publint plus packed CJS/ESM checks pass.

scripts/oxlint-plugin.cjs
HAYDEN-OAI: Fixed in `22a138b0ab51854471cd6141dbf241232422fd4c`. Generic scope now comes from TypeScript’s canonical JSDoc/type AST, so the method-local `Ref` in `method<Ref>(): Ref` does not leak across the newline/member boundary and the imported `Ref` used by `other: Ref` is preserved. The exact compiler-differential regression passes in the 21-group Node 22/24/26 matrix.

scripts/oxlint-plugin.cjs
HAYDEN-OAI: Fixed in `22a138b0ab51854471cd6141dbf241232422fd4c`. JSDoc types are now parsed by TypeScript’s compiler AST rather than heuristic token delimiting, and compiler-backed regressions cover both `keyof (Foo & Bar)` and `keyof { key: Foo }`; the required imports are retained. All 21 regression groups pass on Node 22/24/26, together with the full repository and package validation pipeline.

scripts/oxlint-plugin.cjs
HAYDEN-OAI: Fixed in `22a138b0ab51854471cd6141dbf241232422fd4c`. Classic JSX pragmas now use TypeScript’s actual pragma scanning; the exact leading `/* prose @jsx h */` case matches compiler emit and retains `h`. The focused compiler-backed case and all 21 regression groups pass on Node 22/24/26, with full build/type/package validation green.

scripts/oxlint-regression.test.cjs
HAYDEN-OAI: Fixed in `22a138b0ab51854471cd6141dbf241232422fd4c`. Inline JSDoc links are collected from TypeScript’s structured documentation nodes independently of surrounding quotes/backticks, including quoted `@deprecated` prose and a mixed quoted/backticked-link regression found during final audit. The removed ESLint rule confirms the compatibility behavior; all 21 regression groups pass on Node 22/24/26.

scripts/oxlint-plugin.cjs
```diff
@@ -0,0 +1,605 @@
...
+    const jsDocUsedImports = getJSDocImportUsage(sourceCode);
+
+    return {
+      ImportDeclaration(declaration) {
+        const unusedSpecifiers = [];
+        for (const variable of sourceCode.getDeclaredVariables(declaration)) {
+          if (
+            variable.references.some((reference) => reference.identifier !== variable.identifiers[0]) ||
+            jsxUsedImports.has(variable) ||
+            jsDocUsedImports.has(variable.name)
```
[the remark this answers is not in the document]
HAYDEN-OAI: Fixed in `17152d6d2aa2f9c1d942ba522d54a7f5eb01fe56`. JSDoc type references are now matched to the actual Oxlint/TypeScript lexical binding rather than a file-wide name set, so a function-local `class Foo` no longer keeps an unrelated module-level `import { Foo }`; genuinely imported and shadowed references retain their correct behavior. The exact regression passes in all 24 import-rule groups on Node 22/24/26, alongside 61 files / 1,453 unit tests per runtime and the full format/Oxlint/build/type/package/publint pipeline.

scripts/oxlint-plugin.cjs
```diff
@@ -0,0 +1,605 @@
...
+      const startsLine = startsJSDocLine(parsed.text, originalTag.pos);
+
+      if (inExample && !startsLine) continue;
+      if (inExample && startsLine) inExample = false;
+      if (name === 'example') {
+        inExample = true;
+        continue;
+      }
+
+      if (!startsLine && isWithinTemplateProse(parsed.text, previousTag, originalTag)) continue;
```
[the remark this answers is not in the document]
HAYDEN-OAI: Fixed in `17152d6d2aa2f9c1d942ba522d54a7f5eb01fe56`. Template-prose detection now reacts only to TypeScript’s specific unterminated-template diagnostic, rather than treating unrelated syntax diagnostics from commas or ordinary prose as an open template. The exact `@param {Foo} x \`a\`, \`b\` @returns {Bar}` regression recognizes the sibling tag and preserves `Bar`; all 24 regression groups pass on Node 22/24/26.

scripts/oxlint-plugin.cjs
```diff
@@ -0,0 +1,605 @@
...
+
+      if (
+        documentationTag &&
+        !startsLine &&
+        hasUnclosedBrace(parsed.text.slice(documentationTag.tagName.end, originalTag.pos))
+      ) {
+        continue;
+      }
+
+      if (!TYPE_BEARING_TAGS.has(name) && !REFERENCE_TAGS.has(name)) {
```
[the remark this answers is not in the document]
HAYDEN-OAI: Fixed in `17152d6d2aa2f9c1d942ba522d54a7f5eb01fe56`. The compiler-backed traversal now descends into nested `JSDocSignature` nodes carried by `@callback` and `@overload`, so imported types used by child `@param`/`@returns` tags are retained. Exact callback and overload regressions pass within the 24-group Node 22/24/26 matrix; full build/type/package and packed CJS/ESM validation is green.

scripts/oxlint-plugin.cjs
```diff
@@ -0,0 +1,632 @@
...
+  if (ts.isTypePredicateNode(node)) {
+    collectTypeReferences(node.type, bindings, used);
+    return;
+  }
+
+  if (node.typeParameters?.length) {
+    const innerBindings = new Set(bindings);
+    for (const parameter of node.typeParameters) {
+      collectTypeReferences(parameter.constraint, bindings, used);
+      collectTypeReferences(parameter.default, bindings, used);
```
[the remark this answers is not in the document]
HAYDEN-OAI: Fixed in `71a534c94e43a4e08bedbb47dfa962e69d4ff1c5`. Generic type parameters are now processed sequentially: each parameter’s constraint/default sees binders declared by earlier parameters, so `T` in `<T, U = T>` resolves to the generic binder and no longer keeps an unrelated stale import. Compiler-differential coverage also verifies genuine imported references plus same/later-parameter and observable side-effect cases. All 25 import-rule groups and all 61 files / 1,453 tests pass on Node 22/24/26; the full format/Oxlint/build/type/ATTW/publint, Node-policy, and packed CJS/ESM pipeline is green. The touched files remain below the maintainability bar at 632 and 998 lines.

scripts/oxlint-plugin.cjs
```diff
@@ -0,0 +1,632 @@
...
+    mappedBindings.add(ts.unescapeLeadingUnderscores(node.typeParameter.name.escapedText));
+    collectTypeReferences(node.nameType, mappedBindings, used);
+    collectTypeReferences(node.type, mappedBindings, used);
+    return;
+  }
+
+  if (ts.isConditionalTypeNode(node)) {
+    collectTypeReferences(node.checkType, bindings, used);
+    collectTypeReferences(node.extendsType, bindings, used);
+    const trueBindings = new Set([...bindings, ...getInferBindings(node.extendsType)]);
```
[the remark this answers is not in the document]
HAYDEN-OAI: Fixed in `58955556995c1fd5322eed3a45b83f207fcca3f8`. `getInferBindings` now stops at nested conditional-type nodes, keeping an inner `infer X` scoped to that inner conditional while preserving a genuine imported `X` used by the outer true branch. The compiler-differential regression verifies the inner symbol resolves to the `TypeParameter` and the outer `X` to the `ImportSpecifier`, and it fails on the prior head. All 25 import-rule regression groups and all 61 files / 1,453 tests pass on Node 22, 24, and 26; format, Oxlint, build, TypeScript, ATTW, publint, Node-policy, and packed CJS/ESM validation also pass.

scripts/oxlint-plugin.cjs
```diff
@@ -0,0 +1,632 @@
...
+      }
+
+      documentationTag = undefined;
+      const tag = templates.get(originalTag) ?? normalizeJSDocTag(originalTag, parsed.text);
+      if (!tag) continue;
+
+      if (ts.isJSDocTemplateTag(tag)) {
+        collectTypeReferences(tag.constraint, new Set(), references);
+        for (const parameter of tag.typeParameters) {
+          collectTypeReferences(parameter.default, new Set(), references);
```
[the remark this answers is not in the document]
HAYDEN-OAI: Fixed in `58955556995c1fd5322eed3a45b83f207fcca3f8`. JSDoc template parameters are now accumulated in declaration order, so `@template T, [U=T]` and the equivalent separate-tag form treat `T` as the prior comment-local binder while a genuine imported default such as `[U=Ref]` remains retained. The dedicated compiler-differential regressions fail on the prior head and pass after the fix. All 25 import-rule regression groups and all 61 files / 1,453 tests pass on Node 22, 24, and 26; format, Oxlint, build, TypeScript, ATTW, publint, Node-policy, and packed CJS/ESM validation also pass.

scripts/oxlint-plugin.cjs
```diff
@@ -0,0 +1,635 @@
...
+    messages: { unused: 'Imported bindings {{names}} are never used.' },
+  },
+
+  create(context) {
+    const sourceCode = context.sourceCode;
+    const jsxUsedImports = getJSXImportUsage(context);
+    const jsDocUsedImports = getJSDocImportUsage(sourceCode);
+
+    return {
+      ImportDeclaration(declaration) {
```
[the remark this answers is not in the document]
HAYDEN-OAI: Fixed in `fbc8ff77138373a4b75c78ddf451529048552c15`. The plugin now handles `TSImportEqualsDeclaration` for both external `require(...)` and qualified aliases, while retaining referenced, JSDoc-used, type-used, and exported aliases. Focused regressions verify removal of emitted require/getter side effects and deliberately leave commented declarations as diagnostics without applying a destructive fix. All 29 import-rule groups and all 61 files / 1,453 tests pass on Node 22, 24, and 26; whole-tree Oxfmt/Oxlint, build, TypeScript, ATTW, publint, Node policy, and packed CJS/ESM checks also pass.

scripts/oxlint-plugin.cjs
```diff
@@ -0,0 +1,635 @@
...
+      : 'type';
+  const tail = text.slice(tag.tagName.end, tag.end);
+  return parseJSDocComment(`/** @${canonicalName}${tail} */`).comment?.tags?.[0];
+}
+
+function getJSDocRootBinding(sourceCode, comment, name) {
+  const token = sourceCode.getTokenAfter(comment) ?? sourceCode.getTokenBefore(comment);
+  const node = token && sourceCode.getNodeByRangeIndex(token.range[0]);
+
+  for (let scope = sourceCode.getScope(node ?? sourceCode.ast); scope; scope = scope.upper) {
```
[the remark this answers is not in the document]
HAYDEN-OAI: Fixed in `fbc8ff77138373a4b75c78ddf451529048552c15`. JSDoc references now retain type/value namespace information, and `typeof` lookups for variable-attached arrow/function expressions enter the initializer scope so parameter bindings win without masking genuine type imports. Compiler-backed cases cover arrow, function-expression, exported, default, and destructured parameters; the pre-fix cases fail and the fixed cases remove the stale import and observable module side effect while retaining genuine imports. All 29 import-rule groups and all 61 files / 1,453 tests pass on Node 22, 24, and 26; whole-tree Oxfmt/Oxlint, build, TypeScript, ATTW, publint, Node policy, and packed CJS/ESM checks also pass.

scripts/lint
```diff
@@ -23,15 +23,18 @@ EOF
...
+echo "==> Running oxfmt --check"
+./node_modules/.bin/oxfmt --check .
 
-echo "==> Running eslint"
-./node_modules/.bin/eslint .
+echo "==> Running oxlint"
+./node_modules/.bin/oxlint .
+
+echo "==> Running Oxlint regressions"
+node --test scripts/oxlint-regression.test.cjs
```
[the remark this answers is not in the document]
HAYDEN-OAI: This is already covered by the CI-invoked entrypoint. `scripts/oxlint-regression.test.cjs` explicitly executes `require('./oxlint-review-regression.test.cjs')`, and the required module registers its four `node:test` groups in the same runner. Both `./scripts/lint` and `pnpm test:lint-regressions` therefore ran 29 groups (the prior 25 plus these 4) successfully on the pushed head; the same 29-group total also passed independently on Node 22, 24, and 26. Adding the second path would execute the required module twice, so no code change is appropriate.

scripts/oxlint-plugin.cjs
```diff
@@ -0,0 +1,671 @@
...
+      documented?.type === 'VariableDeclaration' ? documented.declarations[0]?.init : undefined;
+    if (initializer?.type === 'ArrowFunctionExpression' || initializer?.type === 'FunctionExpression') {
+      const parameter = sourceCode.getScope(initializer).set.get(name);
+      if (parameter?.defs.some((definition) => definition.type === 'Parameter')) return parameter;
+    }
+  }
+
+  for (let scope = sourceCode.getScope(node ?? sourceCode.ast); scope; scope = scope.upper) {
+    const [start, end] = scope.block.range;
+    if (comment.range[0] < start || comment.range[1] > end) continue;
```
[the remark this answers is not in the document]
HAYDEN-OAI: Fixed in `ee837bb39ce0676b1e25ab08a77c3374037a30ec`. Value-namespace JSDoc lookups now unwrap export/key hosts and enter the callable scope for function declarations, class/object methods, constructors, and setters, so `typeof` parameter references no longer retain same-named imports; genuine type references and getter value references still retain the import. Compiler-backed pre-fix failures and runtime side-effect cases cover named/exported/default functions, instance/static/object methods, constructors, and setters. All 31 import-rule groups and all 61 files / 1,453 tests pass on Node 22, 24, and 26; whole-tree Oxfmt/Oxlint, build, TypeScript, ATTW, publint, Node policy, and packed CJS/ESM checks also pass.

scripts/oxlint-plugin.cjs
```diff
@@ -0,0 +1,671 @@
...
+    const variable = scope.set.get(name);
+    if (variable) return variable;
+  }
+}
+
+function getJSDocImportUsage(sourceCode) {
+  const used = new Set();
+
+  for (const comment of sourceCode.getAllComments()) {
+    if (comment.type !== 'Block' || !comment.value.startsWith('*')) continue;
```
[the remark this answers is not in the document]
HAYDEN-OAI: Fixed in `ee837bb39ce0676b1e25ab08a77c3374037a30ec`. Import-use collection now builds the attached-JSDoc ranges from TypeScript's parsed source tree and ignores blocks attached only to empty statements/EOF or to no real host, instead of falling back to the token before a detached comment. Focused regressions cover EOF, same-line EOF, consecutive EOF, between-statement, empty-statement, and before-closing-brace cases, and verify the stale import's observable side effect disappears. All 31 import-rule groups and all 61 files / 1,453 tests pass on Node 22, 24, and 26; whole-tree Oxfmt/Oxlint, build, TypeScript, ATTW, publint, Node policy, and packed CJS/ESM checks also pass.

scripts/oxlint-plugin.cjs
```diff
@@ -0,0 +1,708 @@
...
+  }
+}
+
+function getAttachedJSDocComments(sourceCode) {
+  const compilerSource = ts.createSourceFile(
+    'oxlint-jsdoc.tsx',
+    sourceCode.text,
+    { languageVersion: ts.ScriptTarget.Latest, jsDocParsingMode: ts.JSDocParsingMode.ParseAll },
+    true,
+    ts.ScriptKind.TSX,
```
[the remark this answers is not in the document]
HAYDEN-OAI: Fixed in `a83c977276e1314006a54fdbe68bb2dd0ac203e8`. The attachment parser now derives `ScriptKind` from physical/virtual filenames and parser file-path signals, supports explicit `?lang=...` modes, and falls back to the parser's JSX signal for unknown names. Red/green coverage includes `.ts`, `.mts`, `.cts`, `.tsx`, `.jsx`, `.js`, `.mjs`, physical/filePath overrides, query-mode virtual files, and unknown virtual filenames; genuine JSDoc imports remain while stale imports are removed. All 35 import-rule groups and all 61 files / 1,453 tests pass on Node 22, 24, and 26; whole-tree Oxfmt/Oxlint, build, TypeScript, ATTW, publint, Node policy, and packed CJS/ESM checks also pass.

scripts/oxlint-plugin.cjs
```diff
@@ -0,0 +1,708 @@
...
+      const parameter = sourceCode.getScope(callable).set.get(name);
+      if (parameter?.defs.some((definition) => definition.type === 'Parameter')) return parameter;
+    }
+  }
+
+  for (let scope = sourceCode.getScope(node ?? sourceCode.ast); scope; scope = scope.upper) {
+    const [start, end] = scope.block.range;
+    if (comment.range[0] < start || comment.range[1] > end) continue;
+    const variable = scope.set.get(name);
+    if (variable) return variable;
```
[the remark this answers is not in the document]
HAYDEN-OAI: Fixed in `a83c977276e1314006a54fdbe68bb2dd0ac203e8`. Type-namespace resolution now skips value-only variables/functions/parameters and continues outward, while recognizing imports, classes, type aliases, interfaces, enums, namespaces, and type parameters as type-bearing bindings. `typeof` continues to use value-namespace resolution. Compiler-backed red/green cases verify imported-type retention, local type-bearing shadow removal, and runtime import side effects for both type and value counterexamples. All 35 import-rule groups and all 61 files / 1,453 tests pass on Node 22, 24, and 26; whole-tree Oxfmt/Oxlint, build, TypeScript, ATTW, publint, Node policy, and packed CJS/ESM checks also pass.

scripts/oxlint-plugin.cjs
```diff
@@ -0,0 +1,752 @@
...
+    'oxlint-jsdoc',
+    sourceCode.text,
+    { languageVersion: ts.ScriptTarget.Latest, jsDocParsingMode: ts.JSDocParsingMode.ParseAll },
+    true,
+    scriptKind,
+  );
+  const attached = new Map();
+
+  function visit(node) {
+    if (!ts.isEmptyStatement(node) && node.kind !== ts.SyntaxKind.EndOfFileToken) {
```
jbeckwith-oai: **P2 — preserve standalone EOF JSDoc typedefs.** TypeScript deliberately attaches a trailing module-level `/** @typedef {Foo} Bar */` to `EndOfFileToken`, so this blanket exclusion drops a semantically live declaration. Repro:
```js
// @ts-check
import { Foo } from './types.js';

/** @typedef {Foo} Bar */
```
The prior ESLint rule preserves `Foo`, while `oxlint --fix --no-ignore` deletes the import and a subsequent TypeScript 5.8 `checkJs` run reports TS2304 (`Cannot find name 'Foo'`). Please distinguish genuine standalone declaration tags (at least `JSDocTypedefTag`) from orphan annotations when handling EOF-hosted comments, and add a regression that runs the fixer and then verifies the typedef still resolves.

jbeckwith-oai (changes requested):
Re-review of exact head `a83c977276e1314006a54fdbe68bb2dd0ac203e8`. The prior atomic-fix, namespace-binding, incremental-file, JSDoc expression/link/Unicode/template-scope, and JSX pragma blockers are addressed; replacing the hand-written type grammar with TypeScript's JSDoc AST is a substantial structural improvement. One reproducible autofix correctness blocker remains for a valid TypeScript JSDoc host, described inline. Validation: 35/35 focused lint regression groups passed; 61 files / 1,453 unit tests passed; full lint/build/type/package validation and exact-head CI are green.

scripts/oxlint-plugin.cjs
HAYDEN-OAI: Fixed in `8a4b1b5c7e72aacf861d4673b247ab9bfe6e8f24`. The EOF attachment filter now preserves a comment only when TypeScript’s JSDoc AST contains a genuine `JSDocTypedefTag`; detached/orphan EOF annotations remain excluded. The new regression proves TypeScript attaches the standalone typedef to `EndOfFileToken`, runs the autofixer, verifies the `Foo` import remains and TS 5.8 `checkJs` stays clean, and confirms a neighboring orphan does not retain an unrelated import. All 36 import/JSDoc groups, whole-tree Oxfmt/Oxlint, build/type/ATTW/publint, packed CJS/ESM checks, and 61 files / 1,453 tests on Node 22/24/26 passed.

scripts/oxlint-plugin.cjs
```diff
@@ -0,0 +1,759 @@
...
+}
+
+function getJSDocRootBinding(sourceCode, comment, name, namespace) {
+  const token = sourceCode.getTokenAfter(comment) ?? sourceCode.getTokenBefore(comment);
+  const node = token && sourceCode.getNodeByRangeIndex(token.range[0]);
+
+  if (namespace === 'value') {
+    let documented = node;
+    if (documented?.type === 'ExportNamedDeclaration' || documented?.type === 'ExportDefaultDeclaration') {
+      documented = documented.declaration;
```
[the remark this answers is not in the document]
HAYDEN-OAI: Fixed in `32758cb33571318a043c117451e385b07742328f`. JSDoc host unwrapping is now shared across namespaces, and type references first check the documented callable’s genuine type-parameter binding. Regressions cover exported, named/anonymous default, and exported-arrow generics: TypeScript binds `T` to `TypeParameter`, autofix removes the shadowing runtime import, and the binding remains correct afterward. A separate `typeof T` case proves value-namespace import behavior is preserved. All 39 import/JSDoc groups and the full validation matrix passed.

scripts/oxlint-plugin.cjs
```diff
@@ -0,0 +1,759 @@
...
+    retainedNamedSpecifiers.length === 0 &&
+    unusedSpecifiers.some((specifier) => specifier.type === 'ImportSpecifier')
+  ) {
+    return null;
+  }
+
+  const fixes = [];
+  for (const specifier of unusedSpecifiers) {
+    if (sourceCode.getCommentsInside(specifier).length > 0) return null;
+
```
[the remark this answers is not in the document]
HAYDEN-OAI: Fixed in `32758cb33571318a043c117451e385b07742328f`. Atomic removal now declines whenever comments intersect the specifier/comma removal span or the trailing trivia that would be reassigned. Focused cases cover trailing block and line comments plus a commented default import before the comma; each keeps the diagnostic and preserves the source byte-for-byte, while the existing safe atomic-removal matrix still fixes normally. All 39 import/JSDoc groups and the full validation matrix passed.

scripts/oxlint-plugin.cjs
```diff
@@ -0,0 +1,759 @@
...
+  );
+  const attached = new Map();
+
+  function visit(node) {
+    if (!ts.isEmptyStatement(node)) {
+      for (const comment of node.jsDoc ?? []) {
+        if (
+          node.kind !== ts.SyntaxKind.EndOfFileToken ||
+          comment.tags?.some((tag) => ts.isJSDocTypedefTag(tag))
+        ) {
```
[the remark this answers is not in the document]
HAYDEN-OAI: Fixed in `32758cb33571318a043c117451e385b07742328f`. EOF-hosted declarations now admit TypeScript `JSDocCallbackTag` as well as `JSDocTypedefTag`. The regression proves TypeScript attaches the callback to `EndOfFileToken`, both nested `@param`/`@returns` `Foo` references bind to the import, autofix preserves that import/runtime side effect, and a neighboring orphan `@type {Ref}` still does not retain `Ref`. All 39 import/JSDoc groups and the full validation matrix passed.

scripts/oxlint-plugin.cjs
```diff
@@ -0,0 +1,802 @@
...
+  if (namespace === 'type' && callable) {
+    const typeParameter = sourceCode.getScope(callable).set.get(name);
+    if (isTypeParameterBinding(typeParameter)) return typeParameter;
+  }
+
+  for (let scope = sourceCode.getScope(node ?? sourceCode.ast); scope; scope = scope.upper) {
+    const [start, end] = scope.block.range;
+    if (comment.range[0] < start || comment.range[1] > end) continue;
+    const variable = scope.set.get(name);
+    if (variable && (namespace === 'value' || isTypeBinding(variable))) return variable;
```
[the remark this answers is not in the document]
HAYDEN-OAI: Fixed in 2acc47ce. The rule now consults TypeScript’s bound JSDoc symbols for attached comments and suppresses an outer import when the reference resolves to a standalone local `@typedef`/`@callback`. Added module/function/block, declaration-order, outer-to-inner visibility, and non-leaking inner-scope regressions; the runtime side-effect assertions confirm the stale import is removed while genuine imported references remain. The full 41-group lint gate and strict Node 22/24/26 test matrix pass.

scripts/oxlint-plugin.cjs
```diff
@@ -0,0 +1,802 @@
...
+
+  ts.forEachChild(node, (child) => collectTypeReferences(child, bindings, used));
+}
+
+function collectDocumentationLinks(comment, bindings, used, text) {
+  if (!Array.isArray(comment)) return;
+
+  for (const node of comment) {
+    if (!ts.isJSDocLinkLike(node) || text[node.pos - 1] === '\\') continue;
+    addEntityReference(node.name, bindings, used);
```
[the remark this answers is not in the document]
HAYDEN-OAI: Fixed in 2acc47ce. Documentation links now use the TypeScript checker’s resolved symbol flags instead of assuming the type namespace, including aliases and symbols that inhabit both namespaces. Added parameter, variable, type-alias, class, and true-import regressions with runtime side-effect checks; local value/type targets now release the stale import while real imported links remain. The full 41-group lint gate and strict Node 22/24/26 test matrix pass.

scripts/oxlint-plugin.cjs
```diff
@@ -0,0 +1,1001 @@
...
+  }
+
+  return namespaces.size > 0 ? namespaces : undefined;
+}
+
+function isStandaloneJSDocDeclarationSymbol(symbol) {
+  return symbol?.declarations?.some(isStandaloneJSDocDeclarationTag);
+}
+
+function createJSDocChecker(compilerSource) {
```
jbeckwith-oai: **P2 — decompose the plugin before merging.** This three-commit delta grows this file from 752 to 1,001 lines and simultaneously pushes `scripts/oxlint-regression.test.cjs` from 999 to 1,007, crossing the review's explicit 1k maintainability boundary. The plugin now owns JSDoc parsing, compiler-host/program construction, lexical and namespace resolution, JSX pragma analysis, import-edit safety, and rule registration. There is a clear extraction boundary here: move `createJSDocChecker`, `getJSDocNamespaceInfo`, and the related JSDoc symbol-analysis helpers into a focused module, keep the plugin entrypoint orchestration-sized, and split the regression cases by behavior while requiring each suite from the main test entrypoint. This boundary also makes the new compiler cost visible—the exact repo `oxlint .` run increased from roughly 0.87s on the previous head to 2.8–3.7s here—so it can be measured and cached deliberately instead of remaining embedded in a 1k-line rule file.

jbeckwith-oai (changes requested):
Re-review of exact head `2acc47ce7d9e02565f68f5b92083c1d14dbd9004`. The prior EOF-hosted module-level `@typedef` blocker is fixed: the import survives autofix and TypeScript 5.8 `checkJs` remains clean, including the new mixed-orphan coverage. Two blockers remain: the same-line `@example ... @param {Foo}` sibling-tag bug already captured at https://github.com/openai/openai-node/pull/2044#discussion_r3715355555, plus the structural decomposition issue described inline. Validation: 41/41 lint regression groups passed; `oxlint`, Oxfmt, syntax/diff checks, and exact-head CI are green across build, examples, ecosystem tests, Node 22/24/26, breaking-change detection, and CodeQL.

scripts/oxlint-plugin.cjs
```diff
@@ -0,0 +1,1001 @@
...
+    collectDocumentationLinks(doc.comment, bindings, references, parsed.text, commentDocumentationNamespaces);
+
+    let inExample = false;
+    let documentationTag;
+    let previousTag;
+    for (const originalTag of tags) {
+      const name = ts.unescapeLeadingUnderscores(originalTag.tagName.escapedText).toLowerCase();
+      const startsLine = startsJSDocLine(parsed.text, originalTag.pos);
+
+      if (inExample && !startsLine) continue;
```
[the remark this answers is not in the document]
HAYDEN-OAI: Fixed in 0e5415f6. Same-line `@param`/`@arg`/`@argument` tags after `@example` are now recovered only when TypeScript associates the tag with an actual callable parameter; a TypeScript scanner still rejects markers inside example-code comments. The focused suite covers arrows/functions/methods, aliases, defaults, rest/destructuring, following tags, escaped markers, and example lookalikes, with TypeScript binding/type-validity and runtime import-side-effect assertions. All 46 regression groups and the full Node 22/24/26 matrix pass.

scripts/oxlint-plugin.cjs
HAYDEN-OAI: Addressed in 0e5415f6. JSDoc parsing and symbol analysis now live in focused 686- and 152-line modules; the plugin entrypoint is 248 lines. Regression coverage is split by behavior and required by a 6-line canonical entrypoint, with a shared support module and dedicated lifecycle suite. Analysis is context-local, lazy, memoized once per file, and never reused across files. Controlled warmed `oxlint .` medians on the same devbox/dependency graph were 2.118s at the prior commit, 6.997s at the reviewed head, and 1.206s after this change (5.80× faster than the reviewed head). All 46 groups, whole-tree Oxfmt/Oxlint, build/type/ATTW/publint, packed CJS/ESM, and Node 22/24/26 suites pass.

scripts/oxlint-jsdoc.cjs
```diff
@@ -0,0 +1,686 @@
...
+        const parameter =
+          ts.isJSDocParameterTag(originalTag) &&
+          callables.some((callable) =>
+            callable.parameters.some((candidate) =>
+              ts
+                .getJSDocParameterTags(candidate)
+                .some((tag) => tag.pos === comment.range[0] + originalTag.pos),
+            ),
+          );
+        if (!parameter || isInJSDocExampleComment(parsed.text, exampleTag, originalTag)) continue;
```
jbeckwith-oai, line 582: [P1] Preserve compiler-applied non-parameter siblings after `@example`

This escape hatch only recognizes `JSDocParameterTag`, so a direct sibling such as `/** @example foo @returns {Foo} */` remains in example mode and its type reference is ignored. On this exact head, TypeScript checkJs is clean before autofix; `oxlint --fix` deletes the sole `Foo` import; then TypeScript reports TS2304. The same gap also affects other compiler-applied siblings (including direct `@type`, constrained `@template`, `@implements`, and `@typedef`). Please generalize the compiler-application check beyond parameters while continuing to reject true example-body lookalikes, and add a direct `@returns` before/after regression—the current returns case has a recognized `@param` first, which exits example mode before `@returns` is seen.

jbeckwith-oai (changes requested):
The original same-line `@param` regression is fixed, and the thermo decomposition concern is fully resolved (the plugin and main regression runner are now well under 1,000 lines with cohesive extracted modules). However, the same-line recovery remains parameter-specific and still deletes imports required by other compiler-applied sibling tags after `@example`. I confirmed this independently with a direct `@returns` case, and both parallel review passes reproduced it. Exact-head CI is green and the full focused regression suite passes 46/46, but it does not cover this direct-return shape.

jbeckwith-oai (approved):

PR #2047, opened 2026-08-03
fix(streaming): surface mid-stream errors when iterating buffered streams

### LINKED ISSUE #2046: ChatCompletionStream/ResponseStream: mid-stream errors are silently swallowed when iterating with `for await`

### Describe the bug

When consuming a streaming helper with `for await (const chunk of stream)`, a terminal error that occurs **mid-stream can be silently swallowed** — the loop finishes cleanly as if the response completed successfully, when it actually failed (e.g. the connection dropped).

The `[Symbol.asyncIterator]()` implementations in `ChatCompletionStream` (`src/lib/ChatCompletionStream.ts`) and `ResponseStream` (`src/lib/responses/ResponseStream.ts`) buffer incoming events in an internal `pushQueue` when the producer is ahead of the consumer. The `error`/`abort` handlers only **reject readers that are waiting at that exact moment** (`readQueue`); they never store the error. So if the error fires while events are buffered and no reader is currently waiting, it rejects nobody and just sets `done = true`. The consumer then drains the buffered events and `next()` returns `{ done: true }` without ever throwing — the error is lost.

This happens in the common case where the loop body does any `await` (persisting a chunk, processing, etc.), which lets the network/producer run ahead and fill the buffer.

Notably, the SDK's own `EventStream.events()` (`src/lib/EventStream.ts`) already handles this correctly with `failure`/`failureDelivered` bookkeeping; the two `[Symbol.asyncIterator]` implementations were just never given the same treatment.

### To Reproduce

1. Start a streaming request via `openai.chat.completions.stream(...)` (or `openai.responses.stream(...)`).
2. Iterate it with `for await`, doing some `await` work per chunk so the internal buffer fills.
3. Have the stream fail partway through (e.g. connection drop / an error frame).
4. Observe that the `for await` loop **ends normally and does not throw** — the error is swallowed.

### Code snippets

```JavaScript
const stream = openai.chat.completions.stream({
  model: 'gpt-4o',
  messages: [{ role: 'user', content: 'hi' }],
});

try {
  for await (const chunk of stream) {
    await persist(chunk); // the await lets the internal buffer fill
  }
  console.log('completed'); // ← reached even when the stream errored mid-way
} catch (err) {
  console.error('stream failed', err); // ← never runs, error is lost
}
```

### OS

macOS 26.6

### Node version

Node v22.23.1

### Library version

openai v7.3.0

pouyashahrdami:

## Summary

Fixes #2046.

Streaming consumers could silently treat a failed request as successful when chunks or events were already buffered before a terminal error arrived. Chat Completions, Assistants, and streaming tool-runner readable-stream bridges each retained their own lossy iterator implementation; Responses streaming now also uses the same shared implementation consistently.

Extract the existing retained-failure behavior into a shared, documented `EventStream` iterator. Every handwritten streaming adapter now drains queued values, rejects exactly once with the original terminal error or abort, removes producer and lifecycle listeners, and cancels its underlying request when iteration stops early. Assistant events remain cloned when queued, and tool-runner chunk/tool-message listeners remain eager and ordered.

The implementation preserves current Responses API nested/flat SSE error normalization, provider error metadata, structured-output parsing, and GPT-5 custom-tool support. All modified SDK source lives under the handwritten `src/lib/` tree.

## Regression coverage

- Buffered and pending failures across Chat Completion, Assistant, and Responses iterators.
- Buffered failures crossing `ChatCompletionStreamingRunner.toReadableStream()`.
- Exact-once terminal failure delivery, cancellation, completed-stream iteration, listener cleanup, and cloned Assistant snapshots.
- Explicit iteration over `error` and `abort` events without changing their existing semantics.

## Verification

- Full handwritten Vitest suite: **66 files, 1,727 tests passed**.
- Repository lint and formatting checks passed.
- Exact repository breaking-change workflow passed against current `main`.
- Package build and CommonJS/ESM smoke checks passed.
- Current TypeScript and TypeScript 4.9 both successfully type-check the published package.
- Generated-file lint and Node version policy checks passed.


jbeckwith-oai (dismissed):
Blocking: the same buffered-terminal-error loss remains in two public handwritten streaming paths. ChatCompletionStreamingRunner.toReadableStream() overrides the fixed base conversion with its own pushQueue/readQueue/error handlers, and AssistantStream[Symbol.asyncIterator]() has the same old pattern. I reproduced both on this exact head: enqueue two items with no reader waiting, emit OpenAIError('boom'), then consume; each returned both items and ended normally with thrown=null. This means tool-runner readable-stream bridges and Assistants iteration can still report a failed response as successful. Please apply the retained-failure semantics (preferably via a shared/EventStream.events-based adapter so these copies cannot drift), including listener/queue cleanup on return/end, to every buffered adapter, and add regressions for these two paths. The two paths changed here otherwise behave correctly: their 17 targeted tests, TypeScript, ESLint, Prettier, build, and diff checks pass. Please also make the new tests deterministic by waiting for the stream's terminal error/done signal instead of a fixed 10 ms sleep, and assert the exact error/message rather than any Error.

pouyashahrdami:
@jbeckwith-oai Thanks for the thorough review — all points addressed in 5f986015.

**Shared adapter:** Went with your suggestion: extracted the `events()` logic into a shared `EventStream#_createIterator` and rebuilt every buffered adapter on it — both `Symbol.asyncIterator` implementations, `AssistantStream`'s iterator, and `ChatCompletionStreamingRunner.toReadableStream()`. `events()` itself now delegates to it too, so there's exactly one copy of the queue/retained-failure logic left. Per-path semantics are preserved: break-out-of-loop still aborts, the runner bridge still registers listeners eagerly (pre-pull message deltas aren't dropped), and `AssistantStream` still clones at push time.

**Cleanup:** Listeners are removed and queues drained on end/return. This surfaced one subtlety worth flagging: the old leaked listeners incidentally suppressed the intentional unhandled-rejection warning when a consumer breaks out of iteration (the self-triggered abort found no listeners after cleanup). `return()` now marks the terminal promise handled before invoking the abort hook, so existing consumer behavior is unchanged; `events()` doesn't use that hook, so its "don't suppress after cleanup" contract still holds (covered by the existing test).

**Tests:** Added regressions for both paths you flagged (both fail on the previous head), plus edge cases: abort retention through a full buffer, exactly-once failure delivery with a pending reader, break-aborts, and post-end iteration returning done instead of hanging. The buffered-error tests now await the stream's terminal `done()` signal instead of sleeping, and assert the exact error class and message.

Full suite, type check, lint, format, and build are green locally. Ready for another look.

HAYDEN-OAI (approved):
Verified against actual buffered provider failures across Chat Completions, Assistants, Responses, and tool-runner readable streams. Shared iterator preserves exact-once errors, cancellation, listener cleanup, Assistant snapshots, and existing SSE/custom-tool fixes. Full handwritten suite: 1,727 passing; lint, breaking-change check, package build, and TypeScript 4.9/current checks all pass.

pouyashahrdami:
Appreciate your feedback!

PR #2048, opened 2026-08-03
build: migrate release workflow to upstream release-please

jbeckwith-oai:
## Summary

Migrate the release workflow from `stainless-api/trigger-release-please` to the standard `googleapis/release-please-action`.

## What changed

- switched `.github/workflows/create-releases.yml` to manifest-mode `googleapis/release-please-action`
- pinned the action to the latest upstream commit SHA verified during the migration
- split the workflow into separate `release` and `publish` jobs so broader release-please permissions are not granted during publish
- updated the publish job guard to run when `release_created` is true even if the release job later reports failure
- configured release-please to create draft release PRs, allowing `ready_for_review` to trigger normal CI when using `GITHUB_TOKEN`
- updated `release-please-config.json` to point at the upstream `googleapis/release-please` schema

## Why

- removes the dependency on the Stainless-specific wrapper action
- aligns this repo with the standard upstream release-please setup
- preserves the existing release-PR model while tightening permission scope in the publish path
- avoids a partial-failure case where a GitHub release could be created but npm publish would be skipped
- preserves CI coverage for release PRs created with the repository `GITHUB_TOKEN`

## Validation

- `pnpm lint`
- YAML and JSON parse validation
- release-please config schema validation
- `git diff --check`

## Notes

- the workflow uses `github.token` for release-please, not a PAT or GitHub App token
- supersedes #1875 so the branch lives directly in `openai/openai-node`

### DIFF
--- a/.github/workflows/create-releases.yml
+++ b/.github/workflows/create-releases.yml
@@ -6,6 +6,13 @@ on:
     branches:
       - main
       - next
+  pull_request:
+    types:
+      - edited
+      - opened
+      - ready_for_review
+      - reopened
+      - synchronize
 
 jobs:
   promote:
@@ -25,16 +32,50 @@ jobs:
         env:
           GH_TOKEN: ${{ github.token }}
         run: |
-          ahead_by="$(gh api "repos/${GITHUB_REPOSITORY}/compare/main...next" --jq '.ahead_by')"
-          if [[ "${ahead_by}" == '0' ]]; then
+          if git diff --quiet origin/main HEAD --; then
             echo 'No Stainless changes to promote.'
             exit 0
           fi
 
+          last_promoted_sha="$(
+            gh pr list \
+              --repo "${GITHUB_REPOSITORY}" \
+              --state merged \
+              --base main \
+              --head next \
+              --limit 100 \
+              --json headRefOid,mergedAt \
+              --jq 'sort_by(.mergedAt) | last | .headRefOid // empty'
+          )"
+
+          if [[ -n "${last_promoted_sha}" ]]; then
+            if ! git cat-file -e "${last_promoted_sha}^{commit}" 2>/dev/null; then
+              git fetch --no-tags origin "${last_promoted_sha}"
+            fi
+            if ! git merge-base --is-ancestor "${last_promoted_sha}" HEAD; then
+              echo "Last promoted commit ${last_promoted_sha} is not an ancestor of next."
+              echo 'Refusing to classify a rewritten promotion range.'
+              exit 1
+            fi
+            promotion_range="${last_promoted_sha}..HEAD"
+          else
+            promotion_range='origin/main..HEAD'
+          fi
+
+          commit_subjects="$(git log "${promotion_range}" --format='%s')"
+          commit_messages="$(git log "${promotion_range}" --format='%s%n%b')"
+          conventional_commit='^(feat|fix|perf|revert|chore|docs|style|refactor|test|build|ci)(\([^)]*\))?!?:'
+          if ! grep -Eq "${conventional_commit}" <<< "${commit_subjects}" && \
+            ! grep -Eq '^BREAKING[ -]CHANGE:' <<< "${commit_messages}"; then
+            echo "No releasable commits in ${promotion_range}; skipping promotion."
+            exit 0
+          fi
+
           promotion_type='fix(api)'
-          if git log origin/main..HEAD --format='%s%n%b' | grep -Eq '(^[[:alpha:]]+(\([^)]*\))?!:)|(^BREAKING[ -]CHANGE:)'; then
+          if grep -Eq '^[[:alpha:]]+(\([^)]*\))?!:' <<< "${commit_subjects}" || \
+            grep -Eq '^BREAKING[ -]CHANGE:' <<< "${commit_messages}"; then
             promotion_type='feat(api)!'
-          elif git log origin/main..HEAD --format='%s' | grep -Eq '^feat(\([^)]*\))?:'; then
+          elif grep -Eq '^feat(\([^)]*\))?:' <<< "${commit_subjects}"; then
             promotion_type='feat(api)'
           fi
           promotion_title="${promotion_type}: promote Stainless-generated changes"
@@ -70,6 +111,36 @@ jobs:
             gh pr ready "${promotion_pr}" --repo "${GITHUB_REPOSITORY}" --undo
           fi
 
+  validate-release-version:
+    name: validate release PR version
+    if: >-
+      ${{
+        github.event_name == 'pull_request' &&
+        github.event.pull_request.head.ref == 'release-please--branches--main--components--openai' &&
+        github.repository == 'openai/openai-node'
+      }}
+    runs-on: ubuntu-latest
+    permissions:
+      contents: read
+
+    steps:
+      - uses: actions/checkout@de0fac2e4500dabe0009e67214ff5f5447ce83dd # v6
+        with:
+          ref: ${{ github.event.pull_request.head.sha }}
+
+      - name: Compare the release title with the generated version
+        env:
+          PR_TITLE: ${{ github.event.pull_request.title }}
+        run: |
+          package_version="$(jq -r -e '.version' package.json)"
+          expected_title="release: ${package_version}"
+          if [[ "${PR_TITLE}" != "${expected_title}" ]]; then
+            echo "Release PR title '${PR_TITLE}' does not match package.json version ${package_version}."
+            echo 'Do not edit the title to override the release version.'
+            echo 'Add a conventional commit with a Release-As: <version> footer instead.'
+            exit 1
+          fi
+
   release:
     name: release
     if: github.ref == 'refs/heads/main' && github.repository == 'openai/openai-node'
@@ -78,9 +149,6 @@ jobs:
       contents: write
       issues: write
       pull-requests: write
-    outputs:
-      release_created: ${{ steps.release.outputs.release_created }}
-
     steps:
       - name: Check for the legacy Stainless release PR
         id: cutover
@@ -120,15 +188,66 @@ jobs:
             gh pr ready "${RELEASE_PR}" --repo "${GITHUB_REPOSITORY}" --undo
           fi
 
-  publish:
-    name: publish
+  publication-check:
+    name: check publication state
     if: >-
       ${{
         always() &&
         !cancelled() &&
-        needs.release.outputs.release_created == 'true'
+        github.event_name == 'push' &&
+        github.ref == 'refs/heads/main' &&
+        github.repository == 'openai/openai-node'
       }}
     needs: release
+    runs-on: ubuntu-latest
+    permissions:
+      contents: read
+    outputs:
+      should_publish: ${{ steps.check.outputs.should_publish }}
+
+    steps:
+      - uses: actions/checkout@de0fac2e4500dabe0009e67214ff5f5447ce83dd # v6
+
+      - name: Check whether the current release needs publishing
+        id: check
+        env:
+          GH_TOKEN: ${{ github.token }}
+        run: |
+          package_version="$(jq -r -e '.version' package.json)"
+          release_tag="v${package_version}"
+
+          set +e
+          release_info="$(gh api --silent "repos/${GITHUB_REPOSITORY}/releases/tags/${release_tag}" 2>&1)"
+          release_status=$?
+          set -e
+          if [[ ${release_status} -ne 0 ]]; then
+            if grep -q 'HTTP 404' <<< "${release_info}"; then
+              echo "No GitHub release exists for ${release_tag}; nothing to publish."
+              echo 'should_publish=false' >> "${GITHUB_OUTPUT}"
+              exit 0
+            fi
+            echo "Unable to check for GitHub release ${release_tag}:"
+            echo "${release_info}"
+            exit 1
+          fi
+
+          set +e
+          bash ./bin/check-npm-version
+          version_status=$?
+          set -e
+
+          if [[ ${version_status} -eq 0 ]]; then
+            echo 'should_publish=false' >> "${GITHUB_OUTPUT}"
+          elif [[ ${version_status} -eq 1 ]]; then
+            echo 'should_publish=true' >> "${GITHUB_OUTPUT}"
+          else
+            exit "${version_status}"
+          fi
+
+  publish:
+    name: publish
+    if: needs.publication-check.outputs.should_publish == 'true'
+    needs: publication-check
     runs-on: ubuntu-latest
     environment: publish
     permissions:
--- a/.github/workflows/create-releases.yml
+++ b/.github/workflows/create-releases.yml
@@ -35,7 +35,7 @@ jobs:
         env:
           GH_TOKEN: ${{ github.token }}
         run: |
-          legacy_pr="$(bash ./bin/find-legacy-release-pr)"
+          legacy_pr="$(bash ./bin/find-legacy-release-pr open-number)"
           if [[ -n "${legacy_pr}" ]]; then
             echo "Waiting for legacy release PR #${legacy_pr} before promoting next directly."
             exit 0
@@ -46,7 +46,7 @@ jobs:
             exit 0
           fi
 
-          last_promoted_sha="$(
+          promotion_start="$(
             gh pr list \
               --repo "${GITHUB_REPOSITORY}" \
               --state merged \
@@ -57,19 +57,23 @@ jobs:
               --jq 'sort_by(.mergedAt) | last | .headRefOid // empty'
           )"
 
-          if [[ -n "${last_promoted_sha}" ]]; then
-            if ! git cat-file -e "${last_promoted_sha}^{commit}" 2>/dev/null; then
-              git fetch --no-tags origin "${last_promoted_sha}"
+          if [[ -z "${promotion_start}" ]]; then
+            promotion_start="$(bash ./bin/find-legacy-release-pr merged-next-sha)"
+          fi
+
+          if [[ -n "${promotion_start}" ]]; then
+            if ! git cat-file -e "${promotion_start}^{commit}" 2>/dev/null; then
+              git fetch --no-tags origin "${promotion_start}"
             fi
-            if ! git merge-base --is-ancestor "${last_promoted_sha}" HEAD; then
-              echo "Last promoted commit ${last_promoted_sha} is not an ancestor of next."
+            if ! git merge-base --is-ancestor "${promotion_start}" HEAD; then
+              echo "Promotion boundary ${promotion_start} is not an ancestor of next."
               echo 'Refusing to classify a rewritten promotion range.'
               exit 1
             fi
-            promotion_range="${last_promoted_sha}..HEAD"
           else
-            promotion_range='origin/main..HEAD'
+            promotion_start='origin/main'
           fi
+          promotion_range="${promotion_start}..HEAD"
 
           commit_subjects="$(git log "${promotion_range}" --not origin/main --format='%s')"
           commit_messages="$(git log "${promotion_range}" --not origin/main --format='%s%n%b')"
@@ -183,7 +187,7 @@ jobs:
         env:
           GH_TOKEN: ${{ github.token }}
         run: |
-          legacy_pr="$(bash ./bin/find-legacy-release-pr)"
+          legacy_pr="$(bash ./bin/find-legacy-release-pr open-number)"
           echo "legacy_pr=${legacy_pr}" >> "${GITHUB_OUTPUT}"
           if [[ -n "${legacy_pr}" ]]; then
             echo "Waiting for legacy release PR #${legacy_pr} to merge or close before cutover."
--- a/.github/workflows/create-releases.yml
+++ b/.github/workflows/create-releases.yml
@@ -11,6 +11,33 @@ jobs:
     name: release
     if: github.ref == 'refs/heads/main' && github.repository == 'openai/openai-node'
     runs-on: ubuntu-latest
+    permissions:
+      contents: write
+      issues: write
+      pull-requests: write
+    outputs:
+      release_created: ${{ steps.release.outputs.release_created }}
+
+    steps:
+      - uses: googleapis/release-please-action@45996ed1f6d02564a971a2fa1b5860e934307cf7 # v5.0.0
+        id: release
+        with:
+          # Use the default token until the openai-sdks App credentials are provisioned.
+          token: ${{ github.token }}
+          config-file: release-please-config.json
+          manifest-file: .release-please-manifest.json
+          target-branch: main
+
+  publish:
+    name: publish
+    if: >-
+      ${{
+        always() &&
+        !cancelled() &&
+        needs.release.outputs.release_created == 'true'
+      }}
+    needs: release
+    runs-on: ubuntu-latest
     environment: publish
     permissions:
       contents: read
@@ -19,29 +46,18 @@ jobs:
     steps:
       - uses: actions/checkout@de0fac2e4500dabe0009e67214ff5f5447ce83dd # v6
 
-      - uses: stainless-api/trigger-release-please@bb6677c5a04578eec1ccfd9e1913b5b78ed64c61 # v1.4.0
-        id: release
-        with:
-          repo: ${{ github.event.repository.full_name }}
-          stainless-api-key: ${{ secrets.STAINLESS_API_KEY }}
-
       - name: Set up pnpm
-        if: ${{ steps.release.outputs.releases_created }}
         uses: pnpm/action-setup@f40ffcd9367d9f12939873eb1018b921a783ffaa # v4
-
       - name: Set up Node
-        if: ${{ steps.release.outputs.releases_created }}
         uses: actions/setup-node@3235b876344d2a9aa001b8d1453c930bba69e610 # v3
         with:
           node-version-file: '.nvmrc'
           cache: pnpm
 
       - name: Install dependencies
-        if: ${{ steps.release.outputs.releases_created }}
         run: |
           pnpm install --frozen-lockfile
 
       - name: Publish to NPM
-        if: ${{ steps.release.outputs.releases_created }}
         run: |
           bash ./bin/publish-npm
--- a/.github/workflows/create-releases.yml
+++ b/.github/workflows/create-releases.yml
@@ -35,6 +35,12 @@ jobs:
         env:
           GH_TOKEN: ${{ github.token }}
         run: |
+          legacy_pr="$(bash ./bin/find-legacy-release-pr)"
+          if [[ -n "${legacy_pr}" ]]; then
+            echo "Waiting for legacy release PR #${legacy_pr} before promoting next directly."
+            exit 0
+          fi
+
           if git diff --quiet origin/main HEAD --; then
             echo 'No Stainless changes to promote.'
             exit 0
@@ -65,12 +71,12 @@ jobs:
             promotion_range='origin/main..HEAD'
           fi
 
-          commit_subjects="$(git log "${promotion_range}" --format='%s')"
-          commit_messages="$(git log "${promotion_range}" --format='%s%n%b')"
+          commit_subjects="$(git log "${promotion_range}" --not origin/main --format='%s')"
+          commit_messages="$(git log "${promotion_range}" --not origin/main --format='%s%n%b')"
           conventional_commit='^(feat|fix|perf|revert|chore|docs|style|refactor|test|build|ci)(\([^)]*\))?!?:'
           if ! grep -Eq "${conventional_commit}" <<< "${commit_subjects}" && \
             ! grep -Eq '^BREAKING[ -]CHANGE:' <<< "${commit_messages}"; then
-            echo "No releasable commits in ${promotion_range}; skipping promotion."
+            echo "No unreleased conventional commits in ${promotion_range}; skipping promotion."
             exit 0
           fi
 
@@ -177,14 +183,7 @@ jobs:
         env:
           GH_TOKEN: ${{ github.token }}
         run: |
-          legacy_pr="$(
-            gh pr list \
-              --repo "${GITHUB_REPOSITORY}" \
-              --state open \
-              --head 'release-please--branches--main--changes--next--components--openai' \
-              --json number \
-              --jq '.[0].number // empty'
-          )"
+          legacy_pr="$(bash ./bin/find-legacy-release-pr)"
           echo "legacy_pr=${legacy_pr}" >> "${GITHUB_OUTPUT}"
           if [[ -n "${legacy_pr}" ]]; then
             echo "Waiting for legacy release PR #${legacy_pr} to merge or close before cutover."
--- a/.github/workflows/create-releases.yml
+++ b/.github/workflows/create-releases.yml
@@ -9,7 +9,7 @@ on:
     workflows:
       - CI
     types:
-      - completed
+      - requested
     branches:
       - next
   pull_request:
--- a/.github/workflows/create-releases.yml
+++ b/.github/workflows/create-releases.yml
@@ -18,6 +18,9 @@ jobs:
   promote:
     name: promote Stainless changes
     if: github.ref == 'refs/heads/next' && github.repository == 'openai/openai-node'
+    concurrency:
+      group: promote-stainless-${{ github.ref }}
+      cancel-in-progress: true
     runs-on: ubuntu-latest
     permissions:
       contents: read
@@ -80,6 +83,12 @@ jobs:
           fi
           promotion_title="${promotion_type}: promote Stainless-generated changes"
 
+          current_next_sha="$(gh api "repos/${GITHUB_REPOSITORY}/git/ref/heads/next" --jq '.object.sha')"
+          if [[ "${current_next_sha}" != "${GITHUB_SHA}" ]]; then
+            echo "A newer next commit (${current_next_sha}) superseded this run (${GITHUB_SHA})."
+            exit 0
+          fi
+
           promotion_pr="$(
             gh pr list \
               --repo "${GITHUB_REPOSITORY}" \
@@ -132,14 +141,7 @@ jobs:
         env:
           PR_TITLE: ${{ github.event.pull_request.title }}
         run: |
-          package_version="$(jq -r -e '.version' package.json)"
-          expected_title="release: ${package_version}"
-          if [[ "${PR_TITLE}" != "${expected_title}" ]]; then
-            echo "Release PR title '${PR_TITLE}' does not match package.json version ${package_version}."
-            echo 'Do not edit the title to override the release version.'
-            echo 'Add a conventional commit with a Release-As: <version> footer instead.'
-            exit 1
-          fi
+          bash ./bin/check-release-version "${PR_TITLE}"
 
   release:
     name: release
@@ -150,6 +152,26 @@ jobs:
       issues: write
       pull-requests: write
     steps:
+      - uses: actions/checkout@de0fac2e4500dabe0009e67214ff5f5447ce83dd # v6
+
+      - name: Validate a merged release PR version
+        env:
+          GH_TOKEN: ${{ github.token }}
+        run: |
+          release_pr_title="$(
+            gh api \
+              -H 'Accept: application/vnd.github+json' \
+              "repos/${GITHUB_REPOSITORY}/commits/${GITHUB_SHA}/pulls" \
+              --jq '[.[] | select(
+                .merged_at != null and
+                .base.ref == "main" and
+                .head.ref == "release-please--branches--main--components--openai"
+              )][0].title // empty'
+          )"
+          if [[ -n "${release_pr_title}" ]]; then
+            bash ./bin/check-release-version "${release_pr_title}"
+          fi
+
       - name: Check for the legacy Stainless release PR
         id: cutover
         env:
@@ -204,6 +226,7 @@ jobs:
       contents: read
     outputs:
       should_publish: ${{ steps.check.outputs.should_publish }}
+      release_tag: ${{ steps.check.outputs.release_tag }}
 
     steps:
       - uses: actions/checkout@de0fac2e4500dabe0009e67214ff5f5447ce83dd # v6
@@ -230,6 +253,7 @@ jobs:
             echo "${release_info}"
             exit 1
           fi
+          echo "release_tag=${release_tag}" >> "${GITHUB_OUTPUT}"
 
           set +e
           bash ./bin/check-npm-version
@@ -256,6 +280,8 @@ jobs:
 
     steps:
       - uses: actions/checkout@de0fac2e4500dabe0009e67214ff5f5447ce83dd # v6
+        with:
+          ref: ${{ needs.publication-check.outputs.release_tag }}
 
       - name: Set up pnpm
         uses: pnpm/action-setup@f40ffcd9367d9f12939873eb1018b921a783ffaa # v4
--- a/.github/workflows/create-releases.yml
+++ b/.github/workflows/create-releases.yml
@@ -10,6 +10,7 @@ on:
       - CI
     types:
       - requested
+      - completed
     branches:
       - next
   pull_request:
@@ -22,6 +23,44 @@ on:
       - synchronize
 
 jobs:
+  redraft-promotion:
+    name: re-draft promotion PR
+    if: >-
+      ${{
+        github.repository == 'openai/openai-node' &&
+        github.event_name == 'workflow_run' &&
+        github.event.action == 'requested' &&
+        github.event.workflow_run.event == 'push'
+      }}
+    concurrency:
+      group: promote-stainless-next
+      cancel-in-progress: true
+    runs-on: ubuntu-latest
+    permissions:
+      pull-requests: write
+
+    steps:
+      - name: Re-draft the existing promotion PR
+        env:
+          GH_TOKEN: ${{ github.token }}
+        run: |
+          promotion_pr="$(
+            gh pr list \
+              --repo "${GITHUB_REPOSITORY}" \
+              --state open \
+              --base main \
+              --head next \
+              --json isDraft,number \
+              --jq '.[0] // empty'
+          )"
+          if [[ -z "${promotion_pr}" || "$(jq -r '.isDraft' <<< "${promotion_pr}")" == 'true' ]]; then
+            exit 0
+          fi
+          gh pr ready \
+            "$(jq -r -e '.number' <<< "${promotion_pr}")" \
+            --repo "${GITHUB_REPOSITORY}" \
+            --undo
+
   promote:
     name: promote Stainless changes
     if: >-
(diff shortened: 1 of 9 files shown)


HAYDEN-OAI (commented):
Reviewed exact commit `913be722a10eaaf97359411e896ecf3ec971f881`. Existing threads already identify the loss of Stainless generation/promotion and the missing CI reruns after later bot-authored release-PR updates; the three inline comments below identify additional independent rollout blockers.

One existing automated finding is incorrect: the pinned action bundles release-please 17.6.0, whose [upstream versioning-strategy factory explicitly registers `prerelease`](https://github.com/googleapis/release-please/blob/v17.6.0/src/factories/versioning-strategy-factory.ts#L40-L46). The existing `"versioning": "prerelease"` setting is supported and should not be removed on that basis.

.github/workflows/create-releases.yml
```diff
@@ -11,6 +11,33 @@ jobs:
...
+      pull-requests: write
+    outputs:
+      release_created: ${{ steps.release.outputs.release_created }}
+
+    steps:
+      - uses: googleapis/release-please-action@45996ed1f6d02564a971a2fa1b5860e934307cf7 # v5.0.0
+        id: release
+        with:
+          # Use the default token until the openai-sdks App credentials are provisioned.
+          token: ${{ github.token }}
```
HAYDEN-OAI, line 287: **[P1] The repository currently forbids `GITHUB_TOKEN` from creating release PRs.** `GET /repos/openai/openai-node/actions/permissions/workflow` returns `can_approve_pull_request_reviews: false`. GitHub documents that this repository setting controls whether `GITHUB_TOKEN` can **create** pull requests; adding `pull-requests: write` to the job cannot override it. The first attempted draft release PR will therefore fail instead of replacing the existing Stainless automation. Enable “Allow GitHub Actions to create and approve pull requests” if policy permits, or use an independently authorized GitHub App installation token.

.github/workflows/create-releases.yml
```diff
@@ -11,6 +11,33 @@ jobs:
     name: release
     if: github.ref == 'refs/heads/main' && github.repository == 'openai/openai-node'
     runs-on: ubuntu-latest
+    permissions:
+      contents: write
```
HAYDEN-OAI, line 259: **[P1] The release-branch ruleset blocks the newly selected actor.** Active repository ruleset `15951649` (`stainless release-please branches`) matches `refs/heads/release-please--*` and restricts creation, updates, and deletion. Its only integration bypass is the Stainless App (`378072`); `GITHUB_TOKEN` belongs to GitHub Actions (`15368`), and the proposed `openai-sdks` App (`3705508`) is not a bypass actor either. Consequently `release-please--branches--main--components--openai` cannot be created or updated even if Actions PR creation is enabled. Migrate the ruleset/bypass together with this action or authenticate as an actor already permitted to manage these protected branches.

.github/workflows/create-releases.yml
```diff
@@ -11,6 +11,33 @@ jobs:
...
+      issues: write
+      pull-requests: write
+    outputs:
+      release_created: ${{ steps.release.outputs.release_created }}
+
+    steps:
+      - uses: googleapis/release-please-action@45996ed1f6d02564a971a2fa1b5860e934307cf7 # v5.0.0
```
HAYDEN-OAI, line 282: **[P1] Cut over the already-pending Stainless release PR before switching generators.** Release PR #2042 is currently open with `autorelease: pending` on `release-please--branches--main--changes--next--components--openai`. The pinned upstream release-please calculates `release-please--branches--main--components--openai` for this root Node package and adopts an existing release PR only when its head branch matches exactly. Once authentication and branch permissions are fixed, the first run will leave #2042 unmanaged and create a second competing release PR. Merge/close/drain #2042, or explicitly migrate its existing release branch, before enabling the new workflow.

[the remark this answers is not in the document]
jbeckwith-oai: Fixed in 5bb34e5b. The workflow now also runs on pushes to the Stainless-managed next branch and opens or re-drafts a promotion PR from next to main. Generated SDK changes therefore reach main through review before upstream release-please versions those commits and creates the release PR.

.github/workflows/create-releases.yml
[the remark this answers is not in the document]
jbeckwith-oai: Fixed in 5bb34e5b. Whenever release-please reports that a PR was created or updated, the workflow now checks the returned root PR and converts it back to draft if it was already marked ready. Marking the refreshed PR ready again emits ready_for_review and runs CI on the latest head while we continue using GITHUB_TOKEN.

release-please-config.json
```diff
@@ -2,13 +2,14 @@
   "packages": {
     ".": {}
   },
-  "$schema": "https://raw.githubusercontent.com/stainless-api/release-please/main/schemas/config.json",
+  "$schema": "https://raw.githubusercontent.com/googleapis/release-please/main/schemas/config.json",
```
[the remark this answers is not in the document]
jbeckwith-oai: Addressed by verification; no config change is required. The pinned release-please-action v5 commit bundles upstream release-please 17.6.0, and that version registers a native prerelease versioning strategy in its versioning-strategy factory. Keeping the existing setting preserves the intended prerelease semantics.

.github/workflows/create-releases.yml
jbeckwith-oai: Fixed in the repository settings. Allow GitHub Actions to create and approve pull requests is now enabled; the workflow permissions endpoint confirms can_approve_pull_request_reviews: true, so the job-level pull-requests: write permission can create the draft PRs.

.github/workflows/create-releases.yml
jbeckwith-oai: Fixed in ruleset 15951649. The Stainless-only ruleset now excludes exactly refs/heads/release-please--branches--main--components--openai, while continuing to protect the legacy release-please--* branches. GitHub does not allow its built-in Actions integration to be added as an organization-owned bypass actor, so the exact-ref exclusion is the narrowest GITHUB_TOKEN-compatible migration.

.github/workflows/create-releases.yml
jbeckwith-oai: Fixed in ebf804e8. The release job now detects an open PR on the legacy Stainless release branch and skips upstream release-please while it exists. PR #2042 therefore remains the sole release PR; merging it produces the main push that drains the legacy flow and enables the upstream action.

.github/workflows/create-releases.yml
```diff
@@ -1,16 +1,98 @@
...
+              --jq '.[0].number // empty'
+          )"
+
+          if [[ -z "${promotion_pr}" ]]; then
+            gh pr create \
+              --repo "${GITHUB_REPOSITORY}" \
+              --base main \
+              --head next \
+              --draft \
+              --title 'feat(api): promote Stainless-generated changes' \
```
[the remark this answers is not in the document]
jbeckwith-oai: Fixed in ebf804e8. The promotion job now checks out full history and derives its PR title from the highest severity in main..next: breaking commits use feat(api)!; feature commits use feat(api); otherwise it uses fix(api). Existing promotion PR titles are updated on each next push before the PR is re-drafted.

.github/workflows/create-releases.yml
```diff
@@ -1,16 +1,135 @@
...
+              --jq '.[0].number // empty'
+          )"
+          echo "legacy_pr=${legacy_pr}" >> "${GITHUB_OUTPUT}"
+          if [[ -n "${legacy_pr}" ]]; then
+            echo "Waiting for legacy release PR #${legacy_pr} to merge or close before cutover."
+          fi
+
+      - uses: googleapis/release-please-action@45996ed1f6d02564a971a2fa1b5860e934307cf7 # v5.0.0
+        id: release
+        if: steps.cutover.outputs.legacy_pr == ''
```
HAYDEN-OAI, line 284: **[P1] Disable the autonomous Stainless release controller before cutting over.** This guard only waits for the *currently open* legacy PR, but deleting the wrapper action does not disable the installed Stainless App. Stainless documents that every `next` change automatically creates or updates its release PR, and this repository demonstrates it: [PR #2030](https://github.com/openai/openai-node/pull/2030) was created at 17:10:01 immediately after a 17:09:42 `next` push, while the first `create-releases.yml` run was not until 17:30:26. Thus closing #2042 is not a lasting cutover: future generated pushes recreate the legacy PR, causing this condition to suppress upstream indefinitely or leaving two competing release controllers/versions. Disable or migrate Stainless release-PR management itself while preserving code generation before enabling upstream release-please.

.github/workflows/create-releases.yml
```diff
@@ -1,16 +1,135 @@
...
+              --head 'release-please--branches--main--changes--next--components--openai' \
+              --json number \
+              --jq '.[0].number // empty'
+          )"
+          echo "legacy_pr=${legacy_pr}" >> "${GITHUB_OUTPUT}"
+          if [[ -n "${legacy_pr}" ]]; then
+            echo "Waiting for legacy release PR #${legacy_pr} to merge or close before cutover."
+          fi
+
+      - uses: googleapis/release-please-action@45996ed1f6d02564a971a2fa1b5860e934307cf7 # v5.0.0
```
HAYDEN-OAI, line 282: **[P1] Preserve manual version overrides without desynchronizing the tag and package.** Maintainers currently override a release by editing its PR title; [PR #1942](https://github.com/openai/openai-node/pull/1942) was explicitly renamed from `release: 6.42.1` to `release: 6.43.0`, after which Stainless regenerated the package/changelog, and seven historical PRs carry `autorelease: custom version`. Upstream release-please does not regenerate files on a title edit, but its release strategy takes the Git tag version directly from that edited single-package PR title. Repeating the established override therefore creates a GitHub tag/release for the requested version while npm still publishes the old `package.json` version. Preserve title-driven regeneration, reject title/package version mismatches, or replace and document the workflow with an explicit `Release-As` commit before migrating.

.github/workflows/create-releases.yml
```diff
@@ -1,16 +1,135 @@
...
+        env:
+          GH_TOKEN: ${{ github.token }}
+        run: |
+          ahead_by="$(gh api "repos/${GITHUB_REPOSITORY}/compare/main...next" --jq '.ahead_by')"
+          if [[ "${ahead_by}" == '0' ]]; then
+            echo 'No Stainless changes to promote.'
+            exit 0
+          fi
+
+          promotion_type='fix(api)'
```
HAYDEN-OAI: **[P2] Do not turn metadata-only `next` updates into SDK patch releases.** The new promotion code defaults every non-feature/non-breaking commit to `fix(api)` whenever `next` is ahead. Real Stainless pushes include standalone [`codegen metadata` commits](https://github.com/openai/openai-node/commit/64f107f2f31acfe1d09ef5cb76148162ce153b1f) changing only `.stats.yml`; normal CI already explicitly special-cases those commits, and upstream release-please would ignore their non-conventional subject. This workflow instead creates a `fix(api): ...` promotion commit, which release-please treats as a real patch and publishes an SDK version without SDK changes. Skip metadata-only/non-releasable ranges rather than inventing a `fix` classification.

.github/workflows/create-releases.yml
```diff
@@ -1,16 +1,135 @@
...
+            gh pr ready "${RELEASE_PR}" --repo "${GITHUB_REPOSITORY}" --undo
+          fi
+
+  publish:
+    name: publish
+    if: >-
+      ${{
+        always() &&
+        !cancelled() &&
+        needs.release.outputs.release_created == 'true'
```
HAYDEN-OAI: **[P2] Recover npm publication when release creation fails after its tag is written.** The pinned action calls `outputReleases(await manifest.createReleases())`, so `release_created` is emitted only after all release bookkeeping returns. Upstream first creates the GitHub release/tag, then comments on the PR and updates its labels; a transient failure in either follow-up leaves an existing release but no output, making this condition skip npm publication. A retry then encounters the existing tag as a duplicate and still cannot satisfy the guard. Detect the existing unpublished package version independently in an `always()` recovery path and make publishing idempotent.

HAYDEN-OAI (commented):
Reviewed exact current commit `ebf804e840ad26741f34214441951276c18c1381`. Prior review [#4846529834](https://github.com/openai/openai-node/pull/2048#pullrequestreview-4846529834) is not authoritative for this revision: its three findings were addressed by the updated repository permissions, release-branch ruleset, and initial legacy-PR guard. The inline comments below identify separate current-head release-lifecycle defects. The supported upstream `prerelease` strategy, draft-to-`ready_for_review` CI path, protected publishing environment, and OIDC permissions were independently validated; failing checks/tests are not findings.

.github/workflows/create-releases.yml
jbeckwith-oai: Fixed in `009ddb8a`. Promotion now requires a configured conventional-commit subject or a breaking-change footer in the range since the last promotion. A standalone non-conventional `codegen metadata` range is skipped, and its changes remain available to ride with the next releasable generated update.

.github/workflows/create-releases.yml
jbeckwith-oai: Fixed. I removed the Stainless App integration (ID 378072) from the bypass actors on the active `stainless release-please branches` ruleset (ID 15951649). Stainless can no longer create or update its autonomous `release-please--*` branches, while the ruleset does not cover `next`, so Stainless code generation remains enabled. The existing legacy PR guard is now only a one-time drain gate.

.github/workflows/create-releases.yml
```diff
@@ -1,16 +1,135 @@
...
+          GH_TOKEN: ${{ github.token }}
+        run: |
+          ahead_by="$(gh api "repos/${GITHUB_REPOSITORY}/compare/main...next" --jq '.ahead_by')"
+          if [[ "${ahead_by}" == '0' ]]; then
+            echo 'No Stainless changes to promote.'
+            exit 0
+          fi
+
+          promotion_type='fix(api)'
+          if git log origin/main..HEAD --format='%s%n%b' | grep -Eq '(^[[:alpha:]]+(\([^)]*\))?!:)|(^BREAKING[ -]CHANGE:)'; then
```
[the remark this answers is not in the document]
jbeckwith-oai: Fixed in `009ddb8a`. Promotion classification now starts after the `headRefOid` of the most recently merged `next` → `main` PR, verifies that marker is still an ancestor of `next`, and only falls back to `origin/main..HEAD` before the first promotion. Squash-merged feature/breaking commits therefore cannot leak into later patch classifications.

.github/workflows/create-releases.yml
jbeckwith-oai: Fixed in `009ddb8a`. An `always()` publication-state job now runs after the release job on every `main` push, independently verifies the matching GitHub release tag and exact npm package version, and publishes only when the tag exists but npm is missing the version. `bin/check-npm-version` is shared with `bin/publish-npm`, so retries are idempotent and already-published versions exit successfully.

.github/workflows/create-releases.yml
jbeckwith-oai: Fixed in `009ddb8a`. The workflow now validates every edit/update of the upstream release-please branch and rejects a PR title that does not exactly match `package.json`. `.github/CONTRIBUTING.md` now directs maintainers to merge a conventional commit with a `Release-As: <version>` footer into `main`, allowing release-please to regenerate the package, changelog, title, tag, and release consistently.

.github/workflows/create-releases.yml
```diff
@@ -1,16 +1,254 @@
...
+      - uses: actions/checkout@de0fac2e4500dabe0009e67214ff5f5447ce83dd # v6
+        with:
+          ref: ${{ github.event.pull_request.head.sha }}
+
+      - name: Compare the release title with the generated version
+        env:
+          PR_TITLE: ${{ github.event.pull_request.title }}
+        run: |
+          package_version="$(jq -r -e '.version' package.json)"
+          expected_title="release: ${package_version}"
```
HAYDEN-OAI: **[P1] Make the release-version validation a required merge invariant.** The active `main` ruleset (`19363840`) requires only `lint`, `build`, `detect-breaking-changes`, and `test matrix`; `validate release PR version` is not required. Editing an already-approved release PR title changes no commit SHA, so those previously successful required checks remain valid even while this newly triggered optional job fails. A maintainer can therefore merge `release: 7.4.0` with `package.json` still at `7.3.1`; upstream release-please derives its immutable tag from the merged PR title, recreating the exact tag/package split this change claims to prevent. Add this check to the protected-branch required checks, or enforce title/package equality in the privileged release job before creating any tag.

HAYDEN-OAI (commented):
Reviewed exact current head `009ddb8aa3a1c97e99c817330db834674dfb5273` across the complete five-file diff, prior review conversations, upstream release-please v17.6.0, repository Actions settings/rulesets, the pending Stainless release, promotion classification, npm publication, and retry behavior.

Confirmed the Stainless App still bypasses the `next`/generated-branch protections but no longer bypasses its autonomous release-branch ruleset; Actions PR creation and the upstream branch exclusion are configured; the upstream `prerelease` strategy and `Release-As` footer are supported; standalone `codegen metadata` changes are skipped; and the npm version probe returns the intended published/missing states against npmjs.

Two substantive current-head conversations are still unresolved: publication checks out the mutable push SHA rather than the verified release tag (P1), and concurrent `next` runs can regress the promotion PR title (P2). The additional inline finding identifies why the supposedly fixed title/package-version invariant is still bypassable. CI/test failures are not review findings.

.github/workflows/create-releases.yml
```diff
@@ -1,16 +1,254 @@
...
+              --draft \
+              --title "${promotion_title}" \
+              --body 'Promotes the latest Stainless-generated SDK changes from `next`. Mark this PR ready for review to run CI.'
+            exit 0
+          fi
+
+          gh api \
+            --method PATCH \
+            "repos/${GITHUB_REPOSITORY}/pulls/${promotion_pr}" \
+            -f title="${promotion_title}" \
```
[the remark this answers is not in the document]
jbeckwith-oai: Fixed in `fe124a92`. The promotion job now has branch-scoped concurrency with stale-run cancellation, and it also compares the live `next` ref SHA with `GITHUB_SHA` immediately before creating or updating the PR. An older run can no longer overwrite the title for a newer branch head.

.github/workflows/create-releases.yml
```diff
@@ -19,29 +257,18 @@ jobs:
     steps:
       - uses: actions/checkout@de0fac2e4500dabe0009e67214ff5f5447ce83dd # v6
```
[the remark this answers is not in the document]
jbeckwith-oai: Fixed in `fe124a92`. The publication-state job now exports the verified `v${package_version}` release tag, and the publish job checks out that tag before installing, building, and publishing. Recovery therefore always publishes the immutable tagged source, even when a later `main` commit triggered the retry.

.github/workflows/create-releases.yml
jbeckwith-oai: Fixed in `fe124a92`. Title/package equality is now enforced again inside the privileged `release` job before release-please can create a tag. The job identifies a merged upstream release PR associated with `GITHUB_SHA` and runs the same shared validation used by the PR check; a mismatch fails before the release action. The optional PR check remains fast feedback, while the release path is the non-bypassable invariant.

.github/workflows/create-releases.yml
```diff
@@ -1,47 +1,300 @@
...
+              echo "Last promoted commit ${last_promoted_sha} is not an ancestor of next."
+              echo 'Refusing to classify a rewritten promotion range.'
+              exit 1
+            fi
+            promotion_range="${last_promoted_sha}..HEAD"
+          else
+            promotion_range='origin/main..HEAD'
+          fi
+
+          commit_subjects="$(git log "${promotion_range}" --format='%s')"
```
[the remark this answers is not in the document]
jbeckwith-oai: Fixed in `a326c12c`. Both promotion logs now add `--not origin/main`, so classification considers commits after the last promoted `next` SHA while subtracting every commit already reachable from `main`. Merged `main` features or breaking changes can no longer affect a later generated-only promotion.

.github/workflows/create-releases.yml
```diff
@@ -1,47 +1,300 @@
...
+      - edited
+      - opened
+      - ready_for_review
+      - reopened
+      - synchronize
 
 jobs:
+  promote:
+    name: promote Stainless changes
+    if: github.ref == 'refs/heads/next' && github.repository == 'openai/openai-node'
```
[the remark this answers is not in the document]
jbeckwith-oai: Fixed in `a326c12c`. The promotion job now checks for the open legacy Stainless release PR before comparing or promoting `next` and exits while it remains open. The exact lookup is shared with the release job through `bin/find-legacy-release-pr`, so both sides use the same one-time cutover gate.

.github/workflows/create-releases.yml
```diff
@@ -1,47 +1,299 @@
...
 
     steps:
       - uses: actions/checkout@de0fac2e4500dabe0009e67214ff5f5447ce83dd # v6
-
-      - uses: stainless-api/trigger-release-please@bb6677c5a04578eec1ccfd9e1913b5b78ed64c61 # v1.4.0
-        id: release
         with:
-          repo: ${{ github.event.repository.full_name }}
-          stainless-api-key: ${{ secrets.STAINLESS_API_KEY }}
+          ref: ${{ needs.publication-check.outputs.release_tag }}
```
HAYDEN-OAI: **[P1] Resolve the verified release as an explicit tag, not an ambiguous branch name.** `release_tag` is the unqualified string `v${package_version}`. The pinned checkout action fetches both branches and tags for unqualified refs and [explicitly prefers a matching branch](https://github.com/actions/checkout/blob/de0fac2e4500dabe0009e67214ff5f5447ce83dd/src/ref-helper.ts#L52-L58). None of this repository's active rulesets prevents a same-repository branch writer from creating `refs/heads/v7.3.1`, so that branch shadows the GitHub release tag that `publication-check` actually verified. The protected `main` workflow then installs/builds attacker-selected source with `id-token: write` and can publish an artifact unrelated to the immutable release commit. Check out `refs/tags/${{ needs.publication-check.outputs.release_tag }}` explicitly, or preferably pass and verify the release tag's immutable object SHA.

.github/workflows/create-releases.yml
```diff
@@ -1,47 +1,299 @@
...
+      - uses: actions/checkout@de0fac2e4500dabe0009e67214ff5f5447ce83dd # v6
+
+      - name: Validate a merged release PR version
+        env:
+          GH_TOKEN: ${{ github.token }}
+        run: |
+          release_pr_title="$(
+            gh api \
+              -H 'Accept: application/vnd.github+json' \
+              "repos/${GITHUB_REPOSITORY}/commits/${GITHUB_SHA}/pulls" \
```
HAYDEN-OAI: **[P1] Validate every pending merged release PR before release-please can retry it.** Looking up only PRs associated with `${GITHUB_SHA}` protects the original merge run but not later runs: edit a release PR title to a different version and merge it; this step correctly fails on that merge, leaving the PR labeled `autorelease: pending`. The next unrelated push to `main` has no release PR associated with its own SHA, so this guard silently skips validation, while upstream release-please enumerates *all* merged pending release PRs and derives its tag directly from the earlier edited title. That later run therefore creates the exact mismatched immutable tag this guard is supposed to prohibit; cancellations, failed first attempts, or delayed cutover create the same retry path. Discover and validate each pending merged release PR against the package version at its release commit before invoking release-please, or enforce the version check as a required, non-bypassable merge invariant.

HAYDEN-OAI (commented):
Reviewed exact current head `a326c12c7cfca4233b8dff7e56a5589eac683928` across the complete seven-file diff, prior review conversations, repository rulesets/environment boundaries, the pinned `actions/checkout` implementation, and upstream release-please 17.6.0 release discovery/retry paths.

Confirmed the promotion job is branch-serialized and stale-checked; already-merged `main` commits are excluded from promotion classification; the legacy cutover gate is shared; and publication now attempts to build the release ref rather than the triggering `main` SHA. Two substantive release-safety gaps remain: the release ref is unqualified and therefore can resolve to a same-named branch, and the privileged title/package guard only covers the current push rather than all pending merged release PRs. Inline details follow.

.github/workflows/create-releases.yml
```diff
@@ -1,47 +1,299 @@
...
+              git fetch --no-tags origin "${last_promoted_sha}"
+            fi
+            if ! git merge-base --is-ancestor "${last_promoted_sha}" HEAD; then
+              echo "Last promoted commit ${last_promoted_sha} is not an ancestor of next."
+              echo 'Refusing to classify a rewritten promotion range.'
+              exit 1
+            fi
+            promotion_range="${last_promoted_sha}..HEAD"
+          else
+            promotion_range='origin/main..HEAD'
```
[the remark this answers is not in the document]
jbeckwith-oai: Fixed in `e9d528ba`. When no merged direct-promotion PR exists, the workflow now finds the most recently merged legacy release PR, reads its retained `headRefOid`, and uses that release commit's first parent as the captured `next` boundary. That boundary flows through the same fetch and ancestry validation as later direct-promotion markers, so the first post-cutover promotion cannot reclassify commits already shipped by the legacy squash.

.github/workflows/create-releases.yml
jbeckwith-oai: Fixed in `e2818457`. The publish checkout now uses the fully qualified `refs/tags/${release_tag}` ref, so a same-named branch cannot shadow the GitHub release tag verified by the publication-state job.

.github/workflows/create-releases.yml
```diff
@@ -1,47 +1,303 @@
...
+
+      - name: Create or re-draft the promotion PR
+        env:
+          GH_TOKEN: ${{ github.token }}
+        run: |
+          legacy_pr="$(bash ./bin/find-legacy-release-pr open-number)"
+          if [[ -n "${legacy_pr}" ]]; then
+            echo "Waiting for legacy release PR #${legacy_pr} before promoting next directly."
+            exit 0
+          fi
```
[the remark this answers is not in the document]
jbeckwith-oai: Fixed in `e2818457`. Promotion now runs for both configured push branches and for an unmerged close of the exact legacy PR branch. A legacy merge is covered by its `main` push; an unmerged close is covered by the new `pull_request: closed` event. Each path explicitly checks out `refs/heads/next`, shares one concurrency group, and stale-checks the live `next` SHA before updating a PR.

.github/workflows/create-releases.yml
jbeckwith-oai: Fixed in `e2818457`. Before every release-please invocation, `bin/check-pending-release-versions` now enumerates every merged upstream release PR still labeled `autorelease: pending`, reads `package.json` at each PR's immutable merge commit, and validates that version against the current PR title. Retries on unrelated later `main` pushes therefore fail before tag creation if any pending release was renamed inconsistently.

.github/workflows/create-releases.yml
```diff
@@ -1,15 +1,288 @@
...
+            echo 'should_publish=true' >> "${GITHUB_OUTPUT}"
+          else
+            exit "${version_status}"
+          fi
+
+  publish:
+    name: publish
+    if: needs.publication-check.outputs.should_publish == 'true'
+    needs: publication-check
     runs-on: ubuntu-latest
```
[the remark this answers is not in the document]
jbeckwith-oai: Fixed in `da917cba`. The publish job now uses a repository-wide, non-cancelling `publish-npm` concurrency group. Queued jobs acquire that lock before executing any steps, and `bin/publish-npm` immediately rechecks the exact package version on npm, so a later run exits successfully if an earlier run published while it was queued.

bin/check-pending-release-versions
```diff
@@ -0,0 +1,31 @@
...
+  PR_TITLE="$(jq -r -e '.title' <<< "$RELEASE")"
+  MERGE_SHA="$(jq -r -e '.mergeCommit.oid' <<< "$RELEASE")"
+  PACKAGE_VERSION="$(
+    gh api "repos/${GITHUB_REPOSITORY}/contents/package.json?ref=${MERGE_SHA}" \
+      --jq '.content | @base64d | fromjson | .version'
+  )"
+
+  echo "Validating pending release PR #${PR_NUMBER} at ${MERGE_SHA}."
+  bash ./bin/check-release-version "$PR_TITLE" "$PACKAGE_VERSION"
+done < <(jq -c --arg branch "$RELEASE_BRANCH" '.[] | select(.headRefName == $branch)' <<< "$PENDING_RELEASES")
```
HAYDEN-OAI: **[P1] Validate legacy pending releases before upstream consumes them.** Upstream release-please 17.6.0's `findMergedReleasePullRequests()` scans *every* merged `main` PR carrying `autorelease: pending`; it does not restrict the head branch, and its branch parser accepts `release-please--branches--main--changes--next--components--openai`. This filter silently drops exactly that live legacy PR (#2042) after it merges, while the PR-time version check also excludes the legacy branch. Rename #2042 so its title disagrees with its merged `package.json`: this guard validates nothing, the legacy-open gate clears, and upstream derives an immutable Git tag from the unchecked title. Remove the head-branch filter, or match upstream's full candidate predicate and validate every pending release PR before invoking it.

.github/workflows/create-releases.yml
```diff
@@ -18,30 +291,21 @@ jobs:
 
     steps:
       - uses: actions/checkout@de0fac2e4500dabe0009e67214ff5f5447ce83dd # v6
-
-      - uses: stainless-api/trigger-release-please@bb6677c5a04578eec1ccfd9e1913b5b78ed64c61 # v1.4.0
-        id: release
         with:
-          repo: ${{ github.event.repository.full_name }}
-          stainless-api-key: ${{ secrets.STAINLESS_API_KEY }}
+          ref: refs/tags/${{ needs.publication-check.outputs.release_tag }}
```
HAYDEN-OAI: **[P1] Pin privileged npm publication to the release's immutable commit.** The qualified `refs/tags/` ref fixes branch shadowing, but `publication-check` only verifies that a release exists and forwards its *tag name*. This repository has no tag-targeting rulesets, and its existing GitHub releases report `immutable: false`, so a repository writer can repoint `refs/tags/vX.Y.Z` after that check or while environment approval is pending. Checkout then installs/builds attacker-selected source inside a protected-main job with `id-token: write`, allowing an npm artifact unrelated to the original release commit. Resolve and export the release/tag's immutable commit SHA during `publication-check` (dereferencing annotated tags), then check out and verify that SHA; alternatively enforce immutable releases/protected tags.

HAYDEN-OAI (commented):
Re-reviewed the explicitly requested commit `e2818457445495a19e21a37c409c1951b4b76f52` against the complete eight-file diff, previous conversations, the live legacy release PR, upstream release-please 17.6.0 candidate discovery, repository rulesets, and the protected publication environment. The PR has since advanced to `da917cbada8054314ed8efb458570aae9d5b40c3`; its additional publication-serialization fix resolves the existing concurrency thread, but the two P1 release-safety gaps below remain on both revisions. The legacy promotion boundary, exclusion of released main commits, post-cutover promotion wakeups, qualified tag refs, and promotion serialization/stale-run checks are otherwise present. CI/test failures are not review findings.

bin/check-pending-release-versions
jbeckwith-oai: Fixed in `06209f1b`. `bin/check-pending-release-versions` no longer filters candidates by head branch. It now validates every merged `main` PR labeled `autorelease: pending`, which matches the set upstream release-please can consume and includes the legacy `release-please--branches--main--changes--next--components--openai` PR after merge.

.github/workflows/create-releases.yml
jbeckwith-oai: Fixed in `06209f1b`. `publication-check` now resolves the release tag through lightweight or nested annotated tag objects to its commit SHA and exports only that SHA. The privileged publish job checks out the immutable commit directly and verifies `HEAD` equals the captured SHA before installing, building, or requesting npm provenance, so later tag repointing cannot change the published source.

.github/workflows/create-releases.yml
```diff
@@ -1,47 +1,321 @@
...
+            if grep -q 'HTTP 404' <<< "${release_info}"; then
+              echo "No GitHub release exists for ${release_tag}; nothing to publish."
+              echo 'should_publish=false' >> "${GITHUB_OUTPUT}"
+              exit 0
+            fi
+            echo "Unable to check for GitHub release ${release_tag}:"
+            echo "${release_info}"
+            exit 1
+          fi
+          release_sha="$(bash ./bin/resolve-tag-commit "${release_tag}")"
```
HAYDEN-OAI, line 354: **[P1] Bind the resolved tag SHA to the trusted release/main commit before publishing.** Resolving `refs/tags/vX.Y.Z` once and later checking `HEAD == release_sha` only prevents a tag change *after* resolution; it does not establish that the resolved commit was trustworthy. This repository still has no tag rulesets and releases are `immutable: false`, so a repository writer can move the tag to an attacker-controlled commit immediately after release creation, or between the release-existence request and `resolve-tag-commit`. This line then faithfully captures that malicious SHA, and the protected-main publish job verifies, builds, and publishes it with `id-token: write`. The new resolver accepts this scenario unchanged (a mocked repointed tag returns the attacker SHA successfully). Retain the release response instead of `--silent` and verify the tag commit equals the immutable release-please merge/target commit and is reachable from protected `main`, or enforce immutable/protected release tags before trusting the ref.

HAYDEN-OAI (commented):
Re-reviewed the actual current head `06209f1b320b3fb4eb53a6ff83e3e4cd76656f60` after the requested `da917cbada8054314ed8efb458570aae9d5b40c3` was superseded. Confirmed that all merged `autorelease: pending` PRs, including legacy #2042, are now title/version-validated; publication uses a non-cancelling repository-wide lock and rechecks the exact npm version; lightweight and nested annotated tags resolve correctly; and promotion cutover/concurrency and job privilege boundaries otherwise remain intact. One substantive publication-provenance gap remains: the captured commit itself is selected from an unprotected, mutable tag without being bound to the trusted release/protected-main commit. Inline details follow; failing CI/tests are not review findings.

.github/workflows/create-releases.yml
jbeckwith-oai: Fixed in `04db74dd`. The publication check now retains the GitHub release response and requires its `target_commitish` to be a full commit SHA. It then requires the dereferenced tag SHA to equal that release target, verifies the commit is reachable from the protected `main` push, and confirms `package.json` at that commit contains the expected release version. Only that validated commit SHA crosses the environment-approval boundary and the publish job still verifies its checked-out `HEAD` before executing package code.

.github/workflows/create-releases.yml
```diff
@@ -1,47 +1,321 @@
...
+            exit "${version_status}"
+          fi
+
+  publish:
+    name: publish
+    if: needs.publication-check.outputs.should_publish == 'true'
+    needs: publication-check
+    runs-on: ubuntu-latest
+    concurrency:
+      group: publish-npm
```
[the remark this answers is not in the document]
jbeckwith-oai: Fixed in `4b6f625f`. The non-cancelling publication concurrency group is now keyed by the already validated release SHA. Retries for the same immutable release still serialize and recheck npm after acquiring the lock, while jobs for distinct release commits use distinct groups and therefore cannot evict an intermediate version from GitHub's single pending slot.

.github/workflows/create-releases.yml
```diff
@@ -1,47 +1,345 @@
...
+    runs-on: ubuntu-latest
+    permissions:
+      contents: read
+      pull-requests: write
+
+    steps:
+      - uses: actions/checkout@de0fac2e4500dabe0009e67214ff5f5447ce83dd # v6
+        with:
+          fetch-depth: 0
+          ref: refs/heads/next
```
HAYDEN-OAI: **[P1] Bootstrap the migrated workflow and helper onto `next` before checkout.** The live `next` branch is currently `b246809133bcfb1f1b6313a609ad1651fa602e6d`: its `bin/` contains only `publish-npm`, so `bin/find-legacy-release-pr` returns HTTP 404 at that ref, and its `create-releases.yml` is still the old workflow listening only to `main`. Consequently the first promotion run after this PR merges checks out `next` here and immediately fails on `bash ./bin/find-legacy-release-pr`; subsequent generated pushes to `next` never start the migrated workflow at all (the latest actual `next` push launched only CI). The pending generated SDK changes therefore remain stranded after the legacy release closes unless someone separately bootstraps `next`. Land the migrated workflow/helpers on `next` during cutover, or execute the promotion controller/helpers from protected `main` while inspecting the generated `next` tree separately.

.github/workflows/create-releases.yml
```diff
@@ -1,47 +1,345 @@
...
+          commit_subjects="$(git log "${promotion_range}" --not origin/main --format='%s')"
+          commit_messages="$(git log "${promotion_range}" --not origin/main --format='%s%n%b')"
+          conventional_commit='^(feat|fix|perf|revert|chore|docs|style|refactor|test|build|ci)(\([^)]*\))?!?:'
+          if ! grep -Eq "${conventional_commit}" <<< "${commit_subjects}" && \
+            ! grep -Eq '^BREAKING[ -]CHANGE:' <<< "${commit_messages}"; then
+            echo "No unreleased conventional commits in ${promotion_range}; skipping promotion."
+            exit 0
+          fi
+
+          promotion_type='fix(api)'
```
HAYDEN-OAI: **[P2] Preserve feature severity when generated API changes arrive as `chore(api)`.** The current live generated commit `b246809133bcfb1f1b6313a609ad1651fa602e6d` is titled `chore(api): disable stainless codeflow`, but it adds public SDK API surface: the `gpt-5.5` model and new `name`/`namespace` response fields. `chore` passes the conventional-commit gate above, yet this default classifies that actual feature-bearing commit as `fix(api)`; the squash title then drives release-please to publish new public features as a patch rather than a minor release. Derive release severity from the generated API diff or generator metadata, or otherwise avoid treating every accepted non-`feat`/non-breaking generated change as a fix.

HAYDEN-OAI (commented):
Reviewed exact head `4b6f625f3541bfbbf5652a67b7ec0be6a8ff99aa` across the complete nine-file diff, prior conversations, live branch/ruleset state, and upstream release-please behavior. The previous findings are resolved: all pending merged release PRs, including legacy candidates, are title/version-validated; publication binds the dereferenced tag to its immutable release target, protected-main ancestry, and package version before checking out the captured SHA; and publication locks are scoped to the release SHA. Two substantive cutover and version-classification regressions remain; details are inline. CI/test failures are not review findings.

.github/workflows/create-releases.yml
jbeckwith-oai: Fixed in `c9153b22` without modifying `next`. The promotion controller now lives and executes from protected `main`: a `workflow_run` trigger watches completion of the existing `CI` push workflow on `next`, the job checks out `main` at the workspace root for its workflow helpers, and checks out `next` only under `.next` for read-only diff/history inspection. The merge push to `main` handles the initial cutover/current generated changes, and later `next` CI completions wake the same main-owned controller.

.github/workflows/create-releases.yml
jbeckwith-oai: Fixed in `c9153b22`. Generated changes now default to `feat(api)` unless every release-bearing conventional commit is explicitly patch-class (`fix`, `perf`, or `revert`); breaking markers still select `feat(api)!`. This conservatively classifies the live `chore(api): disable stainless codeflow` commit as a feature release, preserving minor-version severity for its new public model and response fields.

.github/workflows/create-releases.yml
```diff
@@ -1,47 +1,364 @@
...
 name: Create releases
 on:
   push:
     branches:
       - main
+  workflow_run:
+    workflows:
+      - CI
+    types:
+      - completed
```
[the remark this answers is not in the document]
jbeckwith-oai: Fixed in `33467240`. The main-owned promotion controller now wakes on the `requested` activity of the existing `CI` workflow for `next`, rather than waiting for `completed`. Each new `next` push therefore triggers title reconciliation and re-drafting as soon as its CI run is created, closing the interval where a ready PR could merge with the previous release severity.

.github/workflows/create-releases.yml
```diff
@@ -1,47 +1,364 @@
...
 name: Create releases
 on:
   push:
     branches:
       - main
+  workflow_run:
+    workflows:
+      - CI
+    types:
+      - requested
```
HAYDEN-OAI, line 12: **[P1] Gate promotion on completed CI for the exact `next` SHA.** Changing this trigger from `completed` to `requested` starts the protected-`main` controller before the existing `next` CI run finishes: it immediately checks out, reads, classifies, and creates/updates the promotion PR for generated code that is still being validated, and no completion event remains to reconcile it afterward. Simply restoring `completed` is also insufficient because checking out floating `refs/heads/next` could use a newer SHA than the completed run. Handle `requested` only by immediately re-drafting an existing promotion PR without reading `next`; perform checkout/classification/promotion on `completed`, pinning the checkout to `github.event.workflow_run.head_sha` and confirming that SHA is still the live `next` head.

HAYDEN-OAI (commented):
Re-reviewed the current head, including protected-main helper ownership, generated-change severity, legacy pending-release validation, immutable release/tag/target binding, main ancestry, publication concurrency, and permission boundaries. One substantive promotion-ordering issue remains; see the inline comment.

.github/workflows/create-releases.yml
jbeckwith-oai: Fixed in `21457af5`. The controller is now explicitly two-phase. A `workflow_run: requested` job performs no checkout or classification and only re-drafts an existing `next` → `main` PR. The full promotion job runs on `completed` only when that CI run succeeded, checks out `github.event.workflow_run.head_sha`, verifies it is still the live `next` head, and requires a successful `ci.yml` push run for that exact SHA before reading, classifying, or updating the promotion. Main and legacy-gate wakeups use the same exact-SHA CI check, so they cannot promote an unvalidated floating head.

.github/workflows/create-releases.yml
```diff
@@ -1,47 +1,422 @@
...
+          current_next_sha="$(gh api "repos/${GITHUB_REPOSITORY}/git/ref/heads/next" --jq '.object.sha')"
+          if [[ "${current_next_sha}" != "${checked_out_next_sha}" ]]; then
+            echo "A newer next commit (${current_next_sha}) superseded this run (${checked_out_next_sha})."
+            exit 0
+          fi
+
+          successful_ci_runs="$(
+            gh run list \
+              --repo "${GITHUB_REPOSITORY}" \
+              --workflow ci.yml \
```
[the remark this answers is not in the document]
jbeckwith-oai: Fixed in `e70d59b2`. The completed-phase `promote` job now explicitly grants `actions: read` alongside its existing `contents: read` and `pull-requests: write` permissions, so `gh run list` can verify successful `ci.yml` runs for the exact `next` SHA while retaining least privilege.

HAYDEN-OAI (approved):
Reviewed exact commit e70d59b26a5a42b4cab5f0e6f72782215956b2fb. Confirmed requested events only re-draft, successful completed events and main/legacy wakeups require successful push CI for the exact next SHA, and promotion has least-privilege Actions read access. Revalidated generated-change classification, legacy cutover and version/title checks, immutable release/tag/main binding, release-scoped publication concurrency, main-owned helpers, and publish privilege boundaries. No substantive issues found.

PR #2052, opened 2026-08-03
fix: defer release PR output parsing

jbeckwith-oai:
Fixes the post-merge failure in the new Create releases workflow.

When the legacy Stainless release PR gate skips release-please, GitHub still evaluates step environment expressions. Parsing the empty `steps.release.outputs.pr` value with `fromJSON` therefore fails before the re-draft step can be skipped.

Move JSON parsing into the guarded shell body so it runs only when `prs_created == 'true'`, and use `jq -e` to retain strict validation.

Failed run: https://github.com/openai/openai-node/actions/runs/30846459093

Validation:
- `pnpm lint`
- focused `jq -er '.number'` payload check
- thermo-nuclear code quality review

### DIFF
--- a/.github/workflows/create-releases.yml
+++ b/.github/workflows/create-releases.yml
@@ -293,8 +293,9 @@ jobs:
         if: steps.release.outputs.prs_created == 'true'
         env:
           GH_TOKEN: ${{ github.token }}
-          RELEASE_PR: ${{ fromJSON(steps.release.outputs.pr).number }}
+          RELEASE_PR_JSON: ${{ steps.release.outputs.pr }}
         run: |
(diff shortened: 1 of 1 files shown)


HAYDEN-OAI (approved):
Reviewed exact head 8f8413f50b7736111751458eb5eb7295505f4346. Moving release-PR JSON parsing into the guarded shell fixes eager fromJSON evaluation when the legacy gate skips release-please; jq -e fails closed on missing, null, or malformed output. Verified valid PR re-drafting and preserved the #2048 cutover, promotion, permission, immutable-publication, and retry invariants. No substantive issues found.

PR #2089, opened 2026-08-05
fix(format): restore formatting for generated SDK files

apcha-oai:

## Changes being requested

Restore formatting for Stainless-generated SDK files without enabling linting for those files.

The Ultracite migration currently excludes generated files from both Oxfmt and Oxlint. Excluding them from linting is intentional: generated SDK output should not need to satisfy handwritten-code lint rules. Excluding them from formatting is not intentional: SDK generation should still produce consistently formatted source files.

When generated output bypasses formatting, subsequent SDK updates accumulate formatting-only churn and can produce avoidable merge conflicts. This change restores the intended boundary:

- **Formatting:** generated and handwritten SDK files are formatted consistently.
- **Linting:** only handwritten files are linted; Stainless-generated files remain excluded.

The generated files already checked into the repository pass the formatter unchanged, so this does not introduce a generated-source reformat.

### Implementation details

Remove the generated-file exclusion from Oxfmt only, preserve the existing Oxlint exclusion, document the distinction, and add a regression test covering both behaviors.

## Additional context & links

Validated with:

- `node node_modules/vitest/vitest.mjs run tests/oxlint-config.test.ts` — 2 tests passed.
- `node node_modules/oxfmt/bin/oxfmt --check .` — all 597 matched files passed.
- `node node_modules/oxlint/bin/oxlint .` — passed.

### DIFF
--- a/.github/CONTRIBUTING.md
+++ b/.github/CONTRIBUTING.md
@@ -145,8 +145,8 @@ This repository uses [Ultracite](https://www.ultracite.ai/) with
 [Oxlint](https://oxc.rs/docs/guide/usage/linter.html) to format and lint its code.
 The Ultracite presets live in `oxfmt.config.ts` and `oxlint.config.ts`, with
 repository-specific formatting options, import rules, fixture exceptions, and
-generated-file exclusions layered on top. Files with the Stainless-generated
-header are excluded from both formatting and linting; handwritten files in the
+generated-file lint exclusions layered on top. Files with the Stainless-generated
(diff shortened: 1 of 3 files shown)


HAYDEN-OAI (approved):
Approved: generated SDK files are formatted while remaining excluded from Oxlint; normal and fast-format entry points and Castiron regeneration retain the intended boundaries. The companion Castiron toolchain update is tracked in openai/openai#1243796.

PR #2124, opened 2026-08-06
fix(lint): recognize Castiron-generated headers

apcha-oai:
Recognize the new Castiron header alongside the existing Stainless header.


jbeckwith-oai (approved):
Reviewed exact head `abdb05481d4c03474aca4ba83ebe6dbdb4770e95`. The Castiron header is handled alongside Stainless without complicating the scanner, the isolated coverage exercises both generators and rejects handwritten files, and the docs match the behavior. Local validation: targeted test, full unit suite (1,456 tests), lint, and TypeScript typecheck all pass.

PR #2125, opened 2026-08-07
chore(api): update generated SDK attribution

apcha-oai:
## Summary

- Update attribution headers in generated SDK files.
- Regenerate the SDK from the latest API reference.

### DIFF
--- a/.castiron.stats.yml
+++ b/.castiron.stats.yml
@@ -1,6 +1,6 @@
 schema_version: 1
-generation_id: 0113a7b0-33ee-4295-9ec1-12934672acf2
-openapi_spec_hash: d342651bd77c6cd6ea10a9d889b35bb8
-openapi_transformed_spec_hash: 259f19fd2d38c6e784bbb40bf2f9a971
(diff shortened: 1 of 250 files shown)


jbeckwith-oai (changes requested):
This attribution-only regeneration also adds 41 unused imported symbols across 24 generated files (for example, `src/resources/admin/admin.ts`, `src/resources/shared.ts`, and `src/resources/fine-tuning/methods.ts`; the two Responses files also add an unused `PagePromise`). The generated-file lint exclusion hides the full set, although code-quality has already flagged representative instances. Please fix Castiron import collection to emit only referenced symbols and regenerate instead of hand-editing generated output. After ignoring the attribution headers, these imports are the only TypeScript changes, so the regenerated source should remain attribution-only.

jbeckwith-oai (approved):
Re-reviewed the exact current generated tree. The prior unused-import blocker is resolved: the 41 spurious imports are gone, the changed generated files pass the generated-import lint policy, the promoted public tree exactly matches the Castiron source tree, and the generation metadata matches the transformed OpenAPI artifact.

PR #2126, opened 2026-08-07
fix(lint): restore generated SDK lint rules

apcha-oai:
## Summary

- Restore historical generated-code checks for unused imports and SDK self-imports with one shared Oxlint configuration.
- Reuse the configuration for CI checks and autofixes, retrying overlapping fixes until stable.
- Preserve stricter handwritten lint rules and generated test import exemptions.


jbeckwith-oai (changes requested):
Blocking: the restored generated-file checks are still absent from the canonical package commands documented in CONTRIBUTING.md. `package.json` maps `pnpm lint` directly to `ultracite check` and `pnpm format` directly to `ultracite fix`; because `oxlint.config.ts` ignores generated files, both commands bypass `scripts/lint-generated.cjs`. A contributor can therefore follow the documented commands, leave an unused generated import behind, and only discover it when CI runs `./scripts/lint`. Please route the package aliases through `./scripts/lint` and `./scripts/format` (or make the documented/canonical contract consistent) and add a regression that exercises those public entrypoints.

apcha-oai:
r? fixed to just route to scripts/format

jbeckwith-oai (approved):
Re-reviewed exact head 60bc8ed. The prior blocker is resolved: the documented `pnpm lint` and `pnpm format` entrypoints now route through the canonical SDK wrappers, and the regression exercises both public commands end to end. I also reviewed the full generated-lint implementation, configuration, fast-format integration, and tests; targeted tests (8/8) and `pnpm lint` pass locally, and CI build/lint/tests/CodeQL/ecosystem checks are green. No remaining correctness, API, or maintainability blockers.

PR #2150, opened 2026-08-08
fix(embeddings): handle compatible endpoints that return float arrays

### LINKED ISSUE #1542: Embedding dimensions are truncated to 256 in some cases

anishesg:
## Changes being requested

- Fixed embedding dimension truncation when using OpenAI-compatible endpoints (ollama, LM Studio)
- Added type check in `src/resources/embeddings.ts` to detect when embeddings are already returned as float arrays
- Added regression test for endpoints that ignore the `encoding_format` parameter

## Additional context & links

Some OpenAI-compatible endpoints like ollama and LM Studio ignore the `encoding_format` parameter and always return embeddings as float arrays instead of base64-encoded strings. The SDK was attempting to decode these arrays as base64, which caused Buffer.from() to interpret the array as bytes rather than a base64 string. For a 1024-element float array, this created a 1024-byte buffer, which when divided by 4 (Float32Array.BYTES_PER_ELEMENT) resulted in exactly 256 dimensions.

The fix adds an Array.isArray() check before attempting base64 decoding. If the embedding is already an array, it's used directly. Otherwise, it's decoded from base64 as before. This preserves backward compatibility with the official OpenAI API while supporting endpoints that return float arrays.

Fixes #1542

### DIFF
--- a/src/resources/embeddings.ts
+++ b/src/resources/embeddings.ts
@@ -57,7 +57,16 @@ export class Embeddings extends APIResource {
       if (response && response.data) {
         response.data.forEach((embeddingBase64Obj) => {
           const embeddingBase64Str = embeddingBase64Obj.embedding as unknown as string;
-          embeddingBase64Obj.embedding = toFloat32Array(embeddingBase64Str);
+          // Some OpenAI-compatible endpoints (e.g., ollama, LM Studio) ignore the encoding_format
+          // parameter and always return float arrays. Check if the embedding is already an array
+          // before attempting base64 decoding.
+          if (Array.isArray(embeddingBase64Str)) {
+            // Already a float array, no decoding needed
+            embeddingBase64Obj.embedding = embeddingBase64Str;
+          } else {
+            // Base64 string, decode it
+            embeddingBase64Obj.embedding = toFloat32Array(embeddingBase64Str);
+          }
         });
       }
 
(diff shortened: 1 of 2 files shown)


HAYDEN-OAI:
Thanks for the report! This occurs when a third-party endpoint ignores the requested base64 embedding encoding; the OpenAI API honors that contract. For compatible endpoints that return float arrays, pass encoding_format: 'float' explicitly, which bypasses base64 decoding. We’re not planning to add provider-specific handling for responses that violate the requested encoding format.

PR #2151, opened 2026-08-09
fix(client): clean up abort listener in fetchWithTimeout to prevent memory leak

lchetanlc:
Fixes #1811

### Description
In `fetchWithTimeout`, an event listener is added to the user-provided `AbortSignal` so the SDK can abort the internal fetch request if the user triggers their signal. 

However, this event listener was never removed upon successful completion of the request. While Node.js ignores `AbortSignal` timers for process lifecycle management, strict runtimes like **Deno** keep the process alive as long as an `AbortSignal.timeout()` has an active listener attached. 

This resulted in SDK calls hanging the entire process until the full timeout duration expired, even if the API responded successfully within milliseconds.

### Fix
This PR simply adds `if (signal) signal.removeEventListener('abort', abort);` to the `finally` block of `fetchWithTimeout` to clean up the orphaned listener, allowing strict JavaScript runtimes to exit cleanly.

### Testing
- Verified 1543 tests pass via `pnpm test`.

### DIFF
--- a/src/client.ts
+++ b/src/client.ts
@@ -795,6 +795,13 @@ export class OpenAI {
 
     const security = options.__security ?? { bearerAuth: true };
     const controller = new AbortController();
+    const abort = this._makeAbort(controller);
+    let abortCleanup: (() => void) | undefined;
+    if (options.signal) {
+      options.signal.addEventListener('abort', abort);
+      abortCleanup = () => options.signal!.removeEventListener('abort', abort);
+    }
+
     const response = await this.fetchWithAuth(url, req, timeout, controller, security).catch(castToError);
     const headersTime = Date.now();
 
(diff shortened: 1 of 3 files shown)


lchetanlc:
@chatgpt-codex-connector Great catch! I have updated the PR. 

Instead of eagerly removing the listener in `fetchWithTimeout`, the cleanup function is now passed through `APIResponseProps` and into the response lifecycle. 

For standard responses, it cleans up immediately after the body is parsed. For streams, it passes the cleanup hook down to the `Stream` async iterator, ensuring the `AbortSignal` is preserved for the entire duration of the stream and only removed when the stream finishes or closes.

HAYDEN-OAI:
Thanks for working on the abort-listener leak. This change also modifies src/client.ts, which is generated from the shared TypeScript client template, so it cannot be merged directly into this SDK repository. The same upstream issue is tracked in #1811 and the more comprehensive investigation in #2086. I am marking this upstream so cancellation cleanup can be fixed at the generator level without losing post-header or streaming abort behavior.

mccrearydamion69-crypto:
Hayden I think I'm on your Linkin park

HAYDEN-OAI:
Resolved upstream in https://github.com/openai/openai/pull/1291086; the abort-listener fix will be available in the next SDK release. Closing this alternative downstream implementation as superseded.

PR #2159, opened 2026-08-10
fix(helpers): preserve constrained object oneOf schemas

### LINKED ISSUE #2152: helpers/standard-schema: normalizeStructuredOutputSchema rewrites `oneOf` into `anyOf` on object nodes, then toStrictJsonSchema rejects its own output
### Summary

On `openai@7.4.0`, `helpers/standard-schema` and `helpers/zod` disagree about one node shape: an object schema that also carries `oneOf`. `resolveStandardJSONSchema()` runs

```js
return toStrictJsonSchema(normalizeStructuredOutputSchema(schema));
```

and `normalizeStructuredOutputSchema()` moves `oneOf` branches into `anyOf` — which produces exactly the shape `toStrictJsonSchema()` refuses two lines later ("Object anyOf schema at `…` cannot be represented…"). The same document handed straight to `toStrictJsonSchema()` is **accepted byte-identical**.

So whether a schema is "valid" depends on which helper you call, at the same SDK version, and the standard-schema path fails on input it did not have to touch.

### Repro (no API key, `openai@7.4.0`)

```js
import { toStrictJsonSchema } from "openai/lib/transform.js";
import { standardResponseFormat } from "openai/helpers/standard-schema.js";

const doc = {
  type: "object",
  properties: {
    f: {
      type: "object",
      properties: { a: { type: "string" } },
      required: ["a"],
      additionalProperties: false,
      oneOf: [{ type: "string" }, { type: "number" }],
    },
  },
  required: ["f"],
  additionalProperties: false,
};

const asStandard = (s) => ({ "~standard": {
  version: 1, vendor: "repro", validate: (v) => ({ value: v }),
  jsonSchema: { input: () => s },
} });

toStrictJsonSchema(doc);                    // ACCEPT (returned byte-identical)
standardResponseFormat(asStandard(doc), "n"); // THROW
```

```
CONTROL (plain object)
  toStrictJsonSchema      : ACCEPT
  standardResponseFormat  : ACCEPT

object node carrying `oneOf`
  toStrictJsonSchema      : ACCEPT
  standardResponseFormat  : THROW: Object anyOf schema at `properties/f` cannot be
                            represented in strict Structured Outputs without changing
                            Draft 7 validation.
```

The rewrite is the cause, not the input. Performing it by hand and handing the result to `toStrictJsonSchema()` reproduces the identical error:

```js
const rewritten = structuredClone(doc);
rewritten.properties.f.anyOf = rewritten.properties.f.oneOf;
delete rewritten.properties.f.oneOf;
toStrictJsonSchema(rewritten); // THROW: Object anyOf schema at `properties/f` …
```

### Reachability — stated plainly, because it is narrower than it looks

This is **not** reachable from `zod`, and I checked rather than assumed. `z.discriminatedUnion()` does emit `oneOf`, but at a bare union node with no object shape, so it is accepted by both families when nested; at the root both correctly reject it for an unrelated reason (`Root schema must have type: 'object'`). Measured on `zod@4.4.3`:

| | `zodResponseFormat` | `standardResponseFormat` |
|---|---|---|
| nested `z.discriminatedUnion` | ACCEPT | ACCEPT |
| root `z.discriminatedUnion` | THROW (root type) | THROW (root type) |

The route that does reach it is the documented escape hatch: `props.schema`, which lets a caller pass a raw JSON Schema (`const { schema, ...formatProps } = props ?? {}`). Anyone adapting an existing JSON Schema — OpenAPI-derived or hand-written, where `oneOf` beside `properties` is an ordinary idiom — goes through `normalizeStructuredOutputSchema()` and gets a throw on a document the zod path accepts unchanged.

### Suggested fix

Skip the `oneOf` → `anyOf` move when the node has object shape and the result would be refused — i.e. when `hasObjectShape(node)` and the existing escape hatch in the object-`anyOf` check would not apply (redundant `object` wrapper type, no object keywords of its own, every branch object-only). Leaving `oneOf` in place there is strictly better than rewriting it: `toStrictJsonSchema()` already accepts that form verbatim.

### One test-design note

A regression test written on a `z.discriminatedUnion` fixture **passes either way** — that shape is a bare union node, which is precisely the case the rewrite handles correctly. The failing case needs `oneOf` on a node that *also* declares `type: "object"` plus at least one object keyword, so it is worth pinning with a raw-JSON-Schema fixture rather than a validator-derived one.

aryanputta:
Fixes #2152.

The Standard Schema helper rewrites every provably exclusive `oneOf` into `anyOf`. For a schema node that already has its own object constraints, this changes the schema into a shape that the strictifier rejects, even though the original `oneOf` is accepted.

This change:
- preserves `oneOf` on object-shaped nodes with their own object constraints;
- continues normalizing standalone and redundant object-union wrappers;
- adds a regression test for a raw JSON Schema with `type: object`, object properties, and `oneOf`.

Validation:
- `git diff --check` passed.
- Node syntax checks passed for the changed TypeScript files.
- pnpm dependency installation and Vitest were blocked because the sandbox could not resolve registry.npmjs.org.

### DIFF
--- a/src/helpers/standard-schema.ts
+++ b/src/helpers/standard-schema.ts
@@ -387,17 +387,36 @@ function normalizeStructuredOutputSchema(schema: JSONSchema): JSONSchema {
           'Standard JSON Schema generated both `anyOf` and `oneOf`, which cannot be represented in an OpenAI strict schema',
         );
       }
-      // `false` can never validate, so it cannot overlap another oneOf
-      // branch. Keep it in place until the existing anyOf normalization runs
-      // so local refs into surviving branch indices can be rewritten before
-      // the impossible alternatives are removed.
-      const possibleBranches = record['oneOf'].filter((branch) => branch !== false);
-      if (!areOneOfBranchesMutuallyExclusive(possibleBranches, normalizedSchema)) {
-        throw new OpenAIError(
-          'Standard JSON Schema generated a `oneOf` whose branches are not provably mutually exclusive. OpenAI strict schemas do not support `oneOf`; use `anyOf` or add a discriminator with distinct literal values.',
+      const hasObjectShape =
+        record['type'] === 'object' ||
+        (Array.isArray(record['type']) && record['type'].includes('object')) ||
+        ['properties', 'required', 'additionalProperties', 'patternProperties', 'propertyNames'].some(
+          (keyword) => record[keyword] !== undefined,
         );
+      const hasOwnObjectConstraints = [
+        'properties',
+        'required',
+        'additionalProperties',
+        'patternProperties',
+        'propertyNames',
+        'minProperties',
+        'maxProperties',
+        'dependentRequired',
+        'dependentSchemas',
+      ].some((keyword) => record[keyword] !== undefined);
+      if (!hasObjectShape || !hasOwnObjectConstraints) {
+        // `false` can never validate, so it cannot overlap another oneOf
+        // branch. Keep it in place until the existing anyOf normalization runs
+        // so local refs into surviving branch indices can be rewritten before
+        // the impossible alternatives are removed.
+        const possibleBranches = record['oneOf'].filter((branch) => branch !== false);
+        if (!areOneOfBranchesMutuallyExclusive(possibleBranches, normalizedSchema)) {
+          throw new OpenAIError(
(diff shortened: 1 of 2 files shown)


HAYDEN-OAI:
Thanks for the contribution. We traced this all the way to the production Structured Outputs validators: strict schemas explicitly reject oneOf, so preserving it here would produce a request the API cannot accept. Safe exclusive unions need to normalize to anyOf, while object-constrained unions that cannot be represented safely must continue to fail closed. The regression fixture also requires the same value to be both an object and either a string or a number, so no JSON value can satisfy it. Closing because this output is outside the supported strict-schema contract.

PR #2275, opened 2026-08-10
chore(lint): enable curly

HAYDEN-OAI:
## Summary

- Removes `curly` from `compatibilityRules`.
- Resolves existing diagnostics with behavior-preserving fixes or narrowly documented exceptions.

## Stack

- Part 183 of 185.
- Base: `dev/hayden/ultracite-182-func-names`.
- This PR is stacked on the prior rule cleanup.

## Validation

Validated cumulatively at the stack head:

- `pnpm lint`
- `NODE_NO_WARNINGS=1 pnpm exec vitest run --config vitest.config.mts --update=none`
- `pnpm run build`
- `node --experimental-strip-types scripts/test-packed-package.ts`

### DIFF
--- a/ecosystem-tests/browser-direct-import/public/index.js
+++ b/ecosystem-tests/browser-direct-import/public/index.js
@@ -40,7 +40,7 @@ async function runTests() {
     displayResults();
   }
   const runningEl = document.querySelector('#running');
-  if (runningEl) runningEl.remove();
+  if (runningEl) {runningEl.remove();}
 }
 
 /** @type {string[]} */
@@ -86,7 +86,7 @@ function expect(received) {
     },
     toBeSimilarTo(comparedTo, expectedDistance) {
       const actualDistance = distance(received, comparedTo);
-      if (actualDistance < expectedDistance) return;
+      if (actualDistance < expectedDistance) {return;}
 
       throw new Error(
         [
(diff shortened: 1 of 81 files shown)


jbeckwith-oai (changes requested):
This PR is scoped and described as behavior-preserving curly cleanup, but the two mock request handlers also change the thrown error from Error to TypeError. That is observable through instanceof/name and is unrelated to adding braces. Please keep the existing Error behavior, or split and explicitly justify/test that behavior change separately.

jbeckwith-oai (approved):
Re-reviewed this exact head comprehensively. The curly transformation is semantics-preserving, and the previous blocker is fixed: both mock-handler sites retain their historical Error identity with narrow documented lint exceptions. CI is green; no remaining blockers.

jbeckwith-oai (approved):
Thermo-nuclear exact-head re-review complete. The rebased commit has the same stable patch-id as the previously approved fix. I verified the original blocker remains resolved: both mock request handlers still throw the historical Error type, with narrow documented prefer-type-error suppressions, so the unrelated observable Error-to-TypeError regression is not present. The remaining curly changes are mechanical and preserve control flow, API/type/runtime behavior. Diff hygiene is clean and no file newly crosses 1k lines. Lint/build/examples/detect-breaking/CodeQL and Node 24/26 pass. Node 22 failed only in an untouched generated skills request test with the known mock-server APIConnectionError; unrelated to this patch. No remaining blockers.

Issue #12, opened 2022-05-01
 Providing a user identifer 


ezzcodeezzlife:
OPENAI Safty best practices:

Issue #11, opened 2022-04-28
429 error despite zero API usage


ezzcodeezzlife:
im using the nodejs example from [the docs](https://beta.openai.com/docs/api-reference/completions/create). Inserted my API key. 

```javascript
const { Configuration, OpenAIApi } = require("openai");
const configuration = new Configuration({
  apiKey: "MY-KEY",
});

async function getCompletion () {
  const openai = new OpenAIApi(configuration);
  const response = await openai.createCompletion("text-curie-001", {
    prompt: "Say this is a test",
    max_tokens: 5
  })
  .catch(err => {
    console.log(err);
  });
  console.log(response);
}

getCompletion();

```
returns: 

```
 response: {
    status: 429,
    statusText: 'Too Many Requests',
    headers: {
      date: 'Thu, 28 Apr 2022 12:51:34 GMT',
      'content-type': 'application/json; charset=utf-8',
      'content-length': '205',
      connection: 'close',
      vary: 'Origin',
      'x-request-id': 'xxxx',
      'strict-transport-security': 'max-age=15724800; includeSubDomains'
    },
    config: {
      transitional: [Object],
      adapter: [Function: httpAdapter],
      transformRequest: [Array],
      transformResponse: [Array],
      timeout: 0,
      xsrfCookieName: 'XSRF-TOKEN',
      xsrfHeaderName: 'X-XSRF-TOKEN',
      maxContentLength: -1,
      maxBodyLength: -1,
      validateStatus: [Function: validateStatus],
      headers: [Object],
      method: 'post',
      data: '{"prompt":"Say this is a test","max_tokens":5}',
      url: 'https://api.openai.com/v1/engines/text-curie-001/completions'
    },
    request: <ref *1> ClientRequest {
      _events: [Object: null prototype],
      _eventsCount: 7,
      _maxListeners: undefined,
      outputData: [],
      outputSize: 0,
      writable: true,
      destroyed: false,
      _last: true,
      chunkedEncoding: false,
      shouldKeepAlive: false,
      _defaultKeepAlive: true,
      useChunkedEncodingByDefault: true,
      sendDate: false,
      _removedConnection: false,
      _removedContLen: false,
      _removedTE: false,
      _contentLength: null,
      _hasBody: true,
      _trailer: '',
      finished: true,
      _headerSent: true,
      socket: [TLSSocket],
      _header: 'POST /v1/engines/text-curie-001/completions HTTP/1.1\r\n' +
[...]

PR #2280, opened 2026-08-11
fix: prevent prototype pollution in assistant stream deltas

HAYDEN-OAI:
## Summary

- Reject unsafe prototype-chain property names before reading, writing, or recursively merging AssistantStream deltas.
- Preserve normal accumulator identity, own-property behavior, inherited values, and null-prototype objects with focused regression coverage.
- Resolve GitHub CodeQL `js/prototype-pollution-utility` alerts [#11](https://github.com/openai/openai-node/security/code-scanning/11) and [#12](https://github.com/openai/openai-node/security/code-scanning/12).

## Why

`AssistantStream.accumulateDelta` recursively copies streamed properties into existing objects. Guarding reserved prototype-chain names before the first dynamic property access prevents those values from reaching either flagged assignment while failing explicitly with the SDK's existing `OpenAIError`.

## Verification

- `pnpm exec vitest run --config vitest.config.mts --update=none tests/lib/AssistantStream.test.ts tests/streaming/assistants/assistant.test.ts` — 44 tests passed on the final commit.
- `pnpm exec tsc --noEmit` — passed on the final commit.
- `pnpm lint` — passed on the final commit.
- `pnpm exec vitest run --config vitest.config.mts --update=none` — 1,464 tests passed before the final formatting-only rebase.
- `pnpm build` — CommonJS and ESM builds passed before the final formatting-only rebase.
- Built-package smoke test passed on supported Node.js 22.

### DIFF
--- a/src/lib/AssistantStream.ts
+++ b/src/lib/AssistantStream.ts
@@ -673,6 +673,10 @@ export class AssistantStream
 
   static accumulateDelta(acc: Record<string, any>, delta: Record<string, any>): Record<string, any> {
     for (const [key, deltaValue] of Object.entries(delta)) {
+      if (key === '__proto__' || key === 'constructor' || key === 'prototype') {
+        throw new OpenAIError(`Assistant stream delta contains an unsafe property: ${key}`);
+      }
+
       if (!hasOwn(acc, key)) {
         acc[key] = deltaValue;
         continue;
--- a/src/lib/AssistantStream.ts
+++ b/src/lib/AssistantStream.ts
@@ -630,6 +630,16 @@ export class AssistantStream
 
         //If this delta does not have content, nothing to process
         if (data.delta.content) {
+          assertSafeAssistantStreamDelta(data.delta);
+
+          for (const contentElement of data.delta.content) {
+            if (!Number.isInteger(contentElement.index) || contentElement.index < 0) {
+              throw new OpenAIError(
+                `Assistant stream delta contains an invalid content index: ${contentElement.index}`,
+              );
+            }
+          }
+
           for (const contentElement of data.delta.content) {
             if (contentElement.index in snapshot.content) {
               const currentContent = snapshot.content[contentElement.index];
@@ -672,6 +682,8 @@ export class AssistantStream
   }
 
   static accumulateDelta(acc: Record<string, any>, delta: Record<string, any>): Record<string, any> {
+    assertSafeAssistantStreamDelta(delta);
+
     for (const [key, deltaValue] of Object.entries(delta)) {
       if (key === '__proto__' || key === 'constructor' || key === 'prototype') {
         throw new OpenAIError(`Assistant stream delta contains an unsafe property: ${key}`);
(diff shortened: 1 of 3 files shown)


src/lib/AssistantStream.ts
```diff
@@ -673,6 +673,10 @@ export class AssistantStream
 
   static accumulateDelta(acc: Record<string, any>, delta: Record<string, any>): Record<string, any> {
     for (const [key, deltaValue] of Object.entries(delta)) {
+      if (key === '__proto__' || key === 'constructor' || key === 'prototype') {
```
jbeckwith-oai, line 688: `accumulateDelta` only checks objects it actually recurses into. The `!hasOwn(acc, key)` path below copies `deltaValue` wholesale, and the new-array-index path copies `deltaEntry` wholesale, so nested unsafe keys bypass this guard. On this exact head, `AssistantStream.accumulateDelta({}, JSON.parse(`{"payload":{"__proto__":{"polluted":true}}}`))` succeeds; `Object.assign({}, result.payload).polluted` is then `true`. An equivalent `items: [{index: 0, "__proto__": ...}]` payload also succeeds. Please recursively preflight every object/array in the delta for `__proto__`, `constructor`, and `prototype` before performing any writes (which also avoids partially mutating `acc` before a later rejection), and add regression tests for root, nested-object, and new-array-entry payloads.

jbeckwith-oai (changes requested):
The top-level key check closes the direct CodeQL sink, but it does not enforce the stated security boundary across the full delta tree. I reproduced this exact head rejecting all three unsafe names at the root while accepting the same names inside a newly inserted object or array entry; the resulting snapshot can still become a prototype-pollution gadget downstream. The added tests only cover benign accumulator/prototype behavior, so neither they nor the passing CodeQL run exercise this bypass. Please validate the entire delta recursively before mutating the accumulator, cover object and array insertion paths, and assert that rejected input leaves the accumulator unchanged.

src/lib/AssistantStream.ts
HAYDEN-OAI: Fixed in e8b0bedf. `assertSafeAssistantStreamDelta` now recursively preflights every object and array before any accumulator write, including newly inserted object properties and indexed array entries. Regression tests cover all three reserved names at the root, inside nested objects, in new array entries, and within streamed message content; rejected updates leave the existing accumulator or snapshot unchanged. The full handwritten suite passes all 1,476 tests.

src/lib/AssistantStream.ts
```diff
@@ -673,6 +673,10 @@ export class AssistantStream
...
+        throw new OpenAIError(`Assistant stream delta contains an unsafe property: ${key}`);
+      }
```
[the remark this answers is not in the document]
HAYDEN-OAI: Fixed in e8b0bedf. `#accumulateMessage` now validates every streamed content index as a non-negative integer in a dedicated preflight pass before the `in` lookup or any snapshot assignment. Regression coverage includes string, negative, fractional, and null indices, and verifies that an earlier valid item is not applied when a later index is invalid.

jbeckwith-oai (approved):
Re-reviewed exact head e8b0bedf47384305cdbba56840c0100a95e318ec. The recursive preflight now rejects __proto__, constructor, and prototype throughout nested objects and arrays before any accumulator mutation, including newly inserted values. Message-content indexes are validated before snapshot writes, and regression coverage exercises root, nested-object, and array paths plus unchanged state on rejection. Focused tests and exact-head Node 22/24/26 CI, lint, build, ecosystem, and CodeQL pass. The detect-breaking-changes failure is unrelated shallow-fetch infrastructure: that job cannot resolve the base tree.

jbeckwith-oai (approved):
Re-reviewed exact head 8485006d13b919ff3c5198a928e72d3c5aee1030. The only change since the previously approved head is the breaking-change workflow exact-base fetch. The event-derived base SHA is quoted, uses read-only contents permissions, and is fetched without tags; exact-head CI confirms that the formerly failing detector now resolves the base tree and passes. The AssistantStream implementation and security regressions are byte-for-byte unchanged, so recursive preflight rejection remains intact across root, nested, newly inserted object, and array paths before mutation. Focused tests pass 53/53, and all exact-head lint, build, Node 22/24/26, ecosystem, CodeQL, and detector checks are green.

PR #2297, opened 2026-08-12
chore: limit Dependabot to 10 open pull requests

HAYDEN-OAI:
## Summary

- Limit each of the two independent npm Dependabot update configurations to five open version-update pull requests.
- Keep the combined maximum at ten without changing root dependency updates or ecosystem-test lockfile-only updates.
- Security update pull requests remain unaffected because GitHub does not apply this setting to them.

## Validation

- `git diff --check`
- Parsed `.github/dependabot.yml` with Ruby and confirmed both update-group limits sum to 10, the root directory remains `/`, and ecosystem tests retain `versioning-strategy: lockfile-only`.

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -2,6 +2,7 @@ version: 2
 updates:
   - package-ecosystem: npm
     directory: /
+    open-pull-requests-limit: 5
     schedule:
       interval: weekly
     cooldown:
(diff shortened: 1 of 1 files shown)


jbeckwith-oai (approved):
Reviewed the exact Dependabot configuration change and validated the per-update-entry limit semantics. The two independent entries correctly cap version-update PRs at five each (ten total), while security updates remain excluded; the YAML remains valid and the change is narrowly scoped.

PR #2302, opened 2026-08-12
fix: await webhook verification in README examples

HAYDEN-OAI:
## Summary\n- Await asynchronous webhook signature verification before parsing or trusting events in both root README examples.\n- Explain that webhook verification is asynchronous.\n- Add a regression test that checks every documented webhook verification call is awaited.\n\n## Why\nThe previous examples processed events and could return HTTP 200 before signature verification rejected, bypassing the surrounding error handler. The dedicated webhook guide already documents the correct await requirement.\n\n## Verification\n- Full handwritten SDK suite: 1,584 tests passed.\n- Focused webhook tests: 7 passed.\n- Repository lint, TypeScript compilation, and Node version policy check passed.

### DIFF
--- a/README.md
+++ b/README.md
@@ -269,6 +269,9 @@ await client.files.create({
 
 Verifying webhook signatures is _optional but encouraged_.
 
+Webhook verification is asynchronous. Always `await` `unwrap()` or `verifySignature()` before trusting
+the event.
+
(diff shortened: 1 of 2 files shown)


jbeckwith-oai (approved):
Reviewed the exact head. The README now awaits both asynchronous webhook verification APIs before using their results, the focused regression covers every matching TypeScript example, and the full CI suite is green.

PR #2303, opened 2026-08-12
fix: preserve skill upload paths without buffering native files

HAYDEN-OAI:
## Summary
- Normalize explicitly preserved Windows-style Skills paths before buffered or streaming multipart encoding.
- Keep directory stripping unchanged for ordinary uploads and cover both Skills endpoints with parsed multipart requests.
- Rename native `File` values and apply metadata overrides without eagerly reading their contents.
- Preserve file identity when the effective name and metadata are unchanged.

## Why
The streaming encoder previously emitted `%5C` for Windows path separators, which the production Skills backend rejects. Native File renames also duplicated complete file contents in memory, including browser directory uploads.

## Verification
- Full handwritten suite: 1,596 tests passed.
- Focused upload suites: 89 tests passed.
- TypeScript, repository lint, and full CJS/ESM package build passed.
- Actual backend normalization and manifest validation accept the repaired filenames.
- A 32 MiB native File rename retains zero additional ArrayBuffer bytes.

### DIFF
--- a/src/internal/to-file.ts
+++ b/src/internal/to-file.ts
@@ -75,8 +75,16 @@ const isResponseLike = (value: any): value is ResponseLike =>
   typeof value.url === 'string' &&
   typeof value.blob === 'function';
 
-const hasFilePropertyOverrides = (options: FilePropertyBag | undefined): boolean =>
-  options?.type != null || options?.lastModified != null || options?.endings != null;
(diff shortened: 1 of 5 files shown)


jbeckwith-oai (approved):
Reviewed the exact head. Native File renames now preserve their Blob payload without calling arrayBuffer(), metadata overrides remain correct, path normalization is scoped to endpoints that explicitly preserve filenames, and the comprehensive buffered/streaming regression coverage and full CI suite are green.

PR #2306, opened 2026-08-12
ci: run Vitest performance benchmarks in CI

HAYDEN-OAI:
## Summary

- Run the existing Vitest performance benchmark suite in the normal CI workflow and upload its JSON report as a 30-day artifact.
- Add nine deterministic request-preparation benchmarks covering header merging, flat and nested query serialization, URL construction, authenticated GET/POST preparation, and an in-memory chat-completions request.
- Include the new benchmark in scheduled-run fixture metadata and document the expanded CI coverage.

The suite runs once with the repository tooling Node.js version, avoids external API calls, and does not introduce noisy shared-runner performance thresholds.

## Validation

- `pnpm bench:json` — 47 benchmark cases passed across three files.
- `pnpm test:unit` — 1,727 tests passed across 66 files.
- `pnpm lint`
- `pnpm exec tsc`
- Validated normal CI benchmark execution/artifact upload and scheduled benchmark fixture hashing.

### DIFF
--- a/.github/CONTRIBUTING.md
+++ b/.github/CONTRIBUTING.md
@@ -121,22 +121,23 @@ $ pnpm bench:json
 ```
 
 This writes `benchmark-results.json` in the repository root. The report is ignored
-by Git and is uploaded by the separate, manually triggered or scheduled benchmark
-workflow alongside a runtime, runner, revision, and fixture-hash metadata file.
-Pass a benchmark name or file filter directly to run only part of the suite, for
-example:
+by Git and uploaded as an artifact by the performance-benchmark job in normal CI.
+The separate, manually triggered or scheduled benchmark workflow also uploads a
+runtime, runner, revision, and fixture-hash metadata file. Pass a benchmark name
+or file filter directly to run only part of the suite, for example:
 
 ```sh
 $ pnpm bench streaming
 ```
 
-Benchmarks cover SSE chunk decoding and JSON parsing, incremental structured
-output parsing, schema generation and validation, and base64-versus-float
-embedding responses. Each case prepares its fixtures before timing and uses
-explicit warmup and repeated measurements. Compare medians and tail latency only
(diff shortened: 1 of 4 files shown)


jbeckwith-oai (changes requested):
The benchmark fixtures and CI execution are otherwise well-structured, and the exact-head suite/checks pass. One merge-gating issue remains:

The new `performance benchmarks` job is not part of any required check. The active `main` ruleset requires `lint`, `build`, `detect-breaking-changes`, and `test matrix`; `test matrix` currently has `needs: [test]` and only checks `needs.test.result`. As a result, a benchmark crash, timeout, fixture assertion failure, or missing artifact can leave every required check green and allow the PR to merge, which defeats the purpose of adding the suite to normal CI.

Please add `benchmarks` to the existing `test_matrix.needs`, pass `needs.benchmarks.result`, and fail the aggregate unless both test and benchmark jobs succeed (or make `performance benchmarks` required directly). This does not require timing thresholds; only successful benchmark execution needs to gate.

HAYDEN-OAI:
Addressed the merge-gating review in `d70775cb`: `test matrix` now depends on `benchmarks`, passes `needs.benchmarks.result`, and fails unless both the test matrix and performance benchmarks succeed.

Validation:
- `git diff --check`
- `pnpm lint`
- `pnpm bench:json` (full benchmark suite passed)

jbeckwith-oai (approved):
The prior merge-gating blocker is fully resolved. The required `test matrix` aggregate now depends on both the Node test matrix and performance benchmarks, runs under `always()`, and fails unless both results are successful. The benchmark job and aggregate share the same event conditions, so the required check cannot remain green when benchmark execution fails.

I re-reviewed the full current diff and found no remaining correctness, workflow-semantics, test-quality, security, or structural concerns. The exact-head benchmark suite and required aggregate passed, the artifact contains all benchmark cases, and all exact-head CI is green.

PR #2311, opened 2026-08-13
fix: avoid TypeError in toFile for Responses without a URL

dfedoryshchev:

## Changes being requested

`toFile()` infers a filename for a `ResponseLike` input by calling `new URL(value.url)` directly, so `await toFile(new Response('contents'))` rejects with `TypeError: Invalid URL` - a `Response` that was constructed rather than returned by `fetch` has `url === ''`. A URL with no path segment, such as `https://example.com/`, yields a `File` named `''` instead. Both contradict the fallback to `unknown_file` that `toFile()` documents.

`getName()` in `src/internal/uploads.ts` already does this inference safely, `url` case included; `toFile()` imports it for its byte and stream path, and `createForm()` uses it for `Response` values. This swaps the inline parsing for that shared helper so those paths agree, and adds coverage for both cases.

## Additional context & links

- `npx vitest run --config vitest.config.mts tests/uploads.test.ts` -> 38 passed; both added cases fail on `main`.
- `npx tsc --noEmit` and `ultracite check` pass.

### DIFF
--- a/src/internal/to-file.ts
+++ b/src/internal/to-file.ts
@@ -148,7 +148,7 @@ export async function toFile(
 
   if (isResponseLike(value)) {
     const blob = await value.blob();
-    name ||= new URL(value.url).pathname.split(/[\\/]/).pop();
+    name ||= getName(value);
 
     const responseOptions =
       options?.type === undefined && blob.type ? { ...options, type: blob.type } : options;
--- a/tests/uploads.test.ts
+++ b/tests/uploads.test.ts
@@ -63,6 +63,18 @@ describe('toFile', () => {
     expect(file.name).toEqual('audio.mp3');
   });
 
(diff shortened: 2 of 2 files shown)


HAYDEN-OAI (commented):
Reviewed the full diff; this looks correct. `getName()` already owns safe URL parsing and returns `undefined` for both empty URLs and URL paths with no basename, allowing `makeFile()` to apply its documented `unknown_file` fallback. The new tests cover both regressions and preserve the response payload, while the existing tests retain coverage for normal URL-derived names and MIME handling. I found no issues.

HAYDEN-OAI (approved):
Approved: this correctly reuses existing filename inference, preserves the documented fallback, and matches current multipart upload behavior.

HAYDEN-OAI:
Thanks for the thoughtful fix and for contributing to the OpenAI Node SDK!

PR #2315, opened 2026-08-13
fix: prevent prototype event names from crashing SDK streams

HAYDEN-OAI:
## Summary

Harden all three hand-written SDK event-listener registries against `Object.prototype` event-name collisions. Incoming WebSocket messages control their `type`, so a message such as `{"type":"__proto__"}` previously escaped its transport callback with an uncaught exception.

## Impact

A server, reverse proxy, or other upstream WebSocket peer could trigger `TypeError: listeners.filter is not a function` with otherwise valid JSON. The exception bypassed SDK error listeners and could terminate a Node.js process or break a browser connection.

Affected paths include:

- stable and beta Realtime connections using both native/browser WebSockets and Node.js `ws`;
- stable and beta Responses WebSocket connections;
- direct listener registration, removal, and event dispatch on the SDK's hand-written `EventStream` and emitter helpers.

## Root cause

The core emitter, Realtime emitter, and event-stream helper initialized their listener dictionaries with `{}`. Looking up untrusted event names such as `__proto__`, `constructor`, `toString`, `hasOwnProperty`, and `valueOf` returned inherited objects or functions instead of listener arrays. Subsequent `.filter()`, `.push()`, or `.findIndex()` calls then threw; inherited functions could also make `_hasListener()` incorrectly return `true`.

## Solution

Initialize each of the three listener dictionaries with `Object.create(null)`. This preserves the existing mapped TypeScript event types and event/listener semantics while making all event names ordinary own keys. Unknown prototype-named messages continue to reach generic `event` listeners without crashing, and explicitly registered prototype-named listeners now support `on`, `once`, `off`, and normal dispatch.

Changes are limited to three hand-written emitter implementations and their existing hand-written tests. No generated SDK resources or generated tests are modified.

## Regression coverage

- Add table-driven unit coverage for all five inherited names across the core emitter, Realtime emitter, and event-stream helper.
- Verify unobserved dispatch, safe removal, listener detection, persistent registration, one-time registration, and listener removal.
- Replay malicious JSON messages through stable and beta browser/Node Realtime transports and stable/beta Responses WebSockets.
- Confirmed the new regressions produced **60 failures before the fix** and pass after the three-line production change.

## Validation

- `pnpm test:unit` — **66 files, 1,834 tests passed**.
- `pnpm exec vitest run --config vitest.config.mts tests/lib/core-event-emitter.test.ts tests/lib/eventEmitter.test.ts tests/lib/EventStream.test.ts tests/realtime-websocket.test.ts tests/lib/responsesWebSocket.test.ts --reporter=dot` — **5 files, 204 tests passed**.
- `pnpm lint` — passed.
- `pnpm exec tsc --noEmit` — passed.
- `git diff --check` — passed.
- Repository generated-file classifier — **8 changed hand-written files; 0 generated files**.

An isolated git worktree reused the existing dependency installation; pnpm commands were run with `pnpm_config_verify_deps_before_run=false` to account for the worktree's distinct workspace path.

### DIFF
--- a/src/core/EventEmitter.ts
+++ b/src/core/EventEmitter.ts
@@ -16,7 +16,7 @@ export type EventParameters<Events, EventType extends keyof Events> = Record<
 export class EventEmitter<EventTypes extends Record<string, (...args: any) => any>> {
   #listeners: {
     [Event in keyof EventTypes]?: EventListeners<EventTypes, Event>;
-  } = {};
+  } = Object.create(null);
 
   /**
    * Adds the listener function to the end of the listeners array for the event.
--- a/src/lib/EventEmitter.ts
+++ b/src/lib/EventEmitter.ts
@@ -16,7 +16,7 @@ export type EventParameters<Events, EventType extends keyof Events> = Record<
 export class EventEmitter<EventTypes extends Record<string, (...args: any) => any>> {
   #listeners: {
     [Event in keyof EventTypes]?: EventListeners<EventTypes, Event>;
-  } = {};
+  } = Object.create(null);
 
   /**
    * Adds the listener function to the end of the listeners array for the event.
--- a/src/lib/EventStream.ts
+++ b/src/lib/EventStream.ts
@@ -17,7 +17,7 @@ export class EventStream<EventTypes extends BaseEvents> {
 
   #listeners: {
     [Event in keyof EventTypes]?: EventListeners<EventTypes, Event>;
-  } = {};
+  } = Object.create(null);
   #abortListeners: { signal: AbortSignal; listener: () => void }[] = [];
 
   #ended = false;
--- a/tests/lib/EventStream.test.ts
+++ b/tests/lib/EventStream.test.ts
@@ -12,6 +12,10 @@ class TestStream extends EventStream<TestEvents> {
     this._emit('foo', value, index);
   }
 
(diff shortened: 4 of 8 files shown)


HAYDEN-OAI (commented):
Reviewed all eight modified files and found no correctness issues. Each changed file is handwritten (none has the canonical Castiron/Stainless banner), so this belongs in `sdk`. Using null-prototype listener registries fixes inherited-name lookups without changing listener ordering, duplicate registration, `once`, `off`, promise, or lifecycle behavior. The table-driven unit coverage exercises registration, detection, dispatch, and removal for all five inherited names across all three registries, and the transport tests cover stable/beta Responses plus browser/Node Realtime dispatch.

jbeckwith-oai (approved):
Reviewed the complete exact-head security fix and regression suite. Null-prototype dictionaries cover all three object-backed hand-written listener stores; the remaining WebSocket adapter registries already use `Map`. Listener ordering, `once`/`off`, `_hasListener`, typing, and public APIs are preserved. The tests exercise inherited property names through direct emitters and stable/beta browser and Node transports using the existing protocol fixtures. All applicable exact-head CI, Node 22/24/26, ecosystem, breaking-change, and CodeQL checks pass. No findings.

PR #2318, opened 2026-08-13
fix(security): prevent global Array.prototype pollution in chat completion streams

HAYDEN-OAI:
## Security impact

A malicious or compromised upstream Chat Completions streaming endpoint can send a string `"__proto__"` in either server-controlled index below. The existing array lookup resolves the JavaScript realm's shared `Array.prototype`, and the subsequent `Object.assign` copies attacker-controlled fields directly onto that prototype. Every array in the same Node.js process/realm then inherits those attacker-controlled properties.

This pull request addresses exactly these two confirmed process-global `Array.prototype` pollution paths in handwritten `ChatCompletionStream` code; it makes no claim about unrelated emitters, assistant streams, SSE paths, `Object.prototype`, or remote code execution.

### Sink 1: streamed choice index

```json
{
  "choices": [
    {
      "index": "__proto__",
      "delta": null,
      "finish_reason": null,
      "logprobs": null,
      "sdkPrototypePolluted": "owned"
    }
  ]
}
```

`snapshot.choices[index]` resolves `Array.prototype`; `Object.assign(choice, other)` writes `sdkPrototypePolluted` onto that shared prototype.

### Sink 2: streamed tool-call index

```json
{
  "choices": [
    {
      "index": 0,
      "finish_reason": null,
      "logprobs": null,
      "delta": {
        "role": "assistant",
        "tool_calls": [
          {
            "index": "__proto__",
            "sdkNestedPolluted": "owned"
          }
        ]
      }
    }
  ]
}
```

`toolCallSnapshots[index] ??= {}` resolves `Array.prototype`; `Object.assign(tool_call, rest)` writes `sdkNestedPolluted` onto that shared prototype.

## Fix and compatibility

- Validate both indices **before** the associated array lookup or mutation and throw a distinct, clear `OpenAIError` for invalid choice versus tool-call indices.
- Require `Number.isSafeInteger(index) && index >= 0 && index < 2 ** 32 - 1`. The upper bound is JavaScript's real array-element boundary, not an arbitrary SDK cap: larger values become ordinary named properties instead of array elements.
- Preserve valid numeric indices, existing multi-tool accumulation, and valid out-of-order streamed choices.
- Limit the change to handwritten `src/lib/ChatCompletionStream.ts` and handwritten `tests/lib/ChatCompletionStream.test.ts`; no Stainless/Castiron-generated source, tests, or snapshots are modified.

## Verification

- Regression-first: both exploit tests failed against current `origin/main`, each observing `"owned"` through a newly created array. Each regression removes **only its own unique injected prototype property** in `finally`.
- Focused handwritten suite: `./scripts/test tests/lib/ChatCompletionStream.test.ts` — **58 passed**.
- Full handwritten suite: `COREPACK_HOME=/tmp/openai-node-corepack-mss4bko6 pnpm_config_verify_deps_before_run=false ./node_modules/.bin/vitest run --config vitest.config.mts` — **66 files, 1,875 passed**.
- Generated API suite inside the full repository run: **82 suites / 559 tests passed; 1 suite / 1 test skipped; 12 snapshots passed**, against the repository's real Steady OpenAPI mock server.
- The combined `./scripts/test` wrapper returns nonzero **after every handwritten and generated test passes** because its pre-existing shutdown trap races an already-exited mock-server PID: `./scripts/test: line 19: kill: (...) - No such process`. No test failure is hidden by this note.
- `./scripts/lint` — passed across **603 files**.
- `./scripts/build` — CommonJS and ESM builds passed, including package and Bedrock-provider import smoke tests.
- Table-driven regressions cover `__proto__`, `constructor`, `toString`, numeric and nonnumeric strings, negative/fractional/non-finite numbers, the array-index boundary, and unsafe integers for **both** sinks.
- Existing tool-call tests preserve indices `0` and `1`; a new replay test confirms valid choice indices arriving out of order still finalize correctly.


src/lib/ChatCompletionStream.ts
```diff
@@ -617,6 +617,10 @@ export class ChatCompletionStream<ParsedT = null>
     }
 
     for (const { delta, finish_reason, index, logprobs = null, ...other } of chunk.choices) {
+      if (!Number.isSafeInteger(index) || index < 0 || index >= 2 ** 32 - 1) {
```
HAYDEN-OAI: The array-element boundary is still far too large for untrusted indices. An upstream can send `index: 4294967294` (or merely a few hundred million); assigning `snapshot.choices[index]` sets the array length to billions, and finalization/consumer operations such as `map`, serialization, or iteration can consume extreme CPU or throw due to the resulting sparse array. The same applies to `toolCallSnapshots`. Please enforce a practical protocol limit (ideally derived from the requested choice count / accumulated contiguous entries, with a small defensive ceiling) for both indices and add regressions for large but technically valid array indices.

HAYDEN-OAI (commented):
Reviewed the complete two-file security change. Both files are handwritten (no canonical Castiron/Stainless banner), so I added `sdk`. The pre-index validation closes the named-property prototype-pollution paths, but the accepted numeric range still leaves a remotely triggerable sparse-array denial of service; noted inline.

jbeckwith-oai (changes requested):
Two validated findings remain.

1. Medium — `src/lib/ChatCompletionStream.ts:619-627` and `:742-747` still accept `2 ** 32 - 2`. That value is a real JavaScript array index, so one tiny upstream chunk immediately sets `choices` or `toolCallSnapshots.length` to `4_294_967_295`. Finalization later calls `.map()` on those arrays (and choice event state also uses indexed arrays), forcing a length-driven traversal and creating an effective CPU denial of service. I reproduced the maximum-length sparse array locally; its map did not finish within 30 seconds. Please use the request's `n` as the choice bound where available and define a documented protocol-sized bound for replayed choices/tool calls, or otherwise accumulate untrusted indices without length-amplifying arrays. Preserve valid out-of-order indices and add final-accepted/first-rejected boundary tests.

2. Low/maintainability — the cohesive 203-line security suite is added to the already 1,376-line `tests/lib/ChatCompletionStream.test.ts`, taking it to 1,579 lines, and duplicates the invalid-index table plus most setup. Please move only the new index-validation suite into a focused discovered test module and share a small choice/tool-call case helper.

The prototype-pollution fix itself is correctly placed before array lookup/mutation. Focused tests pass 58/58, TypeScript/lint/diff hygiene pass, and exact-head CI has no failures (one ecosystem duplicate remains in progress).

src/lib/ChatCompletionStream.ts
HAYDEN-OAI: Addressed in e29a713a9b11d9d95445b958414bac996388dc66.

- Choices are validated as safe integer indices below `min(request.n, 128)` when a valid request count is available, with a hard replay/fallback ceiling of 128. The repository's `api_reference/openapi.transformed.yml` explicitly specifies chat-completion `n` as 1–128.
- Tool-call indices independently use the same conservative defensive ceiling: accepted `0..127`; rejected `128` and above. Both guards execute before snapshot array indexing, chunk emission, or indexed choice-event-state allocation.
- Regression coverage explicitly accepts `127`, rejects `128`, `100_000_000`, and `4_294_967_294` without constructing large arrays, validates request-derived and hostile `n`, protects replayed choices/tool calls, preserves out-of-order choices/tool calls, and retains both prototype-pollution checks.
- Moved the entire index-validation/security suite out of the oversized legacy test into discovered `tests/lib/chat-completion-stream.index-validation.test.ts`, using one shared choice/tool-call fixture and invalid-index table.

Verification: focused security suite **54/54**; stream/function-runner coverage **105/105**; complete handwritten suite **1,898/1,898** across 67 files; full TypeScript check, repository lint/format/generated-file checks, and `git diff --check` all passed.

HAYDEN-OAI (commented):
Re-reviewed the complete updated two-file diff. The new 128-entry hard ceiling prevents attacker-controlled sparse arrays, and valid request `n` values further tighten choice indices before snapshot indexing or chunk emission. Replay paths retain the defensive ceiling, while tool-call indices are independently bounded. The dedicated suite covers prototype names, large numeric indices, boundary values, hostile request counts, replay, and valid out-of-order accumulation. The reported denial-of-service issue is resolved; no new findings. Both CI workflows pass.

jbeckwith-oai (approved):
Both prior blockers are resolved. Choice indices are now bounded before array/event-state access by valid request `n` (capped at the schema maximum of 128), with the same conservative ceiling for replayed/no-parameter streams. Tool-call indices are independently capped at 128 before indexing. This closes prototype/coercion and sparse-array amplification paths while preserving supported out-of-order delivery and existing error semantics.

The security regressions are now isolated in a focused 245-line module and cover hostile names/numbers, request-aware bounds, replay, pre-emission failure, and out-of-order choices/tool calls. Focused tests pass 54/54 and all exact-head CI is green. The broad test file no longer grows; the production class was already over 1,000 lines and receives only the canonical validation at its two sinks.

PR #2322, opened 2026-08-14
fix(security): prevent multipart upload content-type header injection

HAYDEN-OAI:
## Summary

- Reject ASCII C0 control characters (including CR, LF, NUL, and TAB), DEL, and runtime non-string MIME values both when creating a streaming upload and when resolving every multipart upload part.
- Resolve streaming-file, `File`/`Blob`, and response content types exactly once and validate them before emitting that upload part's boundary or headers.
- Preserve MIME spelling, case, parameters, printable non-ASCII/spacing, and the existing `application/octet-stream` default.

## Security impact

The handwritten streaming multipart encoder previously interpolated caller-controlled upload MIME metadata directly into a `Content-Type` part header. A MIME value such as `text/plain\r\nContent-Disposition: form-data; name="purpose"` injected a second `Content-Disposition` header without requiring knowledge of the multipart boundary.

Using Node's WHATWG `Response(...).formData()` parser on the vulnerable baseline, the intended upload disappeared and its bytes became an attacker-controlled text form parameter:

```text
upload: null
purpose: ["sensitive upload bytes", "assistants"]
```

Factory-only validation is insufficient because streaming-file objects are mutable and mixed streaming forms also serialize named `File`/`Blob` parts whose `.type` getters may be overridden. The central sink guard covers both cases, rejects coercion-based object payloads, snapshots changing getters once, and rejects invalid upload parts before even their boundary bytes are emitted.

## Regression coverage

- Seven synchronous control-character cases: CRLF header injection, bare CR, LF, NUL, TAB, U+001F, and DEL.
- Actual WHATWG multipart parsing to demonstrate and prevent duplicate injected form parameters.
- Mutated streaming-file MIME metadata and malicious non-string `toString()` payloads.
- Hostile and changing `File.type` getters in mixed streaming forms.
- No multipart-boundary output for an invalid first upload part.
- Preserved MIME case, parameters, whitespace, Unicode, and the default content type.

## Verification

- Baseline security regression run: 10 expected failures; standalone WHATWG parser reproduced `purpose=["sensitive upload bytes", "assistants"]`.
- `./scripts/test tests/internal/uploads-branches.test.ts --no-cache --update=none` — 59 passed.
- `COREPACK_HOME=/tmp/openai-node-corepack-review-9f4a72 pnpm_config_verify_deps_before_run=false OPENAI_TEST_SUITE=unit ./scripts/test --no-cache --update=none` — 66 files, 1,862 tests passed.
- `./scripts/lint` — passed.
- `./scripts/build` — passed, including CommonJS/ESM package smoke checks.
- `./node_modules/typescript-4-9/bin/tsc --project dist/src/tsconfig.json --noEmit` — passed.
- `./node_modules/typescript/bin/tsc --project dist/src/tsconfig.json --noEmit` — passed.
- The generated-suite wrapper could not start its Steady mock through `npm exec` because the devbox npm cache is read-only; generated tests were not run locally.

### DIFF
--- a/src/internal/uploads.ts
+++ b/src/internal/uploads.ts
@@ -53,7 +53,7 @@ export interface StreamingFile {
  * @param data Async-iterable or readable-stream chunks containing text, binary data, or blobs.
  * @param name Non-empty filename sent in the multipart request.
  * @param options Optional MIME type for the streaming file.
- * @throws {TypeError} If `name` is empty.
+ * @throws {TypeError} If `name` is empty or the content type contains control characters.
  */
 export function toStreamingFile(
   data: StreamingFileInput,
@@ -64,11 +64,16 @@ export function toStreamingFile(
     throw new TypeError('toStreamingFile requires a non-empty file name');
   }
 
+  const type = options?.type;
+  if (type) {
+    validateStreamingFileType(type);
+  }
+
   return {
     [brand_privateStreamingFile]: true,
     data,
     name,
-    ...(options?.type ? { type: options.type } : {}),
+    ...(type ? { type } : {}),
   };
 }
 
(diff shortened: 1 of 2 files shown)


jbeckwith-oai (approved):

HAYDEN-OAI (commented):
Reviewed the complete two-file security change. Both files are handwritten and lack the canonical Castiron/Stainless banner, so I added `sdk`. Content types are snapshotted and validated at both construction and the central serialization sink before any multipart boundary/header bytes are emitted; the guard rejects runtime non-strings plus C0/DEL controls while preserving valid spelling and defaults. Coverage includes mutated streaming metadata, hostile File getters, coercion objects, actual multipart parsing, and pre-boundary failure. No correctness issues found; both CI workflows pass.

PR #2323, opened 2026-08-14
fix(responses): validate streaming event indices before array access

HAYDEN-OAI:
## Summary

Harden handwritten Responses streaming accumulation against remotely controlled event indices without changing generated SDK resources, clients, or public types.

- Validate `output_index`, `content_index`, and `summary_index` before every existing-element lookup, replacement, or append.
- Require existing references to be safe nonnegative integers below the current collection length and to resolve to an own array element.
- Allow additions only at an existing own element or exactly the current collection length, preserving replay behavior while rejecting sparse gaps and inherited keys.
- Bound `annotation_index` to an existing own annotation or a single contiguous append, and reject inherited numeric append setters before assignment.
- Preserve existing `missing output at index …` / `missing content at index …` errors and avoid initializing missing reasoning content when its index is invalid.
- Reuse the existing ES2020-compatible `hasOwn` helper; no exported API, generated file, ChatCompletion streaming, generic SSE, Azure, tool runner, or Bedrock code changes.

## Security impact

Streaming event index fields arrive from the remote response stream and cannot be trusted merely because their generated TypeScript types say `number`.

Previously, an event such as `response.output_item.done` with `output_index: "__proto__"` caused `getOutput` to resolve `Array.prototype` as though it were an existing output. The subsequent assignment replaced the snapshot array's prototype, corrupted local accumulator state, and immediately failed with errors such as `rsp.output is not iterable` or `.push is not a function`. Equivalent inherited-key attacks reached completed message content, reasoning content, reasoning summaries, and annotations; inherited sparse slots and numeric append setters could also be treated as legitimate state.

Separately, `annotation_index: 4294967294` expanded an empty annotations array to length **4,294,967,295** in one assignment, creating an immediate CPU/memory denial-of-service vector. Merely restricting values to JavaScript's legal array-index range does not address this: the fix derives a practical protocol bound directly from the collection's current length and accepts only an existing own slot or one contiguous append. Negative, fractional, nonfinite, unsafe-integer, string, inherited, and out-of-range indices now fail before lookup, allocation, assignment, initialization, or append.

## Regression evidence

- The original handwritten focused suite passed **50/50** tests before new regressions were added.
- Against the unchanged vulnerable accumulator, the new exploit regressions produced **23 failing cases**, including the exact prototype-replacement crash and oversized annotation insertion.
- An independent bypass review then reproduced an inherited numeric annotation setter firing at the append position; a separate regression failed first and now proves the setter is never called.
- The final focused suite passes **99/99** cases, covering inherited names, array-prototype preservation, malformed and unsafe numeric indices, large-but-valid JavaScript indices, sparse inherited values, oversized annotations without expanding array length, inherited append setters, compatible contiguous insertion, and output/content/summary/annotation replay.

## Verification

```sh
./node_modules/.bin/vitest run --config vitest.config.mts tests/lib/ResponseAccumulator.events.test.ts
# 99 passed

COREPACK_HOME=/tmp/openai-node-responses-corepack XDG_CACHE_HOME=/tmp/openai-node-responses-xdg pnpm test:unit --update=none
# 66 test files, 1,893 tests passed; snapshots unchanged

./node_modules/.bin/tsc --noEmit
./scripts/lint
COREPACK_HOME=/tmp/openai-node-responses-corepack XDG_CACHE_HOME=/tmp/openai-node-responses-xdg pnpm build
# TypeScript, repository formatting/lint, CommonJS build, ESM build, and package smoke imports all pass
```

This is distinct from #2318, which only addresses `ChatCompletionStream` indices and does not touch Responses accumulation.

### DIFF
--- a/src/lib/responses/ResponseAccumulator.ts
+++ b/src/lib/responses/ResponseAccumulator.ts
@@ -1,5 +1,6 @@
 import type { Response, ResponseOutputText, ResponseStreamEvent } from '../../resources/responses/responses';
 import { OpenAIError } from '../../error';
+import { hasOwn } from '../../internal/utils';
 import { addOutputText } from '../ResponsesParser';
 
 /** A transport keepalive event that leaves the accumulated response unchanged. */
@@ -32,6 +33,7 @@ export function accumulateResponse(
 
   switch (event.type) {
     case 'response.output_item.added': {
+      validateArrayIndex(snapshot.output, event.output_index, 'output', true);
       snapshot.output.push(structuredClone(event.item));
       if (event.item.type === 'message') {
         addOutputText(snapshot);
@@ -51,15 +53,18 @@ export function accumulateResponse(
       const type = output.type;
       const part = event.part;
       if (type === 'message' && part.type !== 'reasoning_text') {
+        validateArrayIndex(output.content, event.content_index, 'content', true);
         output.content.push(structuredClone(part));
         if (part.type === 'output_text') {
           addOutputText(snapshot);
         }
       } else if (type === 'reasoning' && part.type === 'reasoning_text') {
+        const content = output.content ?? [];
+        validateArrayIndex(content, event.content_index, 'content', true);
         if (!output.content) {
-          output.content = [];
+          output.content = content;
         }
-        output.content.push(structuredClone(part));
+        content.push(structuredClone(part));
       }
       break;
     }
@@ -113,6 +118,7 @@ export function accumulateResponse(
         if (content.type !== 'output_text') {
           throw new OpenAIError(`expected content to be 'output_text', got ${content.type}`);
         }
+        validateArrayIndex(content.annotations, event.annotation_index, 'annotation', true);
         content.annotations[event.annotation_index] = structuredClone(
           event.annotation,
         ) as ResponseOutputText['annotations'][number];
@@ -186,6 +192,7 @@ export function accumulateResponse(
     case 'response.reasoning_summary_part.added': {
       const output = getOutput(snapshot, event.output_index);
       if (output.type === 'reasoning') {
+        validateArrayIndex(output.summary, event.summary_index, 'content', true);
         output.summary.push(structuredClone(event.part));
       }
       break;
@@ -400,6 +407,7 @@ function cloneResponse(response: Response): Response {
 }
 
 function getOutput(snapshot: Response, outputIndex: number): Response['output'][number] {
+  validateArrayIndex(snapshot.output, outputIndex, 'output');
   const output = snapshot.output[outputIndex];
   if (!output) {
     throw new OpenAIError(`missing output at index ${outputIndex}`);
@@ -408,11 +416,28 @@ function getOutput(snapshot: Response, outputIndex: number): Response['output'][
 }
 
 function getContent<T>(content: T[], contentIndex: number): T {
+  validateArrayIndex(content, contentIndex, 'content');
   const part = content[contentIndex];
   if (!part) {
     throw new OpenAIError(`missing content at index ${contentIndex}`);
   }
   return part;
+}
+
+function validateArrayIndex(
+  collection: readonly unknown[],
+  index: number,
+  kind: 'output' | 'content' | 'annotation',
+  allowAppend = false,
+): void {
+  if (
+    !Number.isSafeInteger(index) ||
+    index < 0 ||
+    index > collection.length ||
+    (index === collection.length ? !allowAppend || index in collection : !hasOwn(collection, index))
+  ) {
+    throw new OpenAIError(`missing ${kind} at index ${index}`);
+  }
 }
 
 function assertNever(value: never): never {
--- a/src/lib/responses/ResponseAccumulator.ts
+++ b/src/lib/responses/ResponseAccumulator.ts
@@ -33,7 +33,7 @@ export function accumulateResponse(
 
   switch (event.type) {
     case 'response.output_item.added': {
-      validateArrayIndex(snapshot.output, event.output_index, 'output', true);
+      validateArrayAppend(snapshot.output, event.output_index, 'output');
       snapshot.output.push(structuredClone(event.item));
       if (event.item.type === 'message') {
         addOutputText(snapshot);
@@ -53,14 +53,14 @@ export function accumulateResponse(
       const type = output.type;
       const part = event.part;
       if (type === 'message' && part.type !== 'reasoning_text') {
-        validateArrayIndex(output.content, event.content_index, 'content', true);
+        validateArrayAppend(output.content, event.content_index, 'content');
         output.content.push(structuredClone(part));
         if (part.type === 'output_text') {
           addOutputText(snapshot);
         }
       } else if (type === 'reasoning' && part.type === 'reasoning_text') {
         const content = output.content ?? [];
-        validateArrayIndex(content, event.content_index, 'content', true);
+        validateArrayAppend(content, event.content_index, 'content');
         if (!output.content) {
           output.content = content;
         }
@@ -192,7 +192,7 @@ export function accumulateResponse(
     case 'response.reasoning_summary_part.added': {
       const output = getOutput(snapshot, event.output_index);
       if (output.type === 'reasoning') {
-        validateArrayIndex(output.summary, event.summary_index, 'content', true);
+        validateArrayAppend(output.summary, event.summary_index, 'content');
         output.summary.push(structuredClone(event.part));
       }
       break;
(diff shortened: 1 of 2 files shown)


src/lib/responses/ResponseAccumulator.ts
```diff
@@ -32,6 +33,7 @@ export function accumulateResponse(
 
   switch (event.type) {
     case 'response.output_item.added': {
+      validateArrayIndex(snapshot.output, event.output_index, 'output', true);
       snapshot.output.push(structuredClone(event.item));
       if (event.item.type === 'message') {
         addOutputText(snapshot);
```
HAYDEN-OAI, line 39: For replayed add events, this validates `event.output_index` but mutates a different index via `push()` (`snapshot.output.length`). If the event references an existing own slot (e.g. index 0 in a length-1 array) and the array prototype has a setter at the append position (index 1), validation succeeds and `push()` invokes that inherited setter instead of creating an own element. The same bypass exists for `output.content.push()` and `output.summary.push()`. Please validate that the actual append position is free of inherited properties before every push (or perform an own-data-property append), and add inherited-setter regressions for output/content/summary replay just like the annotation regression.

HAYDEN-OAI (commented):
Reviewed the complete two-file security change. Both files are handwritten and lack the canonical Castiron/Stainless banner, so I added `sdk`. Existing-element validation and contiguous-index bounds close the named and sparse lookup paths, but replayed add events retain an inherited-setter bypass at the actual push position; noted inline.

jbeckwith-oai (changes requested):
The new validation still has one medium-severity bypass for replayed push-based `.added` events. When an event names an existing own slot, `validateArrayIndex(..., true)` validates that declared index, but `push()` mutates `collection.length` instead. For example, `output_index: 0` on a length-1 array passes validation and appends at index 1. Besides desynchronizing protocol indices from stored positions, this skips the inherited-property check at the actual mutation index: if the array prototype has a numeric setter at index 1, `push()` invokes it, advances `length`, and creates no own element. The same path exists for output, content, and reasoning-summary replays, so the inherited-setter class this PR is intended to close remains reachable.

Please validate the actual append position for each push-based event (strict `index === collection.length` is simplest), or handle an existing replay idempotently at its declared slot without pushing elsewhere. Add output/content/summary replay regressions with an inherited setter at the real append position. Annotation replay can retain existing-or-append behavior because it assigns directly to `annotation_index`.

Everything else in the hardening is sound: existing lookups and contiguous appends reject malformed, sparse, inherited, and oversized indices before mutation; focused tests pass 99/99; all exact-head CI is green; and the 445-line source / 910-line focused test stay below the size threshold.

src/lib/responses/ResponseAccumulator.ts
HAYDEN-OAI: Fixed at exact head `9e963a204dcc5edf789bc03f9156a32928be8c0d`.

All push-based `.added` handlers now call `validateArrayAppend`, which requires the declared output/content/summary index to equal the collection's actual `length`, then rejects any inherited property at that append position before `push()`. Replayed existing indices therefore fail instead of appending at a different location. The valid accumulation regression now uses declared contiguous index 1, and a table-driven regression installs an inherited setter at index 1 for output, message content, and reasoning summary; each path throws before the setter runs, retains length 1, and leaves index 1 non-own.

The first CI pass exposed only two nested-ternary lint errors in that new table setup; `9e963a2` rewrote the setup with explicit branches. Exact-head breaking-change/lint workflow 31757815465 now passes, OkTest is running, and the main workflow is still running.

HAYDEN-OAI (commented):
Re-reviewed the complete updated two-file diff. Add events now require the declared index to equal the actual append position, and the shared validator rejects inherited properties at that position before `push()`. The new output/content/summary setter regressions cover the reported bypass, while contiguous accumulation remains covered. The inherited-setter finding is resolved; no new correctness issues found. Breaking-change/lint CI passes and the main CI workflow is still running.

jbeckwith-oai (changes requested):
The production fix now closes the prior blocker: push-backed output/content/summary events require the declared index to equal the actual append position, then validate that slot for inherited properties before mutation.

One low but important regression-test gap remains. The new inherited-setter table installs a setter at slot 1 on a length-1 collection and also declares event index 1. The previous vulnerable validator already rejected that case because `1 in collection`, so this test passes unchanged against the old code and does not lock in the reported replay bypass. The actual exploit declared existing owned index 0, passed old validation, and then `push()` wrote slot 1 and invoked the inherited setter.

Please keep the setter at slot 1 but send index 0 for output/content/summary, then assert the current implementation throws `missing … at index 0`, never calls the setter, and preserves length/ownership. That makes the regression fail on the vulnerable head and prove the new strict append-index check. Focused 102/102 and all exact-head CI are green; source/test remain cohesive at 456/972 lines.

HAYDEN-OAI:
Addressed the requested regression correction at exact head `1247d40e1ee89324cde569d2317d989d13ac34c7`.

The inherited setter remains at actual push slot 1 on each length-1 output/content/summary collection, but every event now declares existing owned index 0. The current strict append-index guard throws `missing output/content at index 0` before `push()`, the setter is never called, length stays 1, and slot 1 remains non-own. This exactly reproduces the prior vulnerable control flow: the old `allowAppend` validator accepted owned index 0 and then `push()` hit inherited slot 1.

Exact-head CI is restarting.

tests/lib/ResponseAccumulator.events.test.ts
```diff
@@ -358,6 +440,452 @@ describe('ResponseAccumulator lifecycle and error handling', () => {
...
+        inheritedSetterCalled = true;
+      },
+    });
+    Object.setPrototypeOf(collection, collectionPrototype);
+
+    expect(() =>
+      applyEvent(snapshot, {
+        type,
+        output_index: 0,
+        [indexField]: 0,
```
HAYDEN-OAI: This change makes the event declare index 0 while every collection already has length 1 and the inherited setter is installed at index 1. `validateArrayAppend` now rejects immediately on `index !== collection.length`, so the test never evaluates `index in collection` at the setter-bearing append position and would still pass if that security check were removed. Please restore `[indexField]: 1` and the expected `... index 1` error so this remains a regression for the actual inherited-setter bypass rather than only the noncontiguous-index guard.

HAYDEN-OAI (commented):
Re-reviewed the complete updated two-file diff. Production validation is unchanged and remains correct, but the latest test-only edit no longer exercises the inherited-setter branch it claims to cover; noted inline.

tests/lib/ResponseAccumulator.events.test.ts
HAYDEN-OAI: Fixed at exact head `b41b2b4f175c4a7b6d07659c8c2cbd8636df519d` by covering both security properties in one compact six-case table.

For each of output/content/summary, one case declares contiguous append index 1 with an inherited setter at slot 1, proving `index in collection` rejects the setter-bearing mutation slot. A second case declares replayed existing owned index 0 while the setter remains at actual push slot 1, reproducing the original bypass and proving the strict `index === length` guard prevents index desynchronization. All six assert the setter never runs, length remains 1, and slot 1 remains non-own.

Exact-head CI is restarting.

tests/lib/ResponseAccumulator.events.test.ts
```diff
@@ -358,6 +440,458 @@ describe('ResponseAccumulator lifecycle and error handling', () => {
     ).toThrow('missing content at index 1');
   });
 
+  test.each(['__proto__', 'constructor', 'push'])(
```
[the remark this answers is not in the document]
HAYDEN-OAI: The security matrix is intentionally cohesive and already consolidated: malformed existing indices use `test.each`, append/inherited cases share kind/type/index-field tables, and the latest setter coverage combines six output/content/summary scenarios in one fixture/assertion body. The repository's human owner review measured the complete focused file at 972 lines and explicitly found it below the size threshold; the production file is 456 lines. Further generic fixture abstraction would hide which collection and mutation slot each security invariant applies to without reducing behavioral cases. No duplicated production setup or exported helper/API is introduced.

I did fix the exact-head formatting failure separately at `f9b20b6242626e66a7be8d0858cfbaaf1c044e32`; CI is restarting.

jbeckwith-oai (approved):
Re-reviewed the complete security hardening at this exact head. The production fix requires the declared output/content/summary index to equal the actual contiguous append position and validates that slot before mutation. The corrected index-0 replay cases now reproduce and kill the prior inherited-setter bypass, while index-1 cases cover genuine appends; all six tests have distinct names.

Focused tests pass 105/105. Formatting, lint, breaking-change detection, build, CodeQL, Node 22/24/26, and the test matrix are green; ecosystem jobs are still running. Source/test remain cohesive at 456/974 lines with no threshold or structural issue.

HAYDEN-OAI (commented):
Re-reviewed the complete two-file diff at this exact head. The append-position validation and inherited-setter regressions now cover both declared contiguous appends and replayed existing indices across output, content, and summary; the later test compaction preserves all six cases. Both exact-head CI workflows are green. The earlier finding is resolved, and I found no new actionable issues.

PR #2325, opened 2026-08-14
fix(realtime): prevent Azure API-key disclosure across WebSocket redirects

HAYDEN-OAI:
## Security impact

**Severity: HIGH — reusable Realtime credentials could cross an attacker-controlled WebSocket redirect.**

Stable and beta Node Realtime transports place Azure static credentials in the custom `api-key` handshake header. With `options.followRedirects: true`, pinned `ws@8.21.1` does not strip custom API-key headers, so a redirect from an Azure endpoint, gateway, proxy, or compromised intermediary forwards the complete secret.

Bearer credentials also had a separate disclosure path: `ws` normally strips cross-origin `Authorization`, but registering its public `redirect` listener disables that built-in stripping. A caller that enabled redirects and registered the listener could therefore forward the generated bearer token to the redirected origin.

Regression-first live-server tests reproduced all six stable/beta × HTTP 302/307/308 Azure static-key disclosures on unmodified `main`. Independent review also reproduced the bearer-listener attack with the redirected origin receiving `Bearer SECRET`.

This is the WebSocket counterpart of the merged Azure HTTP redirect fix in #2320. Upstream [websockets/ws#2013](https://github.com/websockets/ws/pull/2013) intentionally strips standard sensitive headers while leaving custom credentials and listener behavior outside that guarantee.

## Fix

- In both handwritten stable and beta Node Realtime constructors, inspect the **final outgoing headers** after caller options, authentication, and beta headers have been merged.
- Detect `Authorization`, `Cookie`, and API-key-like names case-insensitively; the API-key predicate covers `api-key`, `API-Key`, `x-api-key`, `api_key`, `ApiKey`, and related variants.
- Force `followRedirects: false` **after** spreading caller-provided `ws` options whenever a sensitive header is present. Explicit opt-in, authentication type, header casing, option merge order, and caller-installed redirect listeners cannot bypass the fail-closed policy.
- Preserve all unrelated WebSocket options and the existing generated-Authorization precedence.
- Keep the cohesive policy and real loopback coverage in `tests/realtime-websocket-redirects.test.ts`; the broad `tests/realtime-websocket.test.ts` remains at its pre-change size boundary.
- Change handwritten files only. Generated Responses WebSocket transports are untouched.

### Compatibility tradeoff

Authenticated Node Realtime handshakes now intentionally reject every redirect, including same-origin redirects. Callers that previously opted into redirects must connect directly to the final endpoint. This is required because custom credential headers and public redirect listeners make selective cross-origin stripping insufficient as a security boundary.

## Verification

- Vulnerable static-key baseline: six stable/beta × 302/307/308 live attacks disclosed `AZURE_STATIC_SECRET` before the fix.
- Vulnerable bearer baseline: a cross-origin 302 plus a public `redirect` listener disclosed `Bearer SECRET` before the expanded fix.
- Exact head `c6798c0272c3460649a861d2d47d285d541e5468`: focused redirect-security plus broad Realtime WebSocket suites passed **132/132** in independent review.
- Exact-head CI workflows [31766726732](https://github.com/openai/openai-node/actions/runs/31766726732) and [31766726763](https://github.com/openai/openai-node/actions/runs/31766726763) passed, covering formatting, lint, TypeScript, breaking-change detection, examples, and ecosystem tests.
- Independent exact-head review approved the final-header policy, constructable mocks, Authorization-precedence case, loopback disclosure tests, extraction, and changed-file size boundaries with no actionable findings.

- [x] I understand that most of this repository is generated; this PR changes only independently verified handwritten files.


HAYDEN-OAI (commented):
Reviewed the complete stable, beta, and integration-test diff at this head. The redirect override is applied after caller options and after the final header merge, covers case/separator/prefix variants of API-key headers, and leaves bearer-only redirect behavior unchanged. The real two-origin tests cover both transports and the relevant redirect statuses without introducing a redirect listener. I found no actionable issues.

jbeckwith-oai (changes requested):
Two blocking issues remain at this exact head:

1. **High — bearer credentials can still cross redirects.** The new policy forces `followRedirects: false` only for API-key-like headers, but every bearer connection carries `Authorization`. In pinned `ws@8.21.1`, registering the public/documented `redirect` listener disables its cross-origin Authorization stripping. Both independent passes reproduced a cross-origin 302 destination receiving `Bearer SECRET` with `followRedirects: true` plus that listener. Please disable redirects whenever the final authenticated headers contain Authorization/Cookie/API-key credentials (or unconditionally for authenticated Realtime connections), and add stable/beta live-server bearer regressions with a redirect listener proving the destination receives no request.

2. **Thermo boundary — `tests/realtime-websocket.test.ts` grows from 812 to 1,002 lines.** Please extract the cohesive redirect policy/live-server suite into a focused `realtime-websocket-redirects.test.ts` module.

The static Azure API-key fix and real-server coverage are otherwise sound; focused tests and all exact-head CI are green.

HAYDEN-OAI:
Addressed both requested blockers at exact head `6177152f99e775f284cef94097d3344c4c7f0573`:

- Stable and beta Node transports now force `followRedirects: false` after the final header merge whenever the outgoing headers contain case-insensitive `Authorization`, `Cookie`, or an API-key-like name. A caller-registered public `redirect` listener therefore cannot re-enable the pinned `ws` cross-origin bearer disclosure path.
- The new stable/beta live-server bearer regression explicitly registers that listener, enables redirects, and proves the source receives `Bearer BEARER_SECRET` while the attacker destination receives no request. The existing six stable/beta × 302/307/308 Azure API-key attacks remain covered.
- Redirect policy and live-server coverage moved to focused `tests/realtime-websocket-redirects.test.ts`; `tests/realtime-websocket.test.ts` is restored to its pre-change size boundary.
- Exact-head workflow 31766326850 is green for formatting, lint, TypeScript, and breaking-change detection. The separate ecosystem workflow is still running.

jbeckwith-oai (changes requested):
The production redirect hardening and test-file extraction now look sound, but the new security suite does not execute: all WebSocket mock implementations in `tests/realtime-websocket-redirects.test.ts` are arrow functions, while stable and beta construct them with `new WS.WebSocket(...)`. Exact-head Vitest fails all 40 focused cases with `TypeError: ... is not a constructor`, so neither the option assertions nor the live loopback credential-disclosure checks run; Node 22/24/26 and the aggregate matrix are red. Please use constructable named functions or small classes for the capture mock and both one-shot real-WebSocket implementations (the existing broad realtime test demonstrates the pattern), without adding a lint suppression, then rerun the focused suite and matrix.

HAYDEN-OAI:
Follow-up at exact head `a605e27262ca84684471d154ad5089c75fb16e75`: the focused suite now uses top-level constructable named `CapturingWebSocket` and `ActualWebSocket` functions. The mock is reset to the capture constructor before each case, while each live-server case installs the constructable real-WebSocket bridge for one call. No lint suppression or cast was added. Exact-head workflow 31766590568 passes formatting, lint, TypeScript, and breaking-change detection.

jbeckwith-oai (changes requested):
The redirect implementation, constructable mocks, real loopback coverage, lint, and test-file extraction now look sound. One test-only blocker remains: in the sensitive-header table, the exact-cased `Authorization` row expects `custom-secret`, but the pre-existing final-header merge intentionally assigns generated `Authorization: Bearer test-key` after custom headers. That leaves the focused suite at 38/40 (one failure in stable and beta) and makes Node 22/24/26 plus the aggregate matrix red. Please give the exact `Authorization` row the bearer expectation (or split it into its own assertion); keep the other header rows expecting `custom-secret`, and do not change production authentication precedence.

HAYDEN-OAI:
Fixed the remaining exact-cased `Authorization` assertion at head `c6798c0272c3460649a861d2d47d285d541e5468`. That case is now separate and verifies the existing generated `Bearer test-key` precedence together with `followRedirects: false`; lowercase `authorization` and both Cookie casing rows continue to assert their caller-provided values. Production authentication precedence is unchanged. Exact-head workflow 31766726732 is green for format, lint, TypeScript, and breaking checks.

jbeckwith-oai (approved):
Approved after exact-head rereview. The final-header policy now forces authenticated Realtime connections to `followRedirects: false` for Authorization, Cookie, and API-key-like headers in both stable and beta. Constructable mocks exercise the real constructor path; the dedicated Authorization-precedence case is correct; and the real loopback tests prove static API keys and bearer tokens are not disclosed across redirects, including when a redirect listener is registered. Focused plus broad suites pass 132/132, all completed exact-head CI is green, and the extracted tests keep every changed file below the 1,000-line threshold.

PR #2329, opened 2026-08-14
fix: preserve zero retry-after-ms precedence

jstar0:

## Changes being requested

This fixes retry delay precedence when a response includes both `retry-after-ms: 0` and `retry-after`. The SDK already gives `retry-after-ms` priority for positive values, but `0` was treated as absent because the fallback check used truthiness.

The change keeps the parsed millisecond value whenever it is defined, so an explicit zero millisecond retry remains an immediate retry.

## Additional context & links

Added a regression test that verifies the `retry-after` HTTP-date parser is not consulted when `retry-after-ms` is present with value `0`.

Verification:

- `pnpm test:unit tests/client-behavior.test.ts`
- `pnpm test:unit tests/client-behavior.test.ts tests/index.test.ts tests/lib/azure.test.ts tests/lib/provider.test.ts tests/lib/workload-identity.test.ts`
- `pnpm lint`
- `pnpm build`

### DIFF
--- a/src/client.ts
+++ b/src/client.ts
@@ -1078,7 +1078,7 @@ export class OpenAI {
 
     // About the Retry-After header: https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Retry-After
     const retryAfterHeader = responseHeaders?.get('retry-after');
-    if (retryAfterHeader && !timeoutMillis) {
+    if (retryAfterHeader && timeoutMillis === undefined) {
       const timeoutSeconds = parseFloat(retryAfterHeader);
       if (!Number.isNaN(timeoutSeconds)) {
         timeoutMillis = timeoutSeconds * 1000;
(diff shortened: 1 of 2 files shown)


HAYDEN-OAI (commented):
Reviewed the complete two-file diff. The definedness check preserves an explicit zero-millisecond retry delay, keeps the existing fallback behavior for missing/invalid `retry-after-ms`, and the regression verifies the HTTP-date fallback is not consulted. I found no actionable issues.

HAYDEN-OAI:
Nice catch — preserving `retry-after-ms: 0` with an explicit `undefined` check makes the header precedence unambiguous, and the regression test captures the edge case clearly.

I also verified the latest commit locally: all 2,083 unit tests passed, along with type checking, linting, and the build. The GitHub checks are green across Node.js 22, 24, and 26 as well.

LGTM — thanks for the clean, focused fix!

HAYDEN-OAI (approved):

PR #2331, opened 2026-08-14
fix: parse empty chat content when present

jstar0:
Summary:
- Treat `content: ""` as present chat content during final structured-output parsing.
- Add regression coverage that a branded `response_format` parser receives the empty string.

Reason:
The Chat Completions parser used a truthy content check before invoking the `response_format` parser, so final assistant messages with empty-string content returned `parsed: null`. Present content should be handed to the parser and let that parser decide whether empty text is valid.

Verification:
- `pnpm exec vitest run --config vitest.config.mts tests/lib/parser.test.ts -t "parses present empty assistant content"`
- `pnpm exec vitest run --config vitest.config.mts tests/lib/parser.test.ts`
- `pnpm lint`
- `pnpm build`

### DIFF
--- a/src/lib/ChatCompletionStream.ts
+++ b/src/lib/ChatCompletionStream.ts
@@ -482,7 +482,14 @@ export class ChatCompletionStream<ParsedT = null>
   #emitContentDoneEvents(choiceSnapshot: ChatCompletionSnapshot.Choice) {
     const state = this.#getChoiceEventState(choiceSnapshot);
 
-    if (choiceSnapshot.message.content && !state.content_done) {
+    if (
+      choiceSnapshot.message.content != null &&
+      (choiceSnapshot.message.content !== '' ||
+        (!choiceSnapshot.message.refusal &&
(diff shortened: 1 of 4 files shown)


HAYDEN-OAI (commented):
Reviewed the complete two-file diff. Treating empty-string content as present lets the selected response-format parser decide validity while preserving null/undefined and refusal behavior; the focused branded-parser regression covers the changed branch. I found no actionable issues.

HAYDEN-OAI (approved):
Verified empty-content handling across streaming and non-streaming completions, including tool-call, missing-content, and refusal regressions.

HAYDEN-OAI (commented):
Follow-up review of the final four-file diff: the streaming path now preserves empty content, emits the matching completion event, and keeps empty tool-call-only/refusal choices unparsed; the added focused regressions cover streaming, absent content, tool calls, and refusal behavior. Both exact-head CI workflows passed. I found no new issues.

PR #2332, opened 2026-08-14
feat(api): Ultrafast tier, structured MCP and websocket errors, separate websocket events

apcha-oai:
## Summary

- Add GPT-5.5 dated and Pro model identifiers, the `ultrafast` Responses service tier, optional model shutdown dates, and cache-only web-search controls.
- Export a reusable `McpToolCallError` union for protocol, tool-execution, and HTTP failures, and remove the inaccurate string error variant from MCP tool calls.
- Correct Responses WebSocket error events to match their nested error payload and transport metadata; update the streaming example to read `event.error.message`.
- Add typed output-text annotations and text/image completion-token usage, and clarify completed reasoning-item reuse and file-content responses.
- Callers that previously treated MCP errors as strings or relied on the former WebSocket error shape must update their type narrowing.


jbeckwith-oai (changes requested):
The WebSocket server-event union now models only the nested protocol-error envelope and drops the real statusless streaming-error envelope. ResponseErrorEvent still exists in the streaming union, but every WebSocket frame is cast to ResponsesServerEvent and every type:error frame is routed through WebSocketError. A statusless streaming error is therefore falsely typed, and the updated event.error.message example is undefined for that valid shape. Please keep ResponseWsError and add a separately named WebSocket streaming-error variant based on ResponseErrorEvent (plus optional stream_id) to GA/beta unions; widen WebSocketError/_onError to both envelopes and select event.error?.message ?? event.message, with runtime/type coverage for both same-discriminator frames.

HAYDEN-OAI (commented):
Reviewed exact head a285573f8c221c076452eaa1f690de1b136bce31. The generated API/statistics/spec/example changes were reviewed. The existing blocking finding that the WebSocket server-event union drops the statusless ResponseErrorEvent envelope remains valid; I found no additional issues. Exact-head CI is green.

HAYDEN-OAI (commented):
Re-reviewed exact head 95bd99b578f3cbbb73abcc0423b584c4f80e9f5d. The regenerated GA/beta WebSocket unions still expose only the nested ResponseWsError shape and exclude the valid statusless ResponseErrorEvent streaming envelope, so the existing blocker remains. The other generated changes and exact-head CI are clean; no additional findings.

HAYDEN-OAI (approved):

PR #2333, opened 2026-08-14
fix: harden runTools against inherited function names

HAYDEN-OAI:
## Summary

- Use a null-prototype registry for `runTools()` callbacks so model-supplied names cannot resolve inherited properties and a registered `__proto__` cannot replace the registry prototype.
- Preserve existing unknown-tool recovery, retry behavior, advertised-tool boundaries, and explicitly registered `__proto__` callbacks in both streaming and nonstreaming runners.
- Keep the production change to the guaranteed-handwritten `src/lib/AbstractChatCompletionRunner.ts`.

## Reproduction

1. Register an ordinary runnable tool, then return a tool call named `constructor`, `toString`, `__proto__`, or `hasOwnProperty`. Both runner modes previously crashed with `TypeError: fn.function is not a function` instead of producing the recoverable `Invalid tool_call` message.
2. Register a tool explicitly named `__proto__`, then return an unadvertised call named `function`. The old plain-object registry adopted the tool descriptor as its prototype, allowing that inherited name through the advertised-tool boundary and causing the same crash.

## Validation

- Regression-first: 10 inherited-name/prototype-boundary cases failed on current `main` before the fix and pass after it.
- `./scripts/test tests/lib/ChatCompletionRunFunctions.test.ts` — 38 passed.
- `pnpm test:unit` — 70 files, 2,102 passed.
- `./scripts/lint` — passed.
- `pnpm exec tsc` — passed.

### DIFF
--- a/src/lib/AbstractChatCompletionRunner.ts
+++ b/src/lib/AbstractChatCompletionRunner.ts
@@ -406,7 +406,7 @@ export class AbstractChatCompletionRunner<
       return tool as any as RunnableToolFunction<any>;
     });
 
-    const functionsByName: Record<string, RunnableFunction<any, ToolContext>> = {};
+    const functionsByName: Record<string, RunnableFunction<any, ToolContext>> = Object.create(null);
     for (const f of inputTools) {
       if (f.type === 'function') {
         functionsByName[f.function.name || f.function.function.name] = f.function;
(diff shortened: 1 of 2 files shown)


jbeckwith-oai (approved):
Reviewed the exact current head. The null-prototype runTools registry correctly prevents inherited Object.prototype names and prototype mutation from crossing the advertised-tool boundary, while explicitly registered __proto__ remains callable. Streaming/non-streaming recovery and retry coverage is direct, CI is green, and no structural issue was found. No findings.

HAYDEN-OAI (commented):
Reviewed exact head 04a145b26592a57401dada1d38475ec479c1b356. The null-prototype runTools registry correctly prevents inherited prototype names and mutation from crossing the advertised-tool boundary, with focused regression coverage. Exact-head CI is green; no findings.

PR #2334, opened 2026-08-14
fix(streaming): isolate AssistantStream message and tool-call lifecycles

HAYDEN-OAI:

## Summary

Fix two user-facing lifecycle leaks in the **handwritten** `AssistantStream` implementation when one stream contains consecutive messages or run steps.

### Run-step tool calls

A terminal run-step event cleared the current tool call but retained its index. When the next step reused the normal tool-call index `0`, `AssistantStream` incorrectly treated its first call as a continuation: consumers received `toolCallDelta` but never received that call's `toolCallCreated` or `toolCallDone` events.

### Cross-message content attribution

Message completion left both the active content index and content snapshot attached to the stream. After a message ended on content index `1`, the next message's index-`0` delta re-emitted the old message's text or image-file payload with the **new message snapshot**. This is a cross-message data-attribution/privacy leak for consumers that route or persist completion events by message ID.

## Changes

- Reset both current-content fields when a message is created and after its terminal content/message completion events have been dispatched.
- Reset both current-tool-call fields when a run step is created and after every `completed`, `failed`, `cancelled`, or `expired` step has dispatched its terminal events.
- Clear both tool-call fields when a terminal run finishes an outstanding call.
- Preserve existing lifecycle ordering, including `toolCallDone` before `runStepDone` and `textDone`/`imageFileDone` before `messageDone`.

## Regression coverage

Public `AssistantStream.fromReadableStream` transport tests assert complete ordered lifecycle traces for:

- All four run-step terminal statuses followed by a second step that reuses tool-call index `0`, including both `toolCallCreated` and `toolCallDone`.
- Both terminal message statuses (`completed` and `incomplete`) crossed with prior `text` and `image_file` content, proving `PRIVATE_OLD_TEXT` and `PRIVATE_OLD_FILE` cannot be attributed to the next message.

**Regression-first evidence:** before the fix, the focused suite reported **8 failed / 66 passed**; after the fix it reports **74 passed**.

## Verification

```text
pnpm --config.verify-deps-before-run=ignore exec vitest run --config vitest.config.mts tests/lib/AssistantStream.test.ts
  74 passed

pnpm --config.verify-deps-before-run=ignore test:unit
  70 files passed; 2,098 tests passed

pnpm --config.verify-deps-before-run=ignore lint
  passed; formatting checked across 608 files

pnpm --config.verify-deps-before-run=ignore exec tsc
  passed

pnpm --config.verify-deps-before-run=ignore build
  passed; CommonJS/ESM builds and package/provider import smoke tests succeeded
```

The per-command pnpm option only lets the isolated worktree reuse the existing devbox dependency installation; no dependencies, lockfiles, generated files, or SDK exports changed.

## Scope

Only generator-exempt handwritten files changed: `src/lib/AssistantStream.ts` and `tests/lib/AssistantStream.test.ts`.

### DIFF
--- a/src/lib/AssistantStream.ts
+++ b/src/lib/AssistantStream.ts
@@ -421,6 +421,8 @@ export class AssistantStream
 
     switch (event.event) {
       case 'thread.message.created': {
+        this.#currentContentIndex = undefined;
+        this.#currentContent = undefined;
         this.#emitExposed('messageCreated', event.data);
         break;
       }
@@ -497,6 +499,8 @@ export class AssistantStream
           this.#emitExposed('messageDone', event.data);
         }
 
+        this.#currentContentIndex = undefined;
+        this.#currentContent = undefined;
         this.#messageSnapshot = undefined;
       }
     }
@@ -508,6 +512,8 @@ export class AssistantStream
 
     switch (event.event) {
       case 'thread.run.step.created': {
+        this.#currentToolCallIndex = undefined;
+        this.#currentToolCall = undefined;
         this.#emitExposed('runStepCreated', event.data);
         break;
       }
@@ -551,9 +557,10 @@ export class AssistantStream
         const details = event.data.step_details;
         if (details.type === 'tool_calls' && this.#currentToolCall) {
           this.#emitExposed('toolCallDone', this.#currentToolCall as ToolCall);
-          this.#currentToolCall = undefined;
         }
         this.#emitExposed('runStepDone', event.data, accumulatedRunStep);
+        this.#currentToolCallIndex = undefined;
+        this.#currentToolCall = undefined;
         break;
       }
       case 'thread.run.step.in_progress': {
@@ -727,8 +734,9 @@ export class AssistantStream
         this.#finalRun = event.data;
         if (this.#currentToolCall) {
           this.#emitExposed('toolCallDone', this.#currentToolCall);
-          this.#currentToolCall = undefined;
         }
+        this.#currentToolCallIndex = undefined;
+        this.#currentToolCall = undefined;
         break;
       }
       case 'thread.run.cancelling': {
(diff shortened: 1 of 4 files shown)


jbeckwith-oai (changes requested):
**[Medium] Move the lifecycle regressions into a focused test module.**

[`tests/lib/AssistantStream.test.ts:408-482`](https://github.com/openai/openai-node/blob/b06b60e71da2899a1c2cc21e1efc0ee2bd785de1/tests/lib/AssistantStream.test.ts#L408-L482) and [`:588-664`](https://github.com/openai/openai-node/blob/b06b60e71da2899a1c2cc21e1efc0ee2bd785de1/tests/lib/AssistantStream.test.ts#L588-L664) add 154 lines and push this broad suite from 858 to 1,012 lines, crossing the review threshold. The message and run-step lifecycle matrices are cohesive and independently testable. Please move them to a focused `tests/lib/assistant-stream-lifecycle.test.ts` and extract the small readable-event/AssistantStream transport factory into a shared test helper so both suites keep using the real `fromReadableStream` path.

The production isolation/reset behavior and exact-head CI otherwise look correct.

jbeckwith-oai (approved):
Re-reviewed exact head b0e08b5dcf94080213b4c6e864a2e0bfefd422e9. The prior thermo blocker is resolved: lifecycle matrices now live in a focused 159-line test, the broad AssistantStream suite is back below 1,000 lines, and the shared real-stream helper is appropriately scoped. Lifecycle reset ordering, public behavior, test coverage, and all exact-head CI are sound. No findings.

HAYDEN-OAI (commented):
Reviewed exact head b0e08b5dcf94080213b4c6e864a2e0bfefd422e9. AssistantStream lifecycle state is reset and emitted in the intended order, and the focused lifecycle regressions cover the relevant completion/error paths without enlarging the broad suite. Exact-head CI is green; no findings.

PR #2343, opened 2026-08-14
fix(qs): preserve astral Unicode across query encoding boundaries

HAYDEN-OAI:
## Summary

- Preserve complete UTF-16 surrogate pairs when the handwritten query encoder crosses its 1,024-code-unit batch boundary.
- Keep later batch boundaries anchored to their original absolute positions so existing malformed/lone-surrogate behavior and following characters remain unchanged.
- Add regression coverage for query values, computed keys, consecutive boundaries, RFC formatting, ISO-8859-1, and the public `client.files.list()` pagination path.

## Customer-visible data corruption

Any endpoint query value or key can be corrupted when an astral Unicode character straddles a 1,024-code-unit encoder boundary. Pagination cursors, filters, and arbitrary query keys are all affected. The broken path does not merely replace the emoji: it also combines the orphaned low surrogate with the immediately following character, silently changing or consuming that character.

Reproduction using the public SDK:

```ts
import OpenAI from 'openai';

let requestURL = '';
const client = new OpenAI({
  apiKey: 'test-key',
  fetch: async (url) => {
    requestURL = String(url);
    return Response.json({ object: 'list', data: [], has_more: false });
  },
});

const expected = 'a'.repeat(1023) + '😀' + 'b';
await client.files.list({ after: expected });

const actual = new URL(requestURL).searchParams.get('after')!;
const codePoints = (value: string) =>
  Array.from(value.slice(1023), (character) =>
    'U+' + character.codePointAt(0)!.toString(16).toUpperCase(),
  );

console.log(actual === expected);
console.log({ expected: codePoints(expected), actual: codePoints(actual) });
```

Before this change:

```text
false
{ expected: ['U+1F600', 'U+62'], actual: ['U+1F400', 'U+90062'] }
```

After this change:

```text
true
{ expected: ['U+1F600', 'U+62'], actual: ['U+1F600', 'U+62'] }
```

## Root cause and fix

The handwritten UTF-8 encoder processes `string.slice(j, j + 1024)` and reconstructs surrogate pairs only within each slice. A high surrogate in the final position is paired with an out-of-range code unit, while the next slice pairs its orphaned low surrogate with the following character.

Keep the established absolute 1,024-code-unit boundaries, inspect the code point at each candidate boundary, and extend the segment by one code unit only when that boundary splits a valid high/low surrogate pair. Advance by the actual consumed length without shifting later original boundaries. Batches remain bounded to at most 1,025 UTF-16 code units; the ISO-8859-1 early-return path and existing lone-surrogate/error policy are preserved.

## Regression matrix

| Astral high-surrogate offset | Query value | Query key | Expected behavior |
| --- | --- | --- | --- |
| 1022 | covered | covered | Pair completes immediately before the first boundary. |
| 1023 | covered | covered | Pair straddles the first boundary and round-trips intact. |
| 1024 | covered | covered | Pair starts immediately after the first boundary. |
| 2046 | covered | covered | Pair completes immediately before the second boundary. |
| 2047 | covered | covered | Pair straddles the second boundary and round-trips intact. |
| 2048 | covered | covered | Pair starts immediately after the second boundary. |
| 3071 | covered | covered | Pair straddles the third boundary and round-trips intact. |

Additional coverage includes `😀`, `𐐷`, the maximum Unicode code point `U+10FFFF`, multiple split symbols at fixed and shifted positions, following ASCII/BMP characters, `URLSearchParams` round-trips, RFC1738 and RFC3986 formatting, ISO-8859-1 numeric entities, and lone high/low surrogate extremes (`D800`, `DBFF`, `DC00`, `DFFF`) after an earlier split astral symbol.

## Verification

- Regression-first baseline: **13 failures / 122 existing passes**, including the real `files.list()` cursor.
- Additional malformed-surrogate compatibility baseline: **4 failures** before anchoring later batch boundaries.
- `./scripts/test tests/qs/stringify.test.ts` — **140 tests passed**.
- `pnpm test:unit --update=none` — **70 files / 2,120 tests passed**, with snapshot updates disabled.
- `pnpm lint` — formatting and lint passed across **608 files**.
- `pnpm exec tsc --noEmit` — passed.
- `pnpm build` — CommonJS and ESM builds plus package import smoke checks passed.

Only handwritten `src/internal/qs/utils.ts` and `tests/qs/stringify.test.ts` are changed. Generated/legacy `src/internal/utils/query.ts` and generated tests are untouched.


jbeckwith-oai (changes requested):
Exact-head review at 1bf9fa12c8df91fb37d468eda53c1a816982e390.

The Unicode-boundary fix is correct and all exact-head CI is green. One strict maintainability issue remains: the cohesive 117-line astral/segment-boundary regression suite further grows the already 2,227-line catch-all stringify test to 2,344 lines. The new cases, including the public files.list integration check, are independently understandable and have a clean extraction seam.

Please move them into a focused tests/qs/stringify-unicode-boundaries.test.ts module, retaining the current table-driven coverage and production change. No other findings.

jbeckwith-oai (approved):
Re-reviewed exact head b09932ccf6cc278fc055de7243afd86aecd56b6e. The Unicode boundary fix remains correct across astral pairs, lone-surrogate compatibility, charset/RFC modes, keys and values, and the public query path. The prior structural blocker is fully resolved: the focused regressions now live in a 120-line Unicode-boundaries suite and the broad stringify test is back to its base size. The previously unrelated Node 22 retry is now green, so all exact-head CI passes. Strict thermo is clean; no findings.

HAYDEN-OAI (commented):
Reviewed exact head b09932ccf6cc278fc055de7243afd86aecd56b6e. Segment boundaries now advance past a complete surrogate pair while retaining absolute chunk alignment and existing lone-surrogate, charset, and RFC formatting behavior. Focused public coverage and exact-head CI are green; no findings.

PR #2345, opened 2026-08-14
fix(auth): reject unsafe workload-identity token expirations

HAYDEN-OAI:
## User impact

Successful workload-identity OAuth exchanges previously trusted `expires_in` in unchecked timestamp arithmetic. Strings, objects, arrays, and non-finite values could produce `NaN`; a finite JSON value such as `1e308`, or the JSON numeric literal `1e309`, could produce `Infinity`. Both invalid timestamps make expiration comparisons permanently false, so public calls such as `new OpenAI({ apiKey: null, workloadIdentity }).models.list()` could continue reusing an expired bearer token indefinitely, even after the clock advances by 30 days. Zero and negative lifetimes were also returned and sent despite already being expired.

## Changes

- Require `expires_in` to be a primitive, finite, positive number before accepting a successful OAuth response.
- Require the resulting expiration timestamp to be a safe integer strictly after the current time, preventing multiplication overflow, unsafe epoch totals, and positive durations that round down to an already-expired timestamp.
- Reject malformed exchanges with the existing `OpenAIError` before returning or caching their access tokens.
- Preserve the existing 3,600-second fallback for both omitted and `null` values, valid fractional-second lifetimes, and large durations up to the safe JavaScript timestamp boundary without adding an arbitrary expiration cap.
- Preserve shared in-flight refreshes, retry after rejected refreshes, existing usable credentials during a failed background refresh, and the normal public workload-identity authentication flow.
- Add one dedicated handwritten regression suite; no generated or legacy-generated files are changed.

## Regression-first coverage

Before the implementation change, the new standalone suite produced **21 expected failures and 8 passing compatibility cases** on current `main`. After the fix, all **29 cases pass**, covering numeric and nonnumeric strings, objects, arrays, booleans, zero, negative values, injected `NaN`/infinities, `1e308` multiplication overflow, JSON-parsed `1e309`, unsafe total epoch expiration, representability boundaries, omitted/null defaults, concurrent refresh sharing and reset, background-refresh recovery, stale-token rejection, and real public `OpenAI.models.list()` requests.

## Verification

- `./scripts/test tests/auth/workload-identity-expiration.test.ts tests/auth/workload-identity-auth.test.ts tests/lib/workload-identity.test.ts --reporter=dot` — **64 passed across 3 focused files**.
- `env OPENAI_TEST_SUITE=unit PNPM_CONFIG_VERIFY_DEPS_BEFORE_RUN=false ./scripts/test --reporter=dot` — **2,119 passed across all 71 handwritten files**. The scoped pnpm setting accommodates the isolated private clone's copied workspace dependencies.
- `./scripts/lint` — passed.
- `./scripts/build` — passed.

### DIFF
--- a/src/auth/workload-identity-auth.ts
+++ b/src/auth/workload-identity-auth.ts
@@ -49,7 +49,7 @@ export class WorkloadIdentityAuth {
    *
    * @throws {OAuthError} When the token endpoint rejects the subject token or identity.
    * @throws {APIError} When another unsuccessful HTTP response prevents token exchange.
-   * @throws {OpenAIError} When a successful exchange does not include an access token.
+   * @throws {OpenAIError} When a successful exchange has an invalid access token or expiration.
    */
   async getToken(): Promise<string> {
     if (!this.cachedToken || WorkloadIdentityAuth.isTokenExpired(this.cachedToken)) {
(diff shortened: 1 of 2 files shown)


jbeckwith-oai (approved):
Reviewed exact head `479fd90d`. The workload-identity expiration validation rejects malformed, nonfinite, nonpositive, overflowing, unsafe, and non-advancing lifetimes before cache mutation or token return, while preserving missing/null fallback and millisecond-representable fractional durations. Foreground/background refresh recovery and public-client behavior remain coherent, the focused real-flow coverage is strong, strict thermo is clean, and all applicable exact-head CI is green.

PR #2347, opened 2026-08-14
fix(ws): prevent duplicate listener leaks in Node and browser adapters

HAYDEN-OAI:
## Summary

- Track every platform listener wrapper for each original callback instead of overwriting earlier registrations.
- Make `off()` remove the most recently registered matching listener, while each `once()` registration removes its own exact wrapper before invoking user code.
- Guard one-shot wrappers against reentrant Node.js `EventEmitter` dispatch, preserve native browser `EventTarget` callback identity/deduplication, and release empty listener bookkeeping immediately.
- Add a focused handwritten regression suite that exercises real `EventEmitter` and `EventTarget` behavior across both adapters.

## Impact and root cause

Both adapters previously stored only one wrapped listener per `(event, original callback)`. Re-registering the same callback overwrote that record while earlier platform listeners stayed attached. Matching `off()` calls could therefore leave subscriptions active indefinitely, and duplicate `once()` registrations could remove the wrong wrapper or fire again on later events.

The Node adapter is reachable through the customer-facing Responses WebSocket socket. Leaked subscriptions can process messages after unsubscribe, retain callbacks and their captured data, or repeat side effects that were intended to run only once. Browser behavior was additionally masked by the existing single-listener fake.

Node's `EventEmitter` snapshots listeners during dispatch, so exact-wrapper removal alone is insufficient for reentrant duplicate `once()` callbacks; each wrapper also needs its own fired guard. Browser pass-through events deliberately retain original callback identity, so duplicate native registrations must also be deduplicated in bookkeeping.

## Regression-first evidence

- Against freshly fetched `main`, all **30 new Node/browser listener-lifecycle regressions failed** while the other **2,090 handwritten tests passed**.
- A second targeted regression first reproduced retention of a duplicate browser `open` callback after one native `off()`, then passed after browser bookkeeping was aligned with `EventTarget` deduplication.
- Coverage includes duplicate `on/off`, newest-first removal ordering, duplicate `once`, both `on+once` orderings, repeated/reentrant dispatch, callback re-registration, throwing callbacks, event isolation, normalized close/error data, callback identity, and complete listener/map cleanup.

## Verification

- Focused existing + new adapter suites: `./node_modules/.bin/vitest run --config vitest.config.mts tests/internal/websocket-adapter-listeners.test.ts tests/internal/websocket-adapters.test.ts --no-cache --update=none` — **61 passed**.
- Complete handwritten suite: `env PNPM_CONFIG_VERIFY_DEPS_BEFORE_RUN=false ./node_modules/.bin/vitest run --config vitest.config.mts --no-cache --update=none` — **71 files / 2,120 tests passed**.
- Full generated-file checks, formatting, and lint: `./scripts/lint` — **passed; 609 files checked**.
- Full strict typecheck: `./node_modules/typescript/bin/tsc --noEmit --pretty false` — **passed**.
- Production CommonJS/ESM build and package import smoke tests: `./scripts/build` — **passed**.
- Published-source compatibility: `./node_modules/typescript-4-9/bin/tsc --project dist/src/tsconfig.json --noEmit --pretty false` and the equivalent current-TypeScript command — **both passed**.
- Additional executable checks against both built adapters verified duplicate unsubscribe, reentrant duplicate `once()`, and native-browser deduplication cleanup.

Only the two handwritten platform adapters and one new handwritten focused test change. Each path was checked for the Castiron generated header and against `scripts/generated-files.cjs`; no generated or `legacyFiles` paths are included.

### DIFF
--- a/src/internal/ws-adapter-browser.ts
+++ b/src/internal/ws-adapter-browser.ts
@@ -45,7 +45,7 @@ interface CloseEvent {
  */
 export class BrowserWebSocket implements WebSocketLike {
   private _ws: WebSocket;
-  private _listenerMap = new Map<string, Map<Listener, DOMEventHandler>>();
+  private _listenerMap = new Map<string, Map<Listener, DOMEventHandler[]>>();
 
   /** Wraps an existing browser socket and configures binary frames as array buffers. */
   constructor(ws: WebSocket) {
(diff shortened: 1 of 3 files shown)


jbeckwith-oai (approved):
Reviewed exact head `cbfba9a`. The ordered wrapper registries preserve Node duplicate/LIFO removal and browser EventTarget deduplication, while each `once()` removes its exact wrapper before user code and safely handles reentrant dispatch. Event isolation, registry pruning, normalized payloads, mixed `on`/`once`, throwing/re-registering callbacks, and existing call sites remain coherent. The focused real-EventEmitter/EventTarget coverage is strong, the single documented lint exception is necessary for the native semantics under test, strict thermo is clean, and all applicable exact-head CI is green.

PR #2350, opened 2026-08-14
fix(realtime): prevent remote error payloads from crashing clients

HAYDEN-OAI:
## Summary

- Protect all five server-controlled Realtime error fields (`message`, `code`, `param`, `type`, and `event_id`) with exception-safe string conversion in the stable and beta handwritten Realtime emitters.
- Preserve existing valid-error formatting, the original event/error metadata, documented asynchronous rejection for unhandled errors, and exceptions thrown by application error listeners.
- Add one dedicated handwritten regression suite covering the four public stable/beta × native/Node WebSocket clients.

## Security impact

A remote WebSocket peer that can deliver a valid JSON Realtime error frame can send a noncoercible error value, for example:

```json
{"type":"error","error":{"message":{"toString":null,"valueOf":null}}}
```

The existing formatter interpolates attacker-controlled values before invoking registered SDK error handlers. JavaScript coercion then throws `TypeError: Cannot convert object to primitive value`; Node WebSocket clients can consequently terminate through an uncaught exception even when an SDK `.on('error')` listener is registered. The same sink exists for all five error fields in both stable and beta clients. The impact is remotely triggerable client denial of service, not credential disclosure or authentication bypass.

The fix catches only conversion failures and substitutes `[unserializable error value]`. It does not serialize untrusted objects, change public APIs, swallow application listener exceptions, or modify transport/generated files.

## Verification

- Failure-first baseline on unchanged `main`: **32 failing / 20 passing** security cases, reproducing every field across all four public clients.
- `pnpm test tests/realtime-error-payload-security.test.ts --reporter=dot` — **52/52 passed**.
- `pnpm run test:unit --reporter=dot` — **2,142/2,142 passed across 71 handwritten test files**.
- `pnpm run lint` — passed.
- `pnpm exec tsc` — passed.
- `pnpm run build` — passed, including CommonJS/ESM package smoke checks.
- Verified every changed file lacks the Castiron generated-file header and is absent from the generated/legacy-file manifest.
- Refreshed all open-PR changed-file lists immediately before publication; none overlaps these three handwritten files.

### DIFF
--- a/src/beta/realtime/internal-base.ts
+++ b/src/beta/realtime/internal-base.ts
@@ -8,6 +8,14 @@ import { OpenAIError } from '../../error';
 import type OpenAI from '../../index';
 import { AzureOpenAI } from '../../index';
 
+function safeErrorValue(value: unknown): string {
+  try {
+    return String(value);
+  } catch {
+    return '[unserializable error value]';
+  }
+}
+
 /** An API-reported or client-side error encountered by a beta Realtime connection. */
 export class OpenAIRealtimeError extends OpenAIError {
   /** Stable error name used to identify Realtime connection failures. */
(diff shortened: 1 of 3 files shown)


jbeckwith-oai (approved):
Reviewed the full merge-base diff at this exact head with an independent second pass. Exception-safe string conversion now covers every remote Realtime error field across stable and beta Node/native transports, so hostile coercion cannot crash the client while ordinary error messages and emitter behavior remain intact. The focused 52-case suite exercises the real adapter paths; API compatibility, test quality, and strict thermo are clean with no avoidable suppressions. The previously unrelated Node 22 Skills/Versions failure has passed on retry, and all exact-head Node 22/24/26, build, lint, ecosystem, examples, performance, detect-breaking, and CodeQL checks are green.

PR #2356, opened 2026-08-14
fix(streaming): eliminate quadratic EventStream FIFO draining

HAYDEN-OAI:
## Summary
- Replace both shared `EventStream._createIterator` buffered-event and pending-reader `Array.shift()` queues with a private head-indexed FIFO.
- Release consumed references immediately, reset fully drained queues, and compact only after at least 1,024 consumed entries and when the consumed prefix is at least half of the backing array.
- Preserve FIFO ordering, buffered error/abort delivery, terminal reader cleanup, listener lifecycle, independent iterators, existing public APIs, and existing backpressure behavior.

## Security and performance
A slow consumer or adversarial streaming response can buffer `N` events and make synchronous draining move `N(N-1)/2` array elements, blocking the Node.js event loop. Pending readers and terminal reader cleanup had the same quadratic behavior. The shared adapter is used by `stream.events()`, Chat Completions, Responses, streaming tool runners, and readable-stream conversion.

Measured against latest `main` on the same Devbox:
- **12,000 buffered events:** `71,994,000` shifted elements → `0` shifted elements and only `10,976` copied elements across four amortized compactions.
- **120,000 buffered events:** `8,006.779 ms` → `7.493 ms` synchronous drain time, approximately **1,069× faster**. Timings are host-dependent; regression assertions use deterministic array-movement counts rather than timing thresholds.
- The same 12,000-entry pending-reader case falls from `71,994,000` shifted elements to `0` shifts and `10,976` amortized copies.

## Verification
- Regression-first baseline: the new focused suite produced **4 failures / 15 passes** on unmodified `main`; buffered values, pending readers, terminal reader cleanup, and a real public `ChatCompletionStream` iterator each moved exactly `8,386,560` elements for 4,096 entries, exceeding the `32,768` linear-work bound.
- New focused regression suite: **19 passed**.
- Focused EventStream, Chat Completions, and Responses suites: **92 passed across 4 files**.
- Full handwritten suite: **2,109 passed across 71 files** (`pnpm test:unit`).
- Repository-wide lint and formatter checks: **609 files passed** (`pnpm lint`).
- Full TypeScript check: `pnpm exec tsc --noEmit --pretty false --incremental false`.
- Production CommonJS/ESM build and runtime import smoke checks: `pnpm build`.
- Published source also typechecks under **TypeScript 4.9 and TypeScript 6**; package validation passes, and the private FIFO does not appear in emitted public declarations.

Only handwritten `src/lib/EventStream.ts` and the new handwritten `tests/lib/event-stream-queue-performance.test.ts` are changed.

### DIFF
--- a/src/lib/EventStream.ts
+++ b/src/lib/EventStream.ts
@@ -1,5 +1,49 @@
 import { APIUserAbortError, OpenAIError } from '../error';
 
+type EventQueue<Value> = {
+  readonly length: number;
+  enqueue: (value: Value) => void;
+  dequeue: () => Value | undefined;
+  clear: () => void;
+};
+
+function createEventQueue<Value>(): EventQueue<Value> {
+  let entries: (Value | undefined)[] = [];
+  let head = 0;
+
+  return {
+    get length() {
+      return entries.length - head;
+    },
+    enqueue(value) {
+      entries.push(value);
+    },
+    dequeue() {
+      if (head === entries.length) {
+        return undefined;
+      }
+
+      const value = entries[head];
+      entries[head] = undefined;
+      head += 1;
+
+      if (head === entries.length) {
+        entries = [];
+        head = 0;
+      } else if (head >= 1024 && head * 2 >= entries.length) {
+        entries = entries.slice(head);
+        head = 0;
+      }
+
+      return value;
+    },
+    clear() {
+      entries = [];
+      head = 0;
+    },
+  };
+}
+
 /** An abortable event stream with typed listeners, asynchronous iteration, and lifecycle state. */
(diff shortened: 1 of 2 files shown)


jbeckwith-oai (approved):
Reviewed the full exact-head diff in two independent passes. The private head-indexed queue preserves FIFO, terminal-error, abort, cleanup, early-return, and independent-iterator semantics while clearing consumed references and making drains amortized linear. Coverage exercises value and reader queues plus real streaming consumers; exported API/declarations and strict thermo/test policy are clean. All applicable exact-head CI is green. No findings.

PR #2361, opened 2026-08-15
test: stabilize Node 22 multipart mock-server 404 checks

HAYDEN-OAI:
## Summary

- Stabilize the generated Skills and Skill Versions request-options tests on Node.js 22.
- Buffer multipart uploads only for Steady's intentional `http://127.0.0.1:4010/_stainless_unknown_path` endpoint before forwarding them to the real `fetch` implementation.
- Keep generated tests, production SDK behavior, normal mock-server requests, and external test servers unchanged.

## Root cause

The generated Skills creation tests upload multipart `FormData` to an intentionally nonexistent route and expect `OpenAI.NotFoundError`. Steady sends HTTP 404 immediately when route matching fails, without consuming the request body.

On Node.js 22, Undici can still be writing the multipart upload when Steady closes the connection. That race produces `write EPIPE`, which the SDK correctly surfaces as `APIConnectionError` instead of the expected 404. The suite already runs with `--runInBand`, so limiting Jest workers does not fix the underlying transport race.

Example failure: https://github.com/openai/openai-node/actions/runs/31858956264/job/94948704693

## Fix

Register a generated-Jest-only setup file that wraps `globalThis.fetch`. The wrapper buffers the request only when its body is `FormData`, its origin is exactly `http://127.0.0.1:4010`, and its pathname is exactly `/_stainless_unknown_path`.

It builds a `Request`, reads the complete multipart payload into an `ArrayBuffer`, preserves the generated multipart boundary and original headers/options, and forwards the buffered bytes to the original fetch implementation. Steady still receives the real HTTP request and returns its real HTTP 404.

Handwritten regression coverage verifies multipart file bytes, fields, filenames, MIME types, headers, boundaries, query strings, request options, existing `Request` inputs, and unchanged handling of other routes, hosts, ports, JSON bodies, and `data:` requests.

## Verification

- Node.js 22.23.2 transport stress test: **17 failures / 200 requests before; 1,000 successful requests / 1,000 after**.
- Both previously flaky generated test suites: **16 consecutive Node.js 22 runs passed**.
- Full `./scripts/test` on Node.js 22.23.2 against current `main`: **2,704 handwritten tests and 559 generated tests passed**; one existing generated test remains skipped.
- Focused harness regression suite: **35 tests passed**.
- `pnpm lint` passed.
- `pnpm exec tsc --noEmit` passed.

### DIFF
--- a/jest.config.ts
+++ b/jest.config.ts
@@ -4,6 +4,7 @@ import generatedTestPatterns from './scripts/generated-test-patterns.json';
 const config: JestConfigWithTsJest = {
   preset: 'ts-jest/presets/default-esm',
   testEnvironment: 'node',
+  setupFiles: ['<rootDir>/scripts/jest-setup.ts'],
   testMatch: generatedTestPatterns.map(
     (pattern) => `<rootDir>/${pattern}${pattern.endsWith('.test.ts') ? '' : '/**/*.test.ts'}`,
   ),
--- a/scripts/jest-setup.ts
+++ b/scripts/jest-setup.ts
@@ -1,0 +1,3 @@
+import { bufferSteadyMultipartUploads } from './mock-server-fetch';
+
+globalThis.fetch = bufferSteadyMultipartUploads(globalThis.fetch);
--- a/scripts/mock-server-fetch.ts
+++ b/scripts/mock-server-fetch.ts
@@ -1,0 +1,25 @@
+const steadyOrigin = 'http://127.0.0.1:4010';
+const unknownPath = '/_stainless_unknown_path';
+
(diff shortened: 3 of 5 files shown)


jbeckwith-oai (approved):
Reviewed exact head `97068148d87b07604df0973f0ac3e4ff0cb0d41a` with mandatory two-pass review and strict thermo. No findings. The Jest-only native FormData buffering shim is narrowly scoped to the exact Steady unknown-route test endpoint, preserves serialized bytes/boundary/options and pass-through behavior, and leaves production transport untouched. Focused coverage and all exact CI are green.

PR #2363, opened 2026-08-15
fix(security): prevent streaming upload filename/header injection

HAYDEN-OAI:
## Summary

Supersedes #2338 with a narrowly scoped replacement based on the latest `main`.

- Require nonempty primitive-string filenames in `toStreamingFile()` without coercing user-controlled objects.
- Revalidate a possibly mutated `StreamingFile.name` immediately before serializing its multipart part, before that part's boundary is emitted.
- Percent-escape ASCII C0 controls, DEL, double quotes, and backslashes at the existing quoted multipart-header sink.
- Preserve Unicode filenames, existing `stripFilenames` and `unknown_file` behavior, and lazy single-pass streaming.

## Scope

Changes only handwritten `src/internal/uploads.ts` and one focused handwritten regression test file. This deliberately avoids request-wide preflight, multipart snapshots, upload-classification changes, iterator cleanup/replay changes, and generated files.

## Verification

- `pnpm test -- tests/internal/uploads-filename-security.test.ts tests/internal/uploads-branches.test.ts tests/form.test.ts` — 3 files, 90 tests passed.
- `pnpm lint` — passed across 637 files.
- `pnpm exec tsc --noEmit --project node_modules/tsconfig.streaming-filename-security.json` — passed for all SDK source and handwritten tests using a local config extending the repository's existing TypeScript configuration. Optional example dependencies were excluded from the isolated offline install.

### DIFF
--- a/src/internal/uploads.ts
+++ b/src/internal/uploads.ts
@@ -60,7 +60,7 @@ export function toStreamingFile(
   name: string,
   options?: Pick<FilePropertyBag, 'type'>,
 ): StreamingFile {
-  if (!name) {
+  if (typeof name !== 'string' || !name) {
     throw new TypeError('toStreamingFile requires a non-empty file name');
   }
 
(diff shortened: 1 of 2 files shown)


jbeckwith-oai (approved):
Reviewed the narrow replacement end to end. Primitive nonempty filename validation occurs at construction and again before the affected multipart boundary, getter-backed names are snapshotted once, the canonical header sink covers C0/DEL/quote/backslash without changing Unicode/path/fallback semantics, and streaming remains lazy and single-pass. Focused tests and strict thermo are clean; exact CI is green.

PR #2375, opened 2026-08-17
fix(bedrock): safely sign inherited query parameter names

HAYDEN-OAI:
## Summary

- Build Bedrock SigV4 query parameters in a null-prototype dictionary so valid query names such as `constructor`, `toString`, and `valueOf` are signed correctly instead of crashing with a misleading AWS credential error.
- Reject decoded `__proto__` parameters before invoking the signer, rotating credential provider, or fetch. Smithy's canonical query serializer cannot safely represent that name and would otherwise produce `[object Object]`.
- Resolve the request target before the credential-signing error wrapper so unsupported input produces a deterministic, actionable `OpenAIError`.
- Preserve Mantle/Runtime signing services, regions, credentials, origin checks, request bodies, headers, ordinary parameters, and repeated query values at the provider signing boundary.

## Regression coverage

The focused handwritten suite adds 26 cases covering both Mantle and Runtime, seven inherited property names, real SigV4 authorization and canonical queries, repeated signing-boundary values, ordinary controls, and raw/percent-encoded `__proto__` rejection without credential resolution or network calls.

Regression-first verification on unchanged main produced **24 failing cases and 2 passing controls**; all 26 pass with the fix.

## Verification

- Node 26 focused Bedrock suites: **213 tests passed**.
- Node 22 focused Bedrock suites: **213 tests passed**.
- `pnpm test:unit`: **96 files, 2,900 tests passed**.
- `pnpm lint`: passed.
- `pnpm exec tsc --noEmit`: passed.
- `pnpm build`: passed.
- `node ./node_modules/typescript-4-9/bin/tsc --project dist/src/tsconfig.json --noEmit`: passed.
- `node ./node_modules/typescript/bin/tsc --project dist/src/tsconfig.json --noEmit`: passed.
- `node --experimental-strip-types scripts/test-packed-package.ts`: passed CommonJS, ESM, and 266/303 source checks across 1,212 source maps.
- `pnpm exec publint dist`: passed with the existing unrelated vendored-default-export warning.

### DIFF
--- a/src/providers/bedrock/aws.ts
+++ b/src/providers/bedrock/aws.ts
@@ -101,8 +101,14 @@ function validateStaticCredentials(options: BedrockProviderOptions): AwsCredenti
 }
 
 function requestTarget(parsedURL: URL): { path: string; query: Record<string, string | string[]> } {
-  const query: Record<string, string | string[]> = {};
+  const query: Record<string, string | string[]> = Object.create(null);
   for (const [name, value] of parsedURL.searchParams) {
+    if (name === '__proto__') {
+      throw new Errors.OpenAIError(
+        'The Bedrock SigV4 signer cannot safely sign a `__proto__` query parameter.',
+      );
+    }
+
     const existing = query[name];
     if (existing === undefined) {
       query[name] = value;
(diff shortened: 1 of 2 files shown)


jbeckwith-oai (approved):
Reviewed the complete exact head. The null-prototype accumulator safely handles inherited query names, decoded __proto__ is correctly rejected before credentials/signing/fetch because Smithy writes canonical entries into a normal object, duplicate values and both Bedrock service scopes remain correct, and the focused tests exercise the real Smithy/request boundary. Exact CI and strict thermo are clean.

PR #2379, opened 2026-08-17
fix: drain tee stream queues in linear time

HAYDEN-OAI:
## Summary
- replace per-branch `Array.shift()` queues in public `Stream.tee()` with compact head-index queues that release consumed promises and periodically reclaim buffered storage
- preserve shared source promises, event ordering, independent reader speeds, source errors/completion, and explicit shared-controller cancellation
- add regression-first public streaming coverage for lag in both directions, bounded array movement and memory retention, interleaving, concurrent reads, and exact source-pull counts

## Verification
- reproduced 8,386,560 shifted element positions when draining 4,096 buffered events before the fix
- 71 focused streaming and queue tests
- 2,957 handwritten unit tests across 98 files
- changed-file formatter and all 415 lint rules
- repository TypeScript checks, CommonJS/ESM build, published TypeScript 4.9/current checks, and packed-package verification

### DIFF
--- a/src/core/streaming.ts
+++ b/src/core/streaming.ts
@@ -9,6 +9,45 @@ import type { OpenAI } from '../client';
 
 type Bytes = string | ArrayBuffer | Uint8Array | null | undefined;
 
+type StreamTeeQueue<Item> = {
+  readonly length: number;
+  enqueue: (value: Promise<IteratorResult<Item>>) => void;
+  dequeue: () => Promise<IteratorResult<Item>> | undefined;
+};
+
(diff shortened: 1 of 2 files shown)


jbeckwith-oai (approved):
Reviewed the exact head in two independent passes. The head-index queue preserves tee ordering, shared promise identity, pull counts, completion/rejection, cancellation, and existing backpressure semantics while making dequeue amortized O(1) and releasing consumed references promptly. Focused tests and exact-head CI are clean.

PR #2387, opened 2026-08-17
fix: settle concurrent stream event waiters in linear time

HAYDEN-OAI:
## Summary
- avoid quadratic listener-array shifts when many public `EventStream.emitted()` waiters settle on an event or an error
- index internal companion registrations with weak references, preserve the existing observable public `once()`/`off()` hooks, and compact cancelled listeners once after nested dispatch completes
- keep duplicate listeners, registration order, synchronous removal, nested/reentrant events, error/abort/end handling, and prototype-like event names compatible

## Verification
- regression-first success and error cases each reproduced 8,386,560 shifted elements with 4,096 waiters
- 65,536 successful waiters improved from 4,845 ms and 2,147,450,880 shifted elements to 51.5 ms and zero shifts; error-first waiters finish in 84.9 ms with zero shifts
- 94 focused listener and streaming tests
- all 2,979 handwritten unit tests across 100 files
- changed-file formatter and all 415 lint rules
- repository TypeScript, CommonJS/ESM build, published TypeScript 4.9/current checks, and packed-package verification

### DIFF
--- a/src/lib/EventStream.ts
+++ b/src/lib/EventStream.ts
@@ -65,7 +65,7 @@ export class EventStream<EventTypes extends BaseEvents> {
   #abortListeners: { signal: AbortSignal; listener: () => void }[] = [];
   #emittedListenerRegistrations = new WeakMap<
     object,
-    { event: PropertyKey; registration: { removed?: boolean } }
+    { event: PropertyKey; registration: { removed?: boolean; detached?: boolean } }
   >();
   #pendingListenerCleanup = new Set<PropertyKey>();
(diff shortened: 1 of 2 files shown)


jbeckwith-oai (approved):
Two-pass review complete at exact head `c3b84d904885201d1e7eb47230449745e8f0c340`. The indexed waiter registrations and deferred tombstone cleanup preserve snapshot dispatch, nested/reentrant emissions, duplicate/once/off semantics, error/end/abort settlement, iterator behavior, and linear cleanup. No actionable findings; all executed checks passed.

PR #2391, opened 2026-08-18
fix(security): reject malformed Realtime WebSocket event frames

HAYDEN-OAI:
## Summary
- Validate parsed Realtime WebSocket frames before exposing them to generic or typed listeners across stable/beta and native/Node transports.
- Reject null, arrays, primitives, missing or inherited discriminators, and non-string event types through the existing normalized SDK error path without exposing raw payloads.
- Preserve valid unknown/prototype-named event types, ordinary API errors and metadata, malformed JSON handling, transport decoding errors, and listener exception semantics.

## Security impact
A valid JSON frame such as `{ "type": { "toString": null, "valueOf": null } }` previously reached generic listeners and then threw an uncaught `TypeError` while coercing its discriminator into an event-listener key. A hostile or malformed upstream frame could synchronously terminate a Node process despite a registered SDK error listener and disclose its contents to generic event subscribers. All four public Realtime transports now fail closed before dispatch.

## Validation
- Regression-first: 60 failures before the fix; 104 public four-transport regression/compatibility cases pass after it.
- 384 focused Realtime, Bedrock WebSocket, redirect, error-payload, and compatibility checks.
- 3,088 handwritten tests and 559 generated tests pass.
- Repository formatting, all 415 lint rules, strict source/test TypeScript, and Node 22 production builds pass.
- Packed CommonJS/ESM consumers, published-source navigation, and TypeScript 4.9/current compatibility pass (266 browser-safe sources, 1,212 source maps).

### DIFF
--- a/src/beta/realtime/websocket.ts
+++ b/src/beta/realtime/websocket.ts
@@ -120,7 +120,19 @@ export class OpenAIRealtimeWebSocket extends OpenAIRealtimeEmitter {
     this.socket.addEventListener('message', (websocketEvent: MessageEvent) => {
       const event = (() => {
         try {
-          return JSON.parse(websocketEvent.data.toString()) as RealtimeServerEvent;
+          const parsedEvent = JSON.parse(websocketEvent.data.toString()) as RealtimeServerEvent;
+
+          if (
+            typeof parsedEvent !== 'object' ||
+            parsedEvent === null ||
+            Array.isArray(parsedEvent) ||
+            !Object.getOwnPropertyDescriptor(parsedEvent, 'type') ||
+            typeof parsedEvent.type !== 'string'
+          ) {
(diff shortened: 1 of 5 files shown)


jbeckwith-oai (approved):
Reviewed exact head `11fcfe60eea9127ae1cd61402b0644696dae3420` in two independent passes.

All four public Realtime WebSocket transports now reject malformed upstream event frames before generic/typed dispatch, requiring a non-null non-array object with an own string-valued `type`. Normalized SDK errors preserve causes without disclosing raw payloads, while unknown/prototype-named valid events, server-error metadata, application listener exceptions, credential/origin protections, and existing transport behavior remain compatible.

The regression suite covers 104 four-transport cases; Node 22/24/26, builds, types, ecosystem tests, and security analysis all pass. No findings.

PR #2392, opened 2026-08-18
fix(security): prevent AssistantStream deltas leaking sensitive data

HAYDEN-OAI:
## Summary

- Stop logging complete malformed AssistantStream array-delta entries directly to the global console during both index prevalidation and index application.
- Protect API keys, tool-call arguments, patient data, and other sensitive payloads from bypassing `logLevel: 'off'` or a custom redacting logger.
- Preserve the exact existing plain validation error, wrapped streaming `OpenAIError` and cause, atomic snapshots, stream lifecycle, and valid delta accumulation.
- Update the one historical unit-test expectation that explicitly required the unsafe console output.

## Regression coverage

Seven regression cases failed against unchanged `main` because the complete synthetic credential and patient payload reached `console.error`:

- Missing, null, and undefined indices through the public `AssistantStream.accumulateDelta` helper.
- A changing index that reaches the separate application-phase diagnostic sink.
- Public `AssistantStream.fromReadableStream` restoration.
- Real `client.beta.threads.runs.stream(...).done()` requests with disabled logging and with a custom redacting logger.

All eight new regression/control cases now pass.

## Validation

- 123 focused AssistantStream lifecycle, security, existing, and new privacy tests.
- 2,992 handwritten tests and 559 generated tests.
- Repository TypeScript, formatting, and all 415 lint rules.
- Full package build and published-source checks with TypeScript 4.9 and current TypeScript.
- Packed CommonJS/ESM artifact, 266 of 303 browser-safe source checks, and 1,212 source maps on Node 22.

### DIFF
--- a/src/internal/assistant-stream-delta.ts
+++ b/src/internal/assistant-stream-delta.ts
@@ -87,7 +87,6 @@ function getAssistantStreamDeltaIndex(
   const { index } = deltaEntry;
 
   if (kind === 'array' && (index === null || index === undefined)) {
-    console.error(deltaEntry);
     throw new Error('Expected array delta entry to have an `index` property');
   }
 
@@ -259,7 +258,6 @@ export function defineAssistantStreamArrayEntry(accumulator: unknown[], index: n
 function getRequiredAssistantStreamArrayIndex(deltaEntry: AssistantStreamRecord): number {
   const { index } = deltaEntry;
   if (index === null || index === undefined) {
-    console.error(deltaEntry);
     throw new Error('Expected array delta entry to have an `index` property');
   }
   if (typeof index !== 'number') {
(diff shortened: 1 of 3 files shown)


jbeckwith-oai (approved):
Reviewed both removed sensitive-data console sinks and the full AssistantStream validation/lifecycle paths. Invalid array deltas keep the exact prior error, wrapped cause, atomic snapshot, and stream failure semantics while no longer bypassing disabled logging or custom redaction. Focused public-helper, readable-stream restoration, and real client streaming regressions cover missing indices, changing getters, credentials, patient data, and valid accumulation. Exact-head Node 22/24/26, lint, build, ecosystem, and security checks are green. No actionable findings.

PR #2403, opened 2026-08-18
fix(security): redact malformed SSE diagnostic payloads

HAYDEN-OAI:
## Summary

- Keep malformed server-sent event payloads and raw chunks out of default console logs and configured logger arguments.
- Preserve both existing diagnostic messages, logger routing, disabled logging, SyntaxError identity, request cancellation, and valid stream behavior.
- Cover real public Assistants, Responses, and Chat Completions streams plus the exported SSE helper and both ordinary/thread event branches.

## Validation

- Regression-first: six public privacy cases failed before the fix; all 12 dedicated regressions now pass.
- 76 focused streaming, SSE, and logging tests.
- 3,337 handwritten tests and 559 generated tests.
- Repository lint, strict TypeScript, production build, and packed CommonJS/ESM consumers using TypeScript 4.9 and current TypeScript.

### DIFF
--- a/src/core/streaming.ts
+++ b/src/core/streaming.ts
@@ -114,8 +114,8 @@ export class Stream<Item> implements AsyncIterable<Item> {
             try {
               data = JSON.parse(sse.data) as any;
             } catch (e) {
-              logger.error(`Could not parse message into JSON:`, sse.data);
-              logger.error(`From chunk:`, sse.raw);
+              logger.error(`Could not parse message into JSON:`);
+              logger.error(`From chunk:`);
               throw e;
             }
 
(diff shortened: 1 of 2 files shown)


jbeckwith-oai (changes requested):
[P2] Both malformed SSE branches sanitize logger arguments but immediately rethrow the original JSON.parse SyntaxError. V8 includes malformed input in that error message and stack, so sensitive payloads still reach caller exceptions and monitoring even though direct logs are redacted. For example, JSON.parse of hunter2 exposes hunter2 in both surfaces. This affects ordinary and thread events at src/core/streaming.ts lines 119 and 134. Throw a fresh generic SyntaxError without preserving the sensitive original as cause, retaining existing cancellation and error-type compatibility. Add regressions proving that propagated message, stack, and cause never contain sensitive malformed payloads across both branches.

jbeckwith-oai (approved):
Independent two-pass re-review confirms the prior malformed SSE privacy gap is fixed. Both ordinary and thread event paths now sanitize logs and replace payload-bearing parser exceptions with fresh generic SyntaxError values without sensitive messages, stacks, or causes, while preserving cancellation and existing error-type compatibility. Chat, Responses, and Assistants regression coverage is comprehensive. No remaining findings.

PR #2406, opened 2026-08-18
fix(security): redact sensitive assistant delta diagnostics

HAYDEN-OAI:
## Summary

- Prevent malformed AssistantStream message and run-step deltas from placing assistant text, tool arguments, attacker-controlled keys, array entries, or indices in error messages, stacks, and configured logger output.
- Preserve useful allowlisted field names, safe numeric-index diagnostics, existing exception classes, snapshot rollback, and normal streaming semantics without coercing hostile values.
- Cover the public streamed Assistants API, direct accumulation, error listeners, logger-on/off behavior, malformed arrays, hostile coercion, and both message/run-step snapshots with regression-first tests.

## Validation

- New privacy regressions: 59 passing (53 failed against the original implementation).
- AssistantStream compatibility/security matrix: 186 passing across 6 files.
- Complete handwritten suite: 3,491 passing across 112 files.
- Complete generated suite: 559 passing across 82 files.
- `pnpm lint` and strict `tsc --noEmit`.
- Full package build plus published-source TypeScript 4.9 and TypeScript 6 checks.
- `publint dist` and packed CommonJS/ESM/type-consumer compatibility (267 source checks, 1,216 source maps).

### DIFF
--- a/src/internal/assistant-stream-delta.ts
+++ b/src/internal/assistant-stream-delta.ts
@@ -6,6 +6,28 @@ const MAX_EXTERNALLY_MUTABLE_ASSISTANT_STREAM_ARRAY_LENGTH = 65_536;
 
 type AssistantStreamRecord = Record<string, unknown>;
 
+function getAssistantStreamDiagnosticProperty(property: string): string {
+  switch (property) {
+    case 'value':
+    case 'arguments':
+    case 'input':
+    case 'text':
+    case 'content':
+    case 'annotations':
+    case 'metadata':
+    case 'name':
+    case 'role':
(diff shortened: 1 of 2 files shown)


jbeckwith-oai (approved):
Independent two-pass security review confirms malformed assistant delta diagnostics redact sensitive values and hostile keys without invoking coercion hooks, while retaining trusted protocol context. Errors, stacks, causes, snapshots, public stream entry points, listener lifecycle, and normal accumulation remain safe and compatible. Focused adversarial coverage and CI are sound. No findings.

PR #2407, opened 2026-08-18
fix(security): stop retaining sensitive assistant stream events

HAYDEN-OAI:
## Summary

- Remove an unused private AssistantStream event-history array that retained every raw Assistants API event for the lifetime of the runner.
- Release consumed credential- and patient-bearing events after callbacks or async iterators drain, while preserving event delivery, current snapshots, final runs, messages, and run steps.
- Add deterministic public-stream, WeakRef, and inspector garbage-collection regressions for callbacks, drained async iterators, and growing 256-event histories.

## Validation

- Regression-first: all three new retention tests fail against unchanged main and pass after the four-line production deletion.
- Focused retention regressions pass on Node.js 22.23.0 and 26.7.0.
- Full canonical suite: 3,435 handwritten tests and 559 generated tests pass.
- Focused AssistantStream lifecycle/security suite: 118 tests pass.
- `pnpm lint`
- `pnpm build`
- `./node_modules/typescript-4-9/bin/tsc --project dist/src/tsconfig.json --noEmit`
- `./node_modules/typescript/bin/tsc --project dist/src/tsconfig.json --noEmit`
- `pnpm exec publint dist`
- `node --experimental-strip-types scripts/test-packed-package.ts` passes on Node.js 22 and 26.

### DIFF
--- a/src/lib/AssistantStream.ts
+++ b/src/lib/AssistantStream.ts
@@ -104,9 +104,6 @@ export class AssistantStream
   extends EventStream<AssistantStreamEvents>
   implements AsyncIterable<AssistantStreamEvent>
 {
-  //Track all events in a single list for reference
-  #events: AssistantStreamEvent[] = [];
-
   //Used to accumulate deltas
   //We are accumulating many types so the value here is not strict
   #runStepSnapshots: Record<string, Runs.RunStep> = Object.create(null);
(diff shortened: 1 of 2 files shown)


jbeckwith-oai (approved):
Independent two-pass security review confirms the removed raw-event history was an unused ECMAScript-private accumulator. Sensitive AssistantStream events can now be collected while current-event visibility, callbacks, async iteration, snapshots, final results, cancellation, and error propagation remain unchanged. Deterministic retention tests and supported-runtime CI are sound. No findings.

PR #2409, opened 2026-08-18
fix(streaming): reject named SSE provider error events

HAYDEN-OAI:
## Summary
- Reject named SSE `event: error` frames as typed `APIError`s instead of yielding flat provider failures as successful Chat Completions, Responses, and Assistants stream events.
- Preserve nested provider errors, request headers and request IDs, cancellation, ordinary named events, Assistant event envelopes, completion sentinels, and private malformed-event diagnostics.
- Remove the unreachable error check from the Assistant-only branch and add 19 public-API and exported-stream regression cases.

## Validation
- Regression-first on untouched main: all three actual `new OpenAI({ fetch })` public streaming APIs and both exported event-synthesis modes incorrectly fulfilled before the fix.
- `OPENAI_TEST_SUITE=unit pnpm test -- --maxWorkers=6` — **3,531 handwritten tests passed**.
- `OPENAI_TEST_SUITE=generated pnpm test -- --runInBand` — **559 generated tests passed**.
- `pnpm lint` — zero errors and warnings.
- `pnpm exec tsc --pretty false`; `pnpm build`.
- Published source type-checked with TypeScript **4.9** and **6**.
- `pnpm exec publint dist`; packed-package CommonJS, ESM, and 267 source checks passed on Node **22.23.2** and **26.7.0**.
- Dedicated 19-case streaming regression passed on Node 22 and Node 26.

### DIFF
--- a/src/core/streaming.ts
+++ b/src/core/streaming.ts
@@ -119,6 +119,10 @@ export class Stream<Item> implements AsyncIterable<Item> {
               throw new SyntaxError('Could not parse server-sent event data as JSON.');
             }
 
+            if (sse.event === 'error') {
+              throw new APIError(undefined, data?.error ?? data, undefined, response.headers);
+            }
+
             if (data && data.error) {
               throw new APIError(undefined, data.error, undefined, response.headers);
             }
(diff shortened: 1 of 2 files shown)


jbeckwith-oai (approved):
Independent two-pass review: named SSE provider errors are correctly converted into APIError while preserving nested error behavior, response metadata, request IDs, assistant envelopes, cancellation, and sanitized malformed-event diagnostics. No actionable findings.

PR #2411, opened 2026-08-18
fix(security): bound SSE frames and NDJSON lines before buffering

HAYDEN-OAI:
## Security findings

- `csf_68ab32fb052e825f536dfd7a`: unbounded active server-sent event framing buffer.
- `csf_ebb44bf3e8b19efdf555d9cc`: unbounded active NDJSON line buffer.

## Changes

- Cap SSE frames and NDJSON lines independently at **8 MiB by default**, configurable through `OPENAI_MAX_SSE_EVENT_BYTES` and `OPENAI_MAX_NDJSON_LINE_BYTES`.
- Reject oversized input before UTF-8 encoding, backing-buffer allocation, resize, compaction, or copy; abort and cancel upstream readers immediately.
- Preserve valid packed frames, large chunks containing many short lines, exact-boundary input, fragmented CRLF, Unicode/surrogate pairs, and existing public streaming behavior.
- Preserve merged named-SSE error handling (#2409) and malformed-event diagnostic privacy (#2403).
- Limit changes to handwritten streaming/decoder sources and focused regression tests; generated/upstream-owned logging code is intentionally excluded.

## Verification

- **173 focused tests passed** across SSE limits, NDJSON limits, named-error events, malformed-event privacy, existing streaming behavior, logging, and provider coverage.
- **3,469 handwritten tests passed across 115 files**, excluding two existing environment-dependent infrastructure suites.
- Strict source TypeScript check passed:
  `./node_modules/.bin/tsc --ignoreConfig --pretty false --noEmit --target es2020 --module commonjs --moduleResolution bundler --esModuleInterop --skipLibCheck --strict --types node,jest src/core/streaming.ts src/internal/decoders/line.ts`
- Repository-configured Oxlint, Oxfmt, generated-file checks, and `git diff --check` passed for the owned changes.

**Rollout:** Streams exceeding the new 8 MiB per-event/per-line default now fail closed; deployments that legitimately require larger records can raise the corresponding independent environment limit.

### DIFF
--- a/src/core/streaming.ts
+++ b/src/core/streaming.ts
@@ -4,6 +4,7 @@ import { makeReadableStream, ReadableStreamToAsyncIterable } from '../internal/s
 import { findDoubleNewlineIndex, LineDecoder } from '../internal/decoders/line';
 import { isAbortError } from '../internal/errors';
 import { encodeUTF8 } from '../internal/utils/bytes';
+import { readEnv } from '../internal/utils/env';
 import { loggerFor } from '../internal/utils/log';
 import type { OpenAI } from '../client';
 
@@ -325,6 +326,7 @@ export class Stream<Item> implements AsyncIterable<Item> {
 
 /**
  * Decodes complete SSE records from a response and aborts when its body is absent.
+ * Frames default to an 8 MiB limit, configurable with `OPENAI_MAX_SSE_EVENT_BYTES`.
  *
  * @yields {ServerSentEvent} Each decoded server-sent event in wire order.
  */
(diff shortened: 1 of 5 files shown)


jbeckwith-oai (changes requested):
[P2] The new unconditional limit reads in src/core/streaming.ts:377 and src/internal/decoders/line.ts:49 throw Deno PermissionDenied when environment access is unavailable. This newly breaks the exported Stream.fromSSEResponse and Stream.fromReadableStream utilities, as well as existing clients after environment permission is revoked; these decoding paths previously required no environment access. Existing Deno CI always grants --allow-env and misses the regression. Please catch inaccessible-environment errors and retain the secure default limit; the shared readEnv contract already says inaccessible variables should behave as undefined. Add denied-environment streaming regression coverage.

jbeckwith-oai (approved):
The previous denied-environment Deno regression is fixed with secure fallback and direct public streaming coverage. Independently verified bounded SSE/NDJSON buffering, UTF-8 and fragmented CRLF handling, cancellation, configuration, and sanitized errors. All prior findings are resolved and checks pass.

PR #2420, opened 2026-08-19
fix(security): redact malformed NDJSON stream diagnostics

HAYDEN-OAI:
## Summary

- sanitize only built-in malformed newline-delimited JSON parsing in `Stream.fromReadableStream`, preventing raw streamed patient data, credentials, and other response contents from escaping through error messages, stack traces, or wrapper error causes
- preserve a fresh payload-free `SyntaxError` for the direct stream and the existing `OpenAIError` wrapper/cause behavior across public Chat Completions, Assistants, and Responses stream replay constructors
- retain unrelated exception identity, cancellation/reader cleanup, valid fragmented or unterminated NDJSON, configured line-size failures, disabled logging, and existing server-sent-event behavior

## Security regression

The regression was written and run against unchanged `main` first: **6 failed and 8 passed**. All four real public `fromReadableStream` surfaces exposed synthetic patient and credential values on the failing baseline. With the minimal parser-boundary fix, all **14 privacy and compatibility regressions pass** without storing the original leaking exception as a nested cause.

## Validation

- focused streaming suites: **9 files, 241 tests passed**
- full handwritten suite: **122 files, 3,843 tests passed**
- generated API suite against an isolated local Steady server: **82 suites, 559 tests passed** (1 pre-existing skipped test)
- full repository formatter/lint and strict TypeScript checks passed
- production package build passed; emitted package types checked with current TypeScript and TypeScript 4.9
- package lint passed with its existing unrelated vendor-export warning
- packed npm artifact smoke-tested all four public stream constructors and verified payload-free outer errors and causes

Scope is exactly the handwritten stream implementation and one new dedicated regression test.

### DIFF
--- a/src/core/streaming.ts
+++ b/src/core/streaming.ts
@@ -239,7 +239,18 @@ export class Stream<Item> implements AsyncIterable<Item> {
             continue;
           }
           if (line) {
-            yield JSON.parse(line) as Item;
+            let data: Item;
+            try {
+              data = JSON.parse(line) as Item;
+            } catch (error) {
+              if (error instanceof SyntaxError) {
+                throw new SyntaxError('Error reading response: malformed newline-delimited JSON.');
+              }
+
+              throw error;
+            }
+
+            yield data;
           }
         }
         done = true;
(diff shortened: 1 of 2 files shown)


jbeckwith-oai (approved):
Independent two-pass security review confirms malformed NDJSON now raises a fresh payload-free SyntaxError across public streams and Chat, Assistants, and Responses wrappers. Sensitive data stays out of error messages, causes, and stacks while unrelated errors, valid fragmented input, UTF-8, limits, cancellation, and reader cleanup retain their established behavior. Focused regressions and substantive CI are green; no actionable findings.

PR #2423, opened 2026-08-19
feat(api): Add obfuscation field to ChatCompletionChunk

apcha-oai:
## Summary

Add an optional obfuscation string to streamed chat completion chunks to normalize chunk sizes as a mitigation for certain side-channel attacks. The new property is optional and included by default; it is omitted when stream_options.include_obfuscation is false.

## Changes

- Add new optional property ChatCompletionChunk.obfuscation?: string to the public TypeScript types for streamed chat completion chunks.
- Streamed responses will include an obfuscation string by default to normalize chunk size; the field will be omitted when stream_options.include_obfuscation is false.
- This is a backwards-compatible API extension (optional field); consumers should tolerate an additional optional string on ChatCompletionChunk objects.


apcha-oai:
The Castiron custom-code report is available in the [successful CI run](https://github.com/openai/openai-node/actions/runs/32303840118): open the job summary or download the `castiron-custom-code-32303840118-1` artifact for the full report and patch. I checked that the public report validates the generated baseline and agrees with the reviewed candidate's customization counts and classifications.

This PR also introduces the automatic comment publisher. GitHub only starts that `workflow_run` workflow after it reaches the default branch, so this first rollout PR has a report in Actions but no automatic report comment yet. Once merged, subsequent PR runs will publish/update the comment automatically, including on ordinary branches and drafts. No branch allowlist is required.

jbeckwith-oai (changes requested):
Independent two-pass review confirms the existing Medium finding in discussion #3816897922: chunk-only obfuscation is copied into accumulated/final chat completions. Preserve padding on raw chunks while excluding it from completion snapshots/final objects. No duplicate inline comment added.

jbeckwith-oai (approved):
Independent two-pass re-review confirms the prior chunk-only obfuscation leak is fixed throughout initial and subsequent accumulation: raw chunks remain unchanged, snapshots/final completions omit padding, and public streaming regressions cover serialization and absent/empty/populated values. Substantive CI/CodeQL pass.

PR #2431, opened 2026-08-20
test: make ecosystem CLI checks portable on Windows

huanglianggit:
## Problem

`tests/ecosystem-cli.test.ts` fails on Windows even though its assertions are platform-independent. A CRLF checkout prevents the workflow-condition helper from finding the `if` block, and the fake npm command is written only as a POSIX executable, while Windows command resolution expects `npm.cmd`.

Baseline on Windows: 5 failed, 12 passed.

## Root cause

The test helper parses workflow text with LF-only delimiters, and the subprocess fixture installs only an extensionless shell script on `PATH`.

## Fix

- Normalize workflow line endings before extracting jobs.
- Add an explicit CRLF regression.
- Write the successful npm fixture as `npm.cmd` on Windows while preserving the existing POSIX stub elsewhere.

## Tests

- `pnpm exec vitest run --config vitest.config.mts --update=none tests/ecosystem-cli.test.ts --reporter=verbose` (18 passed)
- `pnpm exec vitest run --config vitest.config.mts --update=none --exclude tests/ecosystem-cloudflare-credential-lifecycle-security.test.ts --reporter=dot` (125 files, 3,966 tests passed)
- `pnpm exec oxfmt --check tests/ecosystem-cli.test.ts`
- `pnpm exec oxlint tests/ecosystem-cli.test.ts`
- `pnpm build`

## Compatibility / Risk

Test-only change. SDK runtime behavior and exports are unchanged. POSIX fixtures keep their existing shell-script behavior; Windows uses the native `.cmd` lookup path. The broader Windows run excludes the pre-existing credential lifecycle suite because it exercises POSIX file-mode, link, and signal semantics outside this patch.


HAYDEN-OAI (approved):
Reviewed the Windows portability changes: CRLF workflow normalization and the native npm.cmd fixture preserve existing POSIX behavior without changing SDK runtime or public APIs. LGTM.

PR #2433, opened 2026-08-20
fix: preserve large streaming responses

apcha-oai:
## Summary

Remove the unreleased SDK-side size limits that reject valid streaming responses. A supported image-generation request can return a single event larger than 8 MiB; the current limit aborts that stream before the caller receives the image. Large final Responses snapshots and NDJSON records have the same problem.

Restore the released behavior without replacing the cap with another arbitrary ceiling. Retain the established event-by-event SSE framing, amortized-linear buffering, and stream cleanup so packed transport chunks do not eagerly decode events the caller may never consume. Add deterministic public-API contract tests for large image, Responses, Chat Completions, and NDJSON payloads, plus guidance against introducing unapproved payload limits in future security or performance fixes.


jbeckwith-oai (changes requested):
The existing exact-head P2 review thread remains actionable: `_iterSSEMessages` now eagerly decodes and materializes every completed line in a transport chunk before yielding the first event or honoring `[DONE]`. Large legitimate/custom-fetch chunks therefore lose backpressure and can cause avoidable memory exhaustion. Process bounded views incrementally and yield lazily while preserving the explicitly unlimited event/line-size contract; do not restore arbitrary rejection limits. No duplicate inline comment.

jbeckwith-oai (changes requested):
The follow-up correctly restores lazy SSE line decoding, but the existing unresolved backpressure issue is only partially fixed. `iterSSEChunks` still allocates and copies the entire incoming transport chunk before yielding its first event (`src/core/streaming.ts:417-427`). Consequently a chunk containing `[DONE]` followed by a large discarded tail still duplicates all trailing bytes and can exhaust memory before cancellation. Please process bounded `subarray` views incrementally and yield completed events immediately, while preserving unlimited event sizes. The existing review thread already covers this; no duplicate inline comment.

apcha-oai:
Responding explicitly to https://github.com/openai/openai-node/pull/2433#issuecomment-5361437522: the whole-chunk copy is a real, pre-existing limitation, and we are accepting it for this PR.

I compared `iterSSEChunks` at the current head (`14d7a468`) with both released 7.5.0 and the code immediately before #2411; the function is byte-for-byte identical. The new eager line-decoding regression has been fixed. This PR restores support for valid large events without introducing another buffering rewrite.

To avoid scope creep and additional compatibility risk in a touchy path, we are deliberately deferring bounded incremental processing to a separate follow-up. That work should preserve unlimited event sizes and cover packed transport chunks, early cancellation, `[DONE]`, split delimiters, and large valid events. Please treat the remaining whole-chunk-copy concern as an accepted limitation here, not a blocker for this compatibility fix.

jbeckwith-oai (approved):

PR #2434, opened 2026-08-20
fix: [responses] Avoid quadratic comparisons when accumulating long text

apcha-oai:
## Summary

#2354 made direct `accumulateResponse(event, snapshot)` calls recount caller-owned output so that edits between calls cannot leave stale text behind. That correctness guarantee is necessary, but the recount also compared the rebuilt string with the previous aggregate. For a long response, those independently accumulated strings can be equal only after walking the entire prefix again on every delta.

Assign the authoritative recount directly. A local replay of 16,000 64-byte deltas (1.024 MB) dropped from about 1.5 seconds to 14 ms on Node 22 and 24. This preserves the existing public API, snapshot-mutation behavior, and privately owned `ResponseStream` context; it does not introduce a cache or pretend the necessary content-part scan can be eliminated.

Add a deterministic public-entrypoint regression for that long-response shape. It detects reads of the old aggregate before the recount is installed, so the original implementation fails without relying on a wall-clock threshold.

### DIFF
--- a/src/internal/responses/canonical-output-text.ts
+++ b/src/internal/responses/canonical-output-text.ts
@@ -44,9 +44,9 @@ export function ensureCanonicalOutputText(context: ResponseAccumulatorContext, s
     text += outputText;
     outputTextIndex.append(outputText.length);
   }
-  if (snapshot.output_text !== text) {
-    snapshot.output_text = text;
-  }
+  // Comparing independently accumulated strings can scan all preceding text on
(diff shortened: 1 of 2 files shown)


jbeckwith-oai (approved):
Independently verified the response-stream accumulator removes repeated full-prefix comparisons while preserving canonical accumulated text, mixed content, ordering, caller-owned snapshots, and existing stream/error behavior. The deterministic 16,000-delta regression directly detects the previous quadratic path, and all CI checks are green.

PR #2435, opened 2026-08-20
test: move custom resource regressions into handwritten suites

apcha-oai:
Move handwritten SDK behavior checks out of generated API-resource tests so regeneration no longer has to preserve those patches. Both generated files now match the verified public generated snapshot exactly. No production source, public declarations, dependencies, API-reference artifacts, or generation metadata changes.

Replacement coverage:

- `tests/lib/embeddings.test.ts` retains `resource embeddings` → `create: encoding_format=default should create float32 embeddings`, `create: encoding_format=float should create float32 embeddings`, and `create: encoding_format=base64 should return base64 embeddings`. It uses the same existing response fixtures and preserves the array/string and exact numeric assertions. The existing four embedding-type regressions remain.
- `tests/lib/responses-output-text.test.ts` → `resource responses output_text` → `create` and `retrieve` preserve the `output_text` presence/type assertions and the raw-response, parsed-data, and `withResponse()` identity checks. These run against a deterministic local fetch response.

The public custom-code report verifies **34 → 32 mixed files**, **2,121 → 2,043 custom-patch lines** (additions plus deletions), and all other **32 customizations unchanged**.

Validation: pinned pnpm 11.5.1 install with frozen lockfile; Oxfmt 0.62.0; full lint; TypeScript 6 check; build; published-source TypeScript 4.9 check; 5,272 handwritten tests; 556 generated tests and 12 snapshots (one test skipped); packed-package CJS/ESM and published-type checks on the exact Node 22.0.0 floor. Generated tests used pinned Steady 0.22.1 on a separate local port.

### DIFF
--- a/tests/api-resources/embeddings.test.ts
+++ b/tests/api-resources/embeddings.test.ts
@@ -1,9 +1,6 @@
 // File generated from our OpenAPI spec by Castiron. See CONTRIBUTING.md for details.
 
 import OpenAI from 'openai';
-import { mockFetch } from '../utils/mock-fetch';
-import fs from 'fs/promises';
-import Path from 'path';
 
 const client = new OpenAI({
   apiKey: 'My API Key',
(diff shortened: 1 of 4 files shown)


jbeckwith-oai (approved):
Independently verified the embedding and Responses `output_text` regressions are moved into handwritten SDK-owned test suites without dropping assertions or altering production code, public APIs, or generation metadata.

PR #2436, opened 2026-08-20
refactor: move embedding encoding into an SDK-owned helper

apcha-oai:
Move the existing embedding encoding optimization into an SDK-owned helper, leaving a thin typed delegate in the generated resource. Keep the request-discriminated overloads from #1554, explicit-format pass-through, default base64 decoding, request-option precedence, logging, response identity/accessors, and exceptions unchanged. No schema, compiler, API-reference, dependency, or generation-metadata changes.

The public custom-code report keeps **34 mixed files** while reducing the embedding resource’s custom patch from **+55/−2 to +15/−2**. Total custom-patch lines fall **2,121 → 2,081**; the other **33 customizations are unchanged**. This is independent of the test-ownership cleanup in #2435.

`tests/lib/embeddings-request.test.ts` adds:

- `preserves $name encoding and response accessors`: omitted, explicit `undefined`, JavaScript `null`/empty string, `float`, and `base64`; checks the wire encoding, logs, APIPromise pass-through, raw/parsed response accessors, and request ID.
- `preserves request-option precedence without mutating caller objects`.
- `preserves response identity and sparse entries from custom parsers`.
- `preserves the API error object through the decoding wrapper`.

All nine characterization cases pass against both the original public implementation and this refactor. The existing four embedding type tests and generated embedding tests remain unchanged.

Validation: frozen pnpm 11.5.1 install; pinned formatting and full lint; TypeScript 6; build; TypeScript 4.9 published-source check; **all 609 pre-existing declaration files byte-identical**; **5,276 handwritten tests**, **559 generated tests** and 12 snapshots (one test skipped); packed CJS/ESM and source checks on Node **22.0.0**; publint (only its existing vendored-export warning).

### DIFF
--- a/src/lib/embeddings.ts
+++ b/src/lib/embeddings.ts
@@ -1,0 +1,61 @@
+import type { OpenAI } from '../client';
+import type { APIPromise } from '../core/api-promise';
+import type { RequestOptions } from '../internal/request-options';
+import { loggerFor, toFloat32Array } from '../internal/utils';
+import type { CreateEmbeddingResponse, Embedding, EmbeddingCreateParams } from '../resources/embeddings';
+
+type Base64EmbeddingResponse = Omit<CreateEmbeddingResponse, 'data'> & {
+  data: (Omit<Embedding, 'embedding'> & { embedding: string })[];
+};
+
+/**
+ * Sends the optimized embeddings request while preserving explicit encodings and
+ * the original APIPromise response accessors.
+ *
+ * @internal
+ */
+export function createEmbedding(
+  client: OpenAI,
+  body: EmbeddingCreateParams,
+  options?: RequestOptions,
+): APIPromise<CreateEmbeddingResponse | Base64EmbeddingResponse> {
+  const hasUserProvidedEncodingFormat = !!body.encoding_format;
+  // No encoding_format specified, defaulting to base64 for performance reasons.
+  // See https://github.com/openai/openai-node/pull/1312.
+  const encodingFormat = hasUserProvidedEncodingFormat ? body.encoding_format : 'base64';
+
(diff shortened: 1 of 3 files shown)


src/lib/embeddings.ts
```diff
@@ -0,0 +1,61 @@
...
+  }
+
+  // The default request uses base64 on the wire, but returns the API's default
+  // numeric embedding representation to the caller.
+  loggerFor(client).debug('embeddings/decoding base64 embeddings from base64');
+
+  return response._thenUnwrap((data) => {
+    if (data && data.data) {
+      // Preserve skipped holes when a custom response parser returns a sparse array.
+      // oxlint-disable-next-line unicorn/no-array-for-each
```
jbeckwith-oai: Avoid introducing `// oxlint-disable-next-line unicorn/no-array-for-each` here. The original generated implementation needed no suppression, and an indexed `for` loop with `if (!(index in data.data)) continue` preserves `forEach`’s sparse-hole behavior while satisfying the linter. Plain `for...of` is not equivalent because it visits holes.

jbeckwith-oai (changes requested):
The extraction preserves embedding behavior, but it introduces an avoidable lint suppression in SDK-owned code. Please replace the suppressed `forEach` with straightforward indexed iteration that skips sparse-array holes; this preserves the existing semantics without weakening lint policy.

jbeckwith-oai (approved):
Re-reviewed the updated embedding helper: the previously requested lint suppression is removed, and bounded indexed iteration preserves original `forEach` semantics for sparse holes, inherited entries, and array mutation, with dedicated regression coverage. Request encoding, response identity, caller options, and error behavior are unchanged; CI is green.

PR #2437, opened 2026-08-20
refactor: move vector store polling into an SDK-owned helper

apcha-oai:
Move the existing vector-store file and file-batch polling loops into one SDK-owned helper, leaving the public resource methods as thin typed delegates. Preserve the existing terminal-status sets (including their different cancellation behavior), interval defaults and parsing, request-option/header precedence, resource overrides, and errors. Upload concurrency and create/upload helpers are unchanged. No schema, compiler, API-reference, dependency, or generation-metadata changes.

The hash-verified public custom-code report keeps **34 mixed files** while reducing total custom-patch lines **2,121 → 2,039**. File batches fall **+118/−0 → +78/−0** and files **+89/−1 → +48/−0**; the other **32 customizations are unchanged**. This PR is independent of #2435 and #2436.

The 18 new public-entrypoint characterization cases in `tests/lib/vector-store-polling.test.ts` pass against both unmodified public-main source and this refactor. They cover zero/negative/NaN intervals, decimal/hexadecimal/zero/empty server headers, caller-header precedence, unknown and cancelled statuses, and unchanged retrieval errors. Together with the existing generated-resource helper suite, all **55 focused cases** pass.

Validation: frozen pnpm 11.5.1 install; pinned formatting and full lint; TypeScript 6; build; TypeScript 4.9 published-source check; **609 existing declaration files unchanged except equivalent generated import ordering in two files**; **5,285 handwritten tests**, **559 generated tests** and 12 snapshots (one test skipped); packed CJS/ESM and source checks on Node **22.0.0**; publint (only its existing vendored-export warning).

### DIFF
--- a/src/lib/vector-store-polling.ts
+++ b/src/lib/vector-store-polling.ts
@@ -1,0 +1,88 @@
+import type { APIPromise } from '../core/api-promise';
+import { buildHeaders } from '../internal/headers';
+import type { NullableHeaders } from '../internal/headers';
+import type { RequestOptions } from '../internal/request-options';
+import { sleep } from '../internal/utils/sleep';
+import type { FileBatches, VectorStoreFileBatch } from '../resources/vector-stores/file-batches';
+import type { Files, VectorStoreFile } from '../resources/vector-stores/files';
+
+type PollOptions = RequestOptions & { pollIntervalMs?: number };
+
+async function poll<T extends { status: string }>(
+  retrieve: (headers: NullableHeaders) => APIPromise<T>,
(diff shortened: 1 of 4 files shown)


jbeckwith-oai (approved):
Independently reviewed the shared vector-store file/batch polling helper. Terminal-status behavior, request options and headers, interval parsing including legacy hexadecimal autodetection, sequential requests, cancellation compatibility, and error identity are preserved. Narrow lint suppressions are explicitly documented and necessary for those existing semantics; focused tests and CI are green.

PR #2438, opened 2026-08-20
refactor: move file processing wait into an SDK-owned helper

apcha-oai:
Move the existing `files.waitForProcessing` loop into an SDK-owned helper, leaving the generated resource method's signature and defaults unchanged. Preserve the terminal states, original return object, polling intervals, timeout boundary and ordering, and retrieval errors. In particular, an initially terminal file still returns without a timeout check, while a later terminal response can still time out. No schema, compiler, API-reference, dependency, or generation-metadata changes.

The hash-verified public custom-code report keeps **32 mixed files** while reducing the file resource's custom patch **+28/−0 → +11/−0**. Total custom-patch lines fall **1,961 → 1,944**; the other **31 customizations are unchanged**.

The 16 new public-entrypoint characterization cases in `tests/lib/file-processing.test.ts` passed against the original implementation before extraction. Together with the existing SDK behavior suite, all **27 focused cases** pass. Coverage includes initially processed/error/deleted files, default/zero/negative/NaN intervals, missing and unknown statuses, default/custom timeout boundaries, timeout-after-retrieval ordering, and error identity.

Validation on the refreshed public-main base: frozen pnpm 11.5.1 install; pinned formatting and full lint; TypeScript 6; build; TypeScript 4.9 published-source check; **all 611 pre-existing declaration files byte-identical**; **5,307 handwritten tests**, **556 generated tests** and 12 snapshots (one test skipped); packed CJS/ESM and source checks on Node **22.0.0**; publint (only its existing vendored-export warning).

### DIFF
--- a/src/lib/file-processing.ts
+++ b/src/lib/file-processing.ts
@@ -1,0 +1,36 @@
+import { APIConnectionTimeoutError } from '../error';
+import { sleep } from '../internal/utils/sleep';
+import type { FileObject, Files } from '../resources/files';
+
+/**
+ * Waits for file processing through the resource's retrieve method. The timeout
+ * is checked after each subsequent retrieval, preserving the existing behavior
+ * for an initially terminal file and for a terminal response received too late.
+ *
+ * @internal
+ */
+export async function waitForFileProcessing(
+  resource: Pick<Files, 'retrieve'>,
+  id: string,
+  pollInterval: number,
+  maxWait: number,
+): Promise<FileObject> {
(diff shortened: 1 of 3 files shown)


jbeckwith-oai (approved):
Independently verified the SDK-owned file-processing helper preserves public signatures, retrieval ordering, terminal statuses, timeout boundaries, polling-interval edge cases, error identity, and responses. The two narrowly documented await-in-loop suppressions are necessary for existing sequential polling semantics; focused coverage and CI are green.

PR #2439, opened 2026-08-20
chore(api): clarify image transparency and audio delta docs

apcha-oai:
## Summary

Clarify image-background documentation and Realtime translation audio-delta sizing. This update changes comments and the transformed API reference only; public TypeScript interfaces and runtime behavior are unchanged.

## Changes

- Document transparent backgrounds as preview support for `gpt-image-2` and `gpt-image-2-2026-04-21`, with `png` or `webp` required for transparent output.
- Clarify that Realtime translation audio deltas contain variable-length PCM16 chunks that clients should decode and queue in full.
- Preserve the existing image-background values and nullability; `null` was already accepted by the image edit/generate interfaces.


jbeckwith-oai (changes requested):
The updated description removes the existing parameter-level restriction that `background` is supported only for GPT Image models. It now limits only transparent backgrounds, incorrectly implying `dall-e-2` / `dall-e-3` accept `background: "opaque"` or `"auto"`. Restore the explicit GPT-Image-only restriction in the canonical OpenAPI description while retaining the new transparency preview and PNG/WebP guidance, then regenerate. The existing review thread already covers this; no duplicate inline comment.

apcha-oai:
overriding as docstring only change; api ref authoritative here

PR #2440, opened 2026-08-20
refactor: share SDK-owned assistant run polling

apcha-oai:
Move the assistant-run polling loop into an SDK-owned delegate and reuse the polling mechanism introduced in #2437. The existing vector-store helper exports remain intact; only their private shared mechanism moves to `src/lib/polling.ts`. Preserve assistant and vector-store status handling, interval parsing, request options, the assistant params/header-wrapper behavior, returned object identity, and errors. Streaming helpers and legacy types are unchanged. Also restore the equivalent generated `RunStepInclude` export ordering.

The hash-verified public custom-code report keeps **32 mixed files** while reducing total custom-patch lines **1,961 → 1,916**. The run resource falls **+134/−1 → +91/−1** and its barrel **+6/−1 → +5/−0**; the other **30 customizations are unchanged**. No schema, compiler, API-reference, dependency, or generation-metadata changes.

The eight new public-entrypoint characterization cases in `tests/lib/assistant-run-polling.test.ts` pass against both the original implementation and this extraction. Combined with the existing vector-store and generated-resource helper suites, all **63 focused cases** pass. Coverage includes interval edge cases, caller-header precedence and wrapper identity, params identity, unknown statuses, terminal states, and unchanged retrieval errors.

Validation: frozen pnpm 11.5.1 install; pinned formatting and full lint; TypeScript 6; build; TypeScript 4.9 published-source check; **611 existing declarations unchanged except the equivalent type-export ordering in two files** (including byte-identical vector-store helper declarations); **5,299 handwritten tests**, **556 generated tests** and 12 snapshots (one test skipped); packed CJS/ESM and source checks on Node **22.0.0**; publint (only its existing vendored-export warning).


jbeckwith-oai (approved):
Independently verified shared Assistant and vector-store polling preserves all intermediate/terminal states, request/header precedence, interval edge cases, retrieve overrides, response identity, and error behavior. Existing justified lint suppressions are relocated unchanged; focused compatibility coverage and CI are green.

PR #2441, opened 2026-08-21
refactor: move vector store batch uploads into an SDK-owned helper

apcha-oai:
Move vector-store batch upload orchestration into an SDK-owned helper while retaining the exact public `uploadAndPoll` signature and defaults. Preserve the shared worker iterator, concurrency limits (including native errors for invalid values), completion-order file IDs, original uploadables and request options, resource overrides, and batch/error identity. The existing all-workers-settle error aggregator is unchanged, including its ordered, non-enumerable rejection causes and quiet logging behavior.

The hash-verified public custom-code report keeps **32 mixed files** while reducing total custom-patch lines **1,961 → 1,932**. The file-batch resource falls **+78/−0 → +49/−0**; the other **31 customizations are unchanged**. No schema, compiler, API-reference, dependency, or generation-metadata changes.

All **14 new public-entrypoint characterization cases** pass against both exact public `main` and this extraction. Together with the existing generated-resource helper and error-privacy suites, **55 focused cases** pass.

Validation: frozen pnpm 11.5.1 install; canonical formatting and full lint; TypeScript 6; build; TypeScript 4.9 published-source check; **all 611 existing declaration files byte-identical**; **5,305 handwritten tests**, **556 generated tests** and 12 snapshots (one test skipped); packed CJS/ESM and source checks on Node **22.0.0**; publint (only its existing vendored-export warning).

### DIFF
--- a/src/lib/vector-store-upload.ts
+++ b/src/lib/vector-store-upload.ts
@@ -1,0 +1,52 @@
+import type { OpenAI } from '../client';
+import type { RequestOptions } from '../internal/request-options';
+import type { FileBatches, VectorStoreFileBatch } from '../resources/vector-stores/file-batches';
+import type { Uploadable } from '../uploads';
+import { allSettledWithThrow } from './Util';
+
+type UploadOptions = RequestOptions & { pollIntervalMs?: number; maxConcurrency?: number };
+
+/**
+ * Uploads files with a shared iterator, then creates and polls the batch. Every
+ * worker settles before upload failures are propagated.
+ *
+ * @internal
+ */
+export async function uploadAndPollVectorStoreFileBatch(
+  resource: Pick<FileBatches, 'createAndPoll'>,
+  client: Pick<OpenAI, 'files'>,
+  vectorStoreId: string,
+  files: Uploadable[],
+  fileIds: string[],
+  options?: UploadOptions,
+): Promise<VectorStoreFileBatch> {
+  if (files === null || files === undefined || files.length === 0) {
+    throw new Error(
+      "No `files` provided to process. If you've already uploaded files you should use `.createAndPoll()` instead",
+    );
+  }
+
+  const configuredConcurrency = options?.maxConcurrency ?? 5;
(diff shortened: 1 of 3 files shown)


src/lib/vector-store-upload.ts
```diff
@@ -0,0 +1,52 @@
...
+  // https://stackoverflow.com/questions/40639432/what-is-the-best-way-to-limit-concurrency-when-using-es6s-promise-all
+  async function processFiles(iterator: IterableIterator<Uploadable>) {
+    for (const item of iterator) {
+      // oxlint-disable-next-line no-await-in-loop -- Each worker uploads one file at a time.
+      const fileObj = await client.files.create({ file: item, purpose: 'assistants' }, options);
+      allFileIds.push(fileObj.id);
+    }
+  }
+
+  // oxlint-disable-next-line unicorn/no-new-array -- Preserve native RangeError for negative, NaN, or fractional limits.
```
jbeckwith-oai: Avoid introducing `// oxlint-disable-next-line unicorn/no-new-array`: a straightforward compliant form preserves the same native array-length validation and exact `RangeError` behavior for negative, fractional, NaN, infinite, and out-of-range concurrency limits:

```ts
const workers: Promise<void>[] = [];
workers.length = concurrencyLimit;
```

The existing indexed worker initialization and sparse-array semantics remain unchanged, so the suppression can be removed without changing compatibility.

jbeckwith-oai (changes requested):
The worker extraction preserves concurrency and error semantics, but it adds an avoidable `unicorn/no-new-array` lint suppression. Initializing an empty typed array and assigning its `length` preserves the identical native array-length validation and RangeError behavior; please use that compliant form and remove the suppression.

jbeckwith-oai (approved):
Re-reviewed the updated concurrent batch-upload helper. The previously requested array-constructor lint suppression is removed; typed empty-array initialization plus explicit length assignment preserves exact native concurrency-limit `RangeError` behavior, sparse worker initialization, completion order, all-workers-settle failures, private error identity, and request options. Expanded edge-case coverage and CI are green.

PR #2446, opened 2026-08-21
chore: set a 3,000-line custom-code budget

apcha-oai:
## Summary

Set a 3,000-line ceiling on additions **plus** deletions in the remaining custom patch against verified generated output. This is the budget-only foundation for the same rollout approved in [Python #3714](https://github.com/openai/openai-python/pull/3714) and [Python #3715](https://github.com/openai/openai-python/pull/3715).

The verified current patch is **+1,589 / −181 = 1,770 lines**, leaving **1,230 lines** of headroom. This is a generous bound on accidental growth, not a target or permission to add custom code.

Only `.castiron-ratchet.json` changes here. The stacked tooling PR will use the base-branch allowance, so a PR cannot raise its own budget. Future increases require a separate budget-only PR, explicit justification, and a human approving review; agents cannot approve them.

No workflow, CODEOWNERS, or ruleset changes are included. Ruleset enforcement is a separate, later decision.

Tooling companion: https://github.com/openai/openai-node/pull/2447

PR #2447, opened 2026-08-21
ci: check custom-code budgets against main's policy

apcha-oai:
## Summary

Catch accidental growth in the SDK's remaining generated-file customization before it becomes a large patch to maintain. This ports the same layout and tooling approved in [Python #3714](https://github.com/openai/openai-python/pull/3714) and [Python #3715](https://github.com/openai/openai-python/pull/3715), including Python's trusted report computation.

The existing workflow pair measures additions **plus** deletions against verified generated output, using the checker and budget from main. A PR cannot increase its own allowance; budget changes must be isolated and explicitly justified, and increases require a human approving review. Agents cannot approve or bypass increases.

Technical details live in `scripts/castiron/CUSTOM_CODE.md`, with references from AGENTS and CONTRIBUTING. No extra workflow files, SDK API changes, checkpoint edits, exclusion changes, or CODEOWNERS changes. **Ruleset enforcement is intentionally deferred.**

The integrated tooling measures **+1,955 / −208 = 2,163 custom lines**, below the **3,000-line** ceiling. This includes the reporter changes themselves.

## Stack

The [budget-only foundation](https://github.com/openai/openai-node/pull/2446) has merged. This tooling commit is now restacked onto `main` at `d1410d190d04`; the tooling patch is unchanged and the PR contains no budget-file change. CI is rerunning against main. The trusted statuses become active after the tooling lands, but will not be required until a separately authorized ruleset rollout.

### DIFF
--- a/.github/workflows/castiron-custom-code-comment.yml
+++ b/.github/workflows/castiron-custom-code-comment.yml
@@ -13,9 +13,168 @@ concurrency:
   cancel-in-progress: false
 
 jobs:
+  compute:
+    name: Compute trusted custom-code report
+    if: contains(fromJSON('["pull_request", "merge_group"]'), github.event.workflow_run.event) && github.event.workflow_run.path == '.github/workflows/castiron-custom-code.yml'
+    runs-on: ubuntu-latest
+    timeout-minutes: 10
+    permissions:
+      contents: read
+      actions: read
+      pull-requests: read
+    outputs:
+      number: ${{ steps.report.outputs.number }}
+      artifact-id: ${{ steps.artifact.outputs.artifact-id }}
+      artifact-run-attempt: ${{ github.run_attempt }}
+      base-sha: ${{ steps.budget.outputs.base_sha }}
+      head-sha: ${{ steps.budget.outputs.head_sha }}
+      isolation: ${{ steps.budget.outputs.isolation }}
+      budget: ${{ steps.budget.outputs.budget }}
+    steps:
+      - name: Check out the trusted reporter
+        uses: actions/checkout@de0fac2e4500dabe0009e67214ff5f5447ce83dd # v6
+        with:
+          ref: main
+          persist-credentials: false
+
+      - name: Compute from the current pull request Git objects
+        id: report
+        if: github.event.workflow_run.event == 'pull_request'
+        env:
+          GH_TOKEN: ${{ github.token }}
+          GIT_CONFIG_COUNT: '2'
+          GIT_CONFIG_KEY_0: credential.helper
+          GIT_CONFIG_VALUE_0: ''
+          GIT_CONFIG_KEY_1: credential.https://github.com.helper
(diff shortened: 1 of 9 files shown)


.github/workflows/castiron-custom-code-comment.yml
```diff
@@ -13,9 +13,168 @@ concurrency:
...
+            const url = `${context.serverUrl}/${context.repo.owner}/${context.repo.repo}/actions/runs/${context.runId}`;
+            for (const [name, result] of [
+              ['Castiron / budget-only change', process.env.ISOLATION_RESULT],
+              ['Castiron / custom-code budget', process.env.BUDGET_RESULT],
+            ]) {
+              const state = fresh && result === 'success' ? 'success' : 'failure';
+              const description = !fresh ? 'Evaluation unavailable or base changed; rerun against current main.'
+                : state === 'success' ? 'Passed against main policy. See the trusted run summary.'
+                : 'Budget check failed. See the trusted run summary.';
+              await github.rest.repos.createCommitStatus({...context.repo, sha: head, context: name,
```
jbeckwith-oai, line 170: P2 — Prevent older source workflows from overwriting newer commit statuses. This publisher unconditionally calls createCommitStatus for the same SHA/context without comparing the source workflow run ID and run attempt with the most recent published result. If an older source run is rerun or handlers complete out of order, its transient failure can replace a newer successful status and incorrectly block the PR or merge queue; branch-level concurrency serializes handlers but does not establish source-run freshness. Apply the same monotonic (run_id, run_attempt) guard used by the comment publisher before writing either status, and add an out-of-order/rerun regression.

jbeckwith-oai (changes requested):
Please address the inline P2: enforce monotonic source workflow run/attempt ordering before publishing commit statuses so older reruns cannot replace newer successful merge-gate results. Trusted policy isolation and budget calculation otherwise look sound.

.github/workflows/castiron-custom-code-comment.yml
apcha-oai: We're accepting this fail-closed availability edge case for the initial rollout and leaving the implementation unchanged.

Older source reruns recompute against current trusted inputs; they don't replay stale budget results. A later transient evaluation failure can replace green statuses with red and bounce an otherwise valid candidate, but this is recoverable by rerunning the latest workflow or requeuing. The merge queue does not prevent that false block; it remains the boundary for checking the combined candidate against current main once the budget statuses are required.

We don't think source-run ordering state is justified without evidence of recurring disruption in normal queue operation. In particular, source run ID alone is not policy freshness: an older source run can legitimately be rerun against a newer main/policy revision. This also keeps the rollout aligned with the approved Python workflow.

We'll revisit if this becomes a recurring operational problem. No code, budget, ruleset, or review-state changes are being made as part of this response.

jbeckwith-oai (approved):
Approved at the reviewer's request.

PR #2448, opened 2026-08-21
chore(api): update SDK generation metadata

apcha-oai:
## Summary

Modify only generation and build metadata; no changes to public API or runtime behavior. Consumers will see no differences in types, methods, parameters, or return values.

## Changes

- No additions, removals, or modifications to public types, methods, parameters, or return types.
- No changes to runtime behavior, error semantics, or observable SDK functionality.
- Only generation and build metadata files were updated; no consumer-facing code was modified.


jbeckwith-oai (approved):
Approved after two independent reviews: generation metadata only, unchanged generated content hash, no runtime/public API/workflow/security changes, and passing public checks.

PR #2461, opened 2026-08-24
feat(api): Add project residency and usage quantity_unit fields

apcha-oai:
## Summary

Add optional typed residency support to project models and project creation, while retaining the existing optional nullable geography inputs for both creation and update. Geography is deprecated, with its original descriptions preserved and migration guidance appended. Add optional quantity_unit fields to organization usage cost results, including known units and unknown string values.

## Changes

- Add the residency configuration enum to Project and ProjectCreateParams; residency remains a create-time input, not a new update parameter.
- Preserve geography?: string | null on ProjectCreateParams and ProjectUpdateParams, marked deprecated. Existing callers can continue using geography; creation must not supply both geography and residency.
- Add optional quantity_unit fields to cost results carried by the usage response types.

Project responses and project-creation inputs now reuse a shared residency model. Usage cost responses reuse a shared quantity-unit model, with unknown unit values and optional/null fields preserved.


jbeckwith-oai (changes requested):
Requesting changes for the existing P2 finding in https://github.com/openai/openai-node/pull/2461#discussion_r3848959931: the transformed contract and generated test permit simultaneous geography and residency despite explicitly forbidding that combination, allowing mocks to accept requests the API rejects. Encode mutual exclusion upstream and generate separate valid legacy/new-residency fixtures.

jbeckwith-oai (approved):
LGTM.

PR #2467, opened 2026-08-25
fix(uploads): preserve Blob files and multipart media types

HAYDEN-OAI:
## Summary

- Keep ordinary unnamed Blob uploads in buffered and streaming multipart requests so public image-edit inputs and masks are never silently omitted.
- Preserve standard FormData Blob filenames, existing streaming fallbacks, safe custom filenames, sparse input ordering, and lazy multipart consumption.
- Retain fetched Response MIME types for public image and audio-transcription uploads.

## Validation

- Added eight regression-first public-client and multipart tests; all eight fail on the original code and pass with this change.
- Full handwritten suite: 6,195 passing tests.
- Full generated suite against an isolated owner-verified Steady instance: 556 passing tests, one pre-existing skip.
- Canonical lint, strict repository typecheck, build, published TypeScript 4.9 and current TypeScript checks, and publint all pass.
- Packed npm verification passes 275/314 source checks across 1,256 source maps.
- Confirmed the actual built npm client preserves Blob image/mask bytes and audio/wav Response uploads.

### DIFF
--- a/src/internal/uploads.ts
+++ b/src/internal/uploads.ts
@@ -121,7 +121,7 @@ export const checkFileSupport = () => {
 /**
  * Values accepted by SDK methods that upload multipart files.
  *
- * Supports native files, fetch responses, named blobs, Node.js filesystem read
+ * Supports native files, fetch responses, blobs, Node.js filesystem read
  * streams, async byte sources, web readable streams, and files created with
  * {@link toStreamingFile}. Use {@link toFile} to materialize compatible content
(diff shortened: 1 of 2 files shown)


jbeckwith-oai (approved):
Reviewed the Blob/multipart changes and public-client regressions. Anonymous Blob uploads, safe explicit filename/path handling, Response-derived media types, sparse ordering, mixed lazy streaming, image masks, audio uploads, and unchanged credential boundaries are all covered; CI is green. Approved.

PR #2477, opened 2026-08-25
fix(polling): honor cancellation during polling delays

HAYDEN-OAI:
## Summary

- Honor the documented caller `RequestOptions.signal` immediately during intermediate assistant-run and vector-store polling delays, including default, server-provided, and custom intervals.
- Reject with `APIUserAbortError` and preserve the exact caller abort reason as its non-enumerable `cause`; safely clear the timer and only the polling-owned listener across completion, pre-abort, missed-event, and synchronous registration races.
- Preserve existing terminal/unknown status handling, header and getter ordering, no-signal sleep behavior, unrelated abort listeners, and structural AbortSignal compatibility.
- Exercise all three actual public polling methods and all seven public create, upload, and submit-tool-output polling delegates.

## Validation

- Node 22.22.0 and Node 24.19.0: 96 focused tests on each runtime.
- Full handwritten suite: 149 files and 6,314 passing tests.
- Full generated suite: 82 files and 556 passing tests against an isolated mock server.
- Repository formatting/lint, strict TypeScript, package build, published TypeScript 4.9 and current TypeScript declarations, and publint.
- Existing packed-package CommonJS/ESM and source-map checks, plus actual built CommonJS and ESM public polling cancellation regressions.

### DIFF
--- a/src/lib/polling.ts
+++ b/src/lib/polling.ts
@@ -14,6 +14,14 @@ function sleepUntilAborted(milliseconds: number, signal: AbortSignal): Promise<v
     let registered: (() => void) | undefined;
     let settled = false;
 
+    const removeAbortListener = (listener: () => void) => {
+      try {
+        signal.removeEventListener('abort', listener);
+      } catch {
+        // Caller-controlled cleanup must never prevent the polling promise from settling.
+      }
+    };
+
     const cleanup = () => {
       if (timer !== undefined) {
         clearTimeout(timer);
@@ -22,7 +30,7 @@ function sleepUntilAborted(milliseconds: number, signal: AbortSignal): Promise<v
       if (registered) {
         const listener = registered;
         registered = undefined;
-        signal.removeEventListener('abort', listener);
+        removeAbortListener(listener);
       }
     };
 
@@ -66,13 +74,13 @@ function sleepUntilAborted(milliseconds: number, signal: AbortSignal): Promise<v
       signal.addEventListener('abort', abort, { once: true });
 
       if (settled) {
-        signal.removeEventListener('abort', abort);
+        removeAbortListener(abort);
       } else if (signal.aborted) {
         abort();
       }
     } catch (error) {
       if (settled) {
-        signal.removeEventListener('abort', abort);
+        removeAbortListener(abort);
       } else {
         settled = true;
         cleanup();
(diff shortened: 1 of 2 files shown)


src/lib/polling.ts
```diff
@@ -1,16 +1,93 @@
...
+
+    const cleanup = () => {
+      if (timer !== undefined) {
+        clearTimeout(timer);
+        timer = undefined;
+      }
+      if (registered) {
+        const listener = registered;
+        registered = undefined;
+        signal.removeEventListener('abort', listener);
```
jbeckwith-oai: **[P2] Always settle polling when structural signal listener removal throws**

This PR explicitly supports structural/polyfilled `AbortSignal`s, but their `removeEventListener` is caller-provided and may throw. On cancellation, `abort()` sets `settled = true` and `cleanup()` clears the timer before this call; if removal throws, execution never reaches `reject(error)`, so `runs.poll`, `files.poll`, `fileBatches.poll`, and delegated helpers remain pending forever. The normal timer callback likewise invokes `cleanup()` before `resolve()`, producing the same permanently pending state and potentially an uncaught timer exception. Reproduce with a structural wrapper around a real signal whose `removeEventListener()` throws, start a public poll, and abort: the timer disappears while the returned promise never settles. Please guard listener removal and guarantee the abort rejection/normal resolution still runs, with focused public-method regressions for both paths.

jbeckwith-oai (changes requested):
The cancellation and race coverage is otherwise strong, but supported structural/polyfilled AbortSignals can throw while removing the polling listener. That currently clears the only recovery timer without settling the poll, leaving every public polling helper permanently pending. Please make cleanup exception-safe and add abort and ordinary timer-completion regressions before landing.

src/lib/polling.ts
HAYDEN-OAI: Fixed in 87bb5fcfc022f3f0e2877301f44773a332ef946e. All three listener-removal paths now preserve abort rejection, timer completion, and original registration errors. Added 23 public-boundary regressions across all three pollers, seven delegated helpers, duplicate events, and registration races. Verified Node 22.22.2/24.19.0: 134 focused tests each; 6,335 applicable unit + 556 generated tests, lint, TypeScript 4.9/6, build, and publint pass (2 Linux-/proc-only tests skipped on macOS).

jbeckwith-oai (approved):
Re-reviewed the updated polling cancellation implementation and confirmed the previously reported permanently pending poll is fixed: structural AbortSignal listener-removal failures are guarded across cancellation, timer settlement, and registration races while preserving abort causes and public polling behavior.

PR #2481, opened 2026-08-26
fix(auth): honor an explicit zero GCP metadata timeout

morgan-coded:

## Changes being requested

`gcpIDTokenProvider` computes its metadata request timeout with `config?.timeout || 10_000`, so an explicit `timeout: 0` is silently replaced with 10,000 ms. The Azure IMDS provider in the same file uses `config?.timeout ?? 10_000` for the identically documented option, so the two providers currently behave differently for the same configuration.

This change switches the GCP provider to the nullish coalescing the file already uses for Azure, and adds explicit-zero regression tests for both providers in `tests/auth/subject-token-providers.test.ts` (mocked fetch, no live network; the Azure test is a parity control pinning the already-correct behavior).

Verified locally (Node 24.18.0): the GCP regression fails on `main` (the single `setTimeout` call receives `10000` instead of `0`) and passes with this change; the provider test file passes 14/14; lint, build, and the full test suite pass apart from the two pre-existing macOS-only failures addressed in #2480.

## Additional context & links

One-line production change; the only other `config?.timeout` default in the file already used `??`.

### DIFF
--- a/src/auth/subject-token-providers.ts
+++ b/src/auth/subject-token-providers.ts
@@ -370,7 +370,7 @@ export function gcpIDTokenProvider(
     fetch?: Fetch;
   },
 ): SubjectTokenProvider {
-  const timeout = config?.timeout || 10_000;
+  const timeout = config?.timeout ?? 10_000;
 
   return {
     tokenType: 'id',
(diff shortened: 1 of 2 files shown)


HAYDEN-OAI (approved):
Reviewed the GCP metadata timeout handling. Explicit timeout: 0 is now preserved, the 10-second default remains unchanged, and the behavior matches the existing Azure provider. No actionable concerns.

PR #2483, opened 2026-08-26
fix(standard-schema): preserve validator bindings across helpers

HAYDEN-OAI:
## Summary

- Bind all four Standard Schema helpers to the same validator metadata used for strict schema generation.
- Preserve explicit-schema laziness, original method receivers, and existing tool-option getter ordering.
- Add focused integrated regressions for fresh metadata, converter mutation, validator swapping, and malformed JSON.

## Validation

- Node 22.23 and Node 26.7: 6,541 handwritten and 556 generated tests on each runtime.
- Canonical lint, strict TypeScript 6, published TypeScript 4.9/6 declarations, and package builds on both runtimes.
- publint and installed packed-package CJS/ESM adversarial helper checks on both runtimes.

### DIFF
--- a/src/helpers/standard-schema.ts
+++ b/src/helpers/standard-schema.ts
@@ -505,12 +505,30 @@ function normalizeStructuredOutputSchema(schema: JSONSchema): JSONSchema {
   return normalizedSchema;
 }
 
+type StandardSchemaBinding<Schema extends StandardSchemaLike> = {
+  standard: Schema['~standard'];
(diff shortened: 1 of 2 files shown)


jbeckwith-oai (approved):
Reviewed all four Standard Schema helper surfaces and independently validated lazy single-read metadata/validator binding, correct validator receiver semantics via Reflect.apply, explicit-schema lazy parsing, reentrant option/metadata getters, unchanged async error handling, and focused regression coverage.

PR #2486, opened 2026-08-26
fix: publish Castiron checks for behind-main pull requests

HAYDEN-OAI:
## Summary

Fix Castiron's required budget statuses getting stuck in the expected state whenever an otherwise valid pull request is behind `main`.

- Authenticate the pull request by its exact head commit, source repository, upstream repository, and target branch instead of requiring its recorded base SHA to equal the latest `main` commit.
- Evaluate behind-main pull requests against the current trusted `main` policy by recomputing their report inside the existing trusted, bare Git object store; reuse an existing report only when its authenticated base is current.
- Publish both required statuses for valid behind-main pull requests and publish explicit failures if `main` changes between trusted evaluation and publication.
- Add executable regressions covering behind-main status publication, a `main` race, trusted recomputation, and rejection of a report bound to the wrong pull-request base.

This fixes the failure observed on #2468, #2478, and #2482, where the baseline workflow passed but `Castiron / budget-only change` and `Castiron / custom-code budget` were never published.

## Security

The trusted workflow still checks out and executes only current `main`, verifies the immutable candidate head and source repository, requires exactly one valid upstream pull request, evaluates the policy from current `main`, preserves merge-queue enforcement, and never executes candidate code or consumes candidate artifacts. Results are published only when their evaluated head and current policy revision still match; otherwise both required statuses fail closed.

## Verification

- `PYTHONDONTWRITEBYTECODE=1 python3 -m unittest discover -s scripts/castiron -p 'test_custom_code*.py'`
- `ruff format --check scripts/castiron/custom_code_budget.py scripts/castiron/test_custom_code_budget.py`
- `go run github.com/rhysd/actionlint/cmd/actionlint@v1.7.12 .github/workflows/castiron-custom-code.yml .github/workflows/castiron-custom-code-comment.yml`
- `git diff --check`

### DIFF
--- a/.github/workflows/castiron-custom-code-comment.yml
+++ b/.github/workflows/castiron-custom-code-comment.yml
@@ -165,7 +165,8 @@ jobs:
               const current = [];
               for (const number of numbers) {
                 const {data: pr} = await github.rest.pulls.get({...context.repo, pull_number: number});
-                if (pr.number === number && pr.state === 'open' && pr.head.sha === head && pr.base.sha === base &&
+                if (pr.number === number && pr.state === 'open' && pr.head.sha === head &&
+                    /^[0-9a-f]{40}$/.test(pr.base.sha) &&
                     pr.head.repo?.full_name === source.full_name && pr.base.ref === 'main' &&
                     pr.base.repo.full_name === upstream) current.push(pr);
               }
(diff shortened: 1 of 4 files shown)


jbeckwith-oai (approved):
Reviewed behind-main Castiron evaluation against the current main policy, trusted PR/head/base identity, isolated Git-object handling, fail-closed status publication, and regression coverage. Looks good.

PR #2488, opened 2026-08-26
feat(api): make function call output call IDs optional

apcha-oai:
## Summary

Align function call output types with the API schema: `call_id` is optional in stable and beta input and response types, and input types also accept `null`.


jbeckwith-oai (approved):
Verified stable and beta response-output call_id is correctly optional/non-nullable, input call_id is optional/nullable, existing supplied IDs remain supported, and the generated changes match the coordinated canonical API schema. Looks good.

PR #2490, opened 2026-08-27
fix(auth): clamp the workload-identity refresh buffer to the token lifetime

morgan-coded:

## Changes being requested

`WorkloadIdentityAuth` starts its proactive background refresh when a cached token is within `refreshBufferSeconds` (default 1,200) of expiry, but the buffer is never limited by the token's actual lifetime. Any token issued with `expires_in` at or below the buffer is inside the refresh window the moment it is cached, so every `getToken()` call starts another background token exchange: one serialized OAuth round trip per API request for as long as the process runs. Reproduced deterministically: with `expires_in: 600` and the default buffer, four sequential `getToken()` calls perform four exchanges.

The X.509 workload-identity path already handles this by capping the buffer at half the actual token lifetime (`Math.min(refreshBufferMs, lifetime / 2)`, documented in the README as "capped at half of the actual token lifetime"). This change applies the same cap to the subject-token flavor: `refreshAt` is computed once at cache time as `expiresAt` minus the smaller of the configured buffer and half the lifetime, and `needsRefresh` compares against that precomputed instant. Normal-lifetime tokens are unaffected: a 3,600-second token with the default buffer still begins background refresh at 2,400 seconds, exactly as before.

Tests: a regression pins the cap from both sides for a 600-second token (served from cache one millisecond before half-life, background refresh firing at half-life, every call resolving the cached token), and a control pins the unchanged 1,200-second window for a 3,600-second token including the millisecond-before boundary. Existing short-token tests advance their mocked clocks into the genuine refresh window; no assertion was weakened. The `refreshBufferSeconds` doc comment and README sentence now state the cap, matching the X.509 documentation.

Verified locally (Node 24.18.0): the regression fails on `main` (four exchanges) and passes with this change; lint, build, and the full test suite pass (158 files, 6,831 tests).

## Additional context & links

Found while auditing the new workload-identity surface; the fix restores parity between the two auth flavors rather than introducing a new policy.


HAYDEN-OAI (approved):
Reviewed the workload-identity refresh changes. The half-lifetime cap preserves caching for short-lived tokens while retaining existing behavior for normal token lifetimes. Expiration, concurrent refresh, and invalidation paths look correct. LGTM.

PR #2495, opened 2026-08-27
fix: restore native browser ESM imports

jbeckwith-oai:
- [x] I understand that this repository is auto-generated and my pull request may not be merged

## Changes being requested

Fixes #2494.

- Rewrite emitted ESM references to the private X.509 state package import into browser-resolvable relative imports.
- Keep one private state registry across Node CommonJS and ESM consumers while allowing native browsers to load the SDK directly.
- Preserve browser credential protections, blocked internal package exports, and compatibility with browser `process`, `module`, and `exports` shims.

This is intentionally the production-only release fix. Expanded browser and bundler regression coverage will follow separately.

## Additional context & links

Verified locally:

- Real headless Chrome directly imports the built SDK without an import map and rejects browser API-key usage by default, including with browser `process`, `module`, and `exports` shims.
- Browser-targeted ES2020 bundling succeeds.
- Node 22.0.0 preserves both mixed CommonJS/ESM X.509 authentication combinations.
- Existing packed-package checks pass, including CommonJS, ESM, browser conditions, and optional dependencies.
- `pnpm build`, `pnpm lint`, `pnpm exec tsc`, and 196 focused authentication/browser-security tests pass.
- The complete existing test suite previously passed: 7,548 tests, with 2 skipped.

Issue #2494, opened 2026-08-27
Browser direct import fails on unresolved #x509-transport-state specifier


huanglianggit:
## Confirm this is a Node library issue

- [x] This is an issue with the Node library.

## Describe the bug

The package's native-browser direct-import path fails before SDK initialization because the emitted ESM contains a bare package-internal import:

```js
import { isTransientX509ConnectionError, markApprovedX509Client } from '#x509-transport-state';
```

`package.json#imports` maps this specifier for Node and supporting bundlers, but a browser loading the built `openai/index.mjs` directly does not resolve package import maps. Chrome reports:

```
Failed to resolve module specifier #x509-transport-state. Relative references must start with either /, ./, or ../.
```

This leaves `ecosystem-tests/browser-direct-import` at `Running tests...` before any API request.

The regression is also visible in the current `main` push CI at `6f6411e0` / v7.7.0:

- [Failing CI run](https://github.com/openai/openai-node/actions/runs/33028888654)
- [Failing examples job](https://github.com/openai/openai-node/actions/runs/33028888654/job/98376539391)

## To reproduce

1. Check out `main` at `6f6411e0`.
2. Run `pnpm install --frozen-lockfile` and `pnpm build`.
3. Expose the built package under `ecosystem-tests/browser-direct-import/public/node_modules/openai`, as the existing ecosystem harness does.
4. Serve `ecosystem-tests/browser-direct-import/public` and open `index.html` in Chrome.
5. Observe the module-resolution error before any API request.

A static confirmation after the build is:

```bash
rg -n '#x509-transport-state' dist/client.mjs
```

which currently reports the unresolved import in `dist/client.mjs`.

## Expected behavior

The repository's native-browser direct-import ecosystem path should load the SDK root module without requiring a page-level import map. Node CJS/ESM behavior and the X.509 state-sharing invariant should remain intact.

## Environment

- OS: Windows 11 locally; Ubuntu GitHub Actions also reproduces the failure
- Node.js: v24.19.0 locally
- Library: current `main` / v7.7.0
- Browser: Chrome

mccrearydamion69-crypto:
Hi

PR #2496, opened 2026-08-27
test: verify packed browser bundles target ES2020

jbeckwith-oai:
## Summary

- Extend the existing packed-package integration check to install the actual npm tarball and bundle an ordinary SDK import with Vite/Rolldown for browser-targeted ES2020 ESM.
- Fail when package-internal imports cannot resolve, Node built-ins enter the browser graph, or the emitted bundle contains post-ES2020 syntax such as top-level `await`.
- Reuse the existing Vitest/Vite and ts-node/Acorn dependency trees while keeping optional peers absent and leaving package metadata, lockfiles, webpack fixtures, CI policy, and production code unchanged.

Follow-up to #2494. Its native-browser production fix is handled independently in #2495; this PR is based directly on `main` and does not include or require that patch to pass.

## Verification

- `PATH=/Users/jbeckwith/.nvm/versions/node/v24.16.0/bin:$PATH ./scripts/build`
- `PATH=/Users/jbeckwith/.nvm/versions/node/v24.16.0/bin:$PATH node --experimental-strip-types scripts/test-packed-package.ts`
- `node --experimental-strip-types scripts/test-packed-package.ts` on Node 22.22.0
- `PATH=/Users/jbeckwith/.nvm/versions/node/v24.16.0/bin:$PATH ./scripts/lint`
- `PATH=/Users/jbeckwith/.nvm/versions/node/v24.16.0/bin:$PATH node_modules/.bin/tsc`
- Injected unresolved `#package` imports, `node:fs`, and top-level `await` into the ignored built artifact and confirmed the packed-artifact check rejected all three.
- Two consecutive adversarial review rounds with two independent read-only reviewers per round reported no in-scope findings.

### DIFF
--- a/scripts/test-packed-package.ts
+++ b/scripts/test-packed-package.ts
@@ -1,5 +1,8 @@
 import type { Dirent } from 'node:fs';
 
+const packedPackageAcorn: {
+  parse: (source: string, options: { ecmaVersion: 2020; sourceType: 'module' }) => unknown;
+} = require(require.resolve('acorn', { paths: [require.resolve('ts-node/package.json')] }));
 const packedPackageAssert = require('node:assert/strict');
 const packedPackageChildProcess = require('node:child_process');
 const packedPackageFs = require('node:fs');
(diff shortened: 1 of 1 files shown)


HAYDEN-OAI (approved):
Reviewed the installed npm tarball consumer path, Vite browser resolution, rejection of Node built-ins, optional-peer isolation, and ES2020 parsing of the emitted ESM bundle. The gate exercises the real packed artifact without introducing credential or runtime dependencies.

PR #2497, opened 2026-08-27
ci: require keyless ecosystem checks on release PRs

jbeckwith-oai:
## Summary

- Run the existing credential-free ecosystem/browser job on release-please branches as well as ordinary CI runs.
- Include ecosystem checks in the existing required `test_matrix` aggregate, failing it when the ecosystem job fails, is cancelled, or is skipped.
- Preserve read-only permissions, pinned actions, concurrent jobs, fork safety, and protected-main-only live API credentials.

This is an independently landable CI-only follow-up to #2494. The production browser-import fix is tracked separately in #2495 and is intentionally not included here.

## Verification

- `./scripts/test tests/ecosystem-cli.test.ts tests/ecosystem-browser-credential-security.test.ts` (24 tests passed)
- `./scripts/lint`
- Parsed the workflow and verified credential isolation, read-only permissions, pinned actions, release-branch execution, and required aggregate dependencies.
- Executed the aggregate script with ecosystem results `success`, `failure`, `cancelled`, and `skipped`; only `success` passes.
- Completed two consecutive clean adversarial-review rounds with two independent read-only reviewers per round.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -199,7 +199,7 @@ jobs:
   test_matrix:
     name: ${{ (github.event_name == 'push' || github.event_name == 'merge_group' || github.event.pull_request.head.repo.fork || github.event.action == 'ready_for_review') && 'test matrix' || 'test matrix (not run)' }}
     if: ${{ always() && (github.event_name == 'push' || github.event_name == 'merge_group' || github.event.pull_request.head.repo.fork || github.event.action == 'ready_for_review') }}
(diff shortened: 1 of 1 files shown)


HAYDEN-OAI (approved):
Reviewed release-branch execution, required aggregate dependency and failure semantics, fork-safe read-only permissions, pinned actions, and separation from protected live API credentials. The keyless ecosystem job now correctly participates in the required release and merge checks.

PR #2498, opened 2026-08-27
test: execute webpack browser bundle without live credentials

jbeckwith-oai:

## Changes being requested

- Execute the existing `ts-browser-webpack` fixture's actual webpack output in headless Chrome during the normal credential-free ecosystem run.
- Fail immediately on bundle/module evaluation errors, block all non-file network requests, and assert that browser credentials remain disabled by default.
- Preserve the existing live browser suite, webpack/module configuration, dependencies, workflows, and SDK production code.

## Verification

- `./scripts/test tests/ecosystem-browser-credential-security.test.ts tests/ecosystem-cli.test.ts` (24 tests)
- `pnpm lint`
- `pnpm exec tsc --noEmit`
- `env -u OPENAI_API_KEY PUPPETEER_EXECUTABLE_PATH=<installed-headless-chrome> PUPPETEER_SKIP_DOWNLOAD=true pnpm tsn ecosystem-tests/cli.ts ts-browser-webpack --verbose`
- Confirmed that an injected webpack evaluation error and an injected outbound request each fail the real-browser regression.

## Additional context & links

Follow-up to #2494. The production direct-browser-import fix is tracked separately in #2495 and is intentionally not included or required for this independently passing webpack regression.

### DIFF
--- a/ecosystem-tests/cli.ts
+++ b/ecosystem-tests/cli.ts
@@ -418,6 +418,7 @@ const projectRunners = {
 
     await run('npm', ['run', 'tsc']);
     await run('npm', ['run', 'build']);
+    await run('npm', ['run', 'test:bundle']);
 
     if (state.live) {
       await run('npm', ['run', 'test:ci']);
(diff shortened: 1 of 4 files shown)


HAYDEN-OAI (approved):
Reviewed the packed webpack fixture execution in Chrome, immutable synthetic-only browser credentials, blocking and surfacing every non-file request, bundle evaluation errors, and the preserved separate live suite. The default browser API-key safeguard remains explicitly verified without exposing real credentials.

PR #2502, opened 2026-08-27
fix(streaming): surface server error frames and settle streams past throwing listeners

morgan-coded:
`ChatCompletionStream.fromReadableStream()` turned a server error frame into `TypeError: chunk.choices is not iterable`, while `AssistantStream` returned a bare `Error` and a throwing `error` or `abort` listener could leave `done()` pending forever. This converts object-valued error frames in those helper paths to `APIError` with their server payload intact and settles terminal state before rethrowing the first listener failure; later listeners remain skipped, secondary `end` listener failures are suppressed, and the caller's catch runs after `end` listeners. `ResponseStream` already carries this contract, and #2047 is the existing cross-helper streaming precedent. I checked the red cases against the base, then ran `pnpm exec tsc`, lint, the three focused stream files with 391 tests passing, and the full suite with 6,999 handwritten and 556 generated tests passing. One otherwise-valid chunk with an extra top-level object-valued `error` key now fails with `APIError`; string error bodies remain unchanged, the check follows the function's existing property-detection style, and I can move the conversion into generic `Stream.fromReadableStream` if you prefer that broader central contract.

### DIFF
--- a/src/lib/EventStream.ts
+++ b/src/lib/EventStream.ts
@@ -1896,6 +1896,41 @@ export class EventStream<EventTypes extends BaseEvents> {
     return Boolean(this.#listeners[event]?.some((listener) => !listener.removed));
   }
 
+  #settleTerminalEvent<Event extends keyof EventTypes>(
+    event: Event,
+    args: EventParameters<EventTypes, Event>,
+    hasListeners: boolean,
+  ): void {
+    if (event === 'abort') {
+      const error = args[0] as APIUserAbortError;
+      if (!this.#catchingPromiseCreated && !hasListeners) {
+        Promise.reject(error);
+      }
+      this.#rejectConnectedPromise(error);
+      this.#rejectEndPromise(error);
+      this._emit('end');
+      return;
+    }
+
+    if (event === 'error') {
+      // NOTE: _emit('error', error) should only be called from #handleError().
+
+      const error = args[0] as OpenAIError;
+      if (!this.#catchingPromiseCreated && !hasListeners) {
+        // Trigger an unhandled rejection if the user hasn't registered any error handlers.
+        // If you are seeing stack traces here, make sure to handle errors via either:
+        // - runner.on('error', () => ...)
+        // - await runner.done()
+        // - await runner.finalChatCompletion()
+        // - etc.
+        Promise.reject(error);
+      }
+      this.#rejectConnectedPromise(error);
+      this.#rejectEndPromise(error);
+      this._emit('end');
+    }
+  }
+
   /** Dispatches a connection, failure, cancellation, or completion lifecycle event. */
   _emit<Event extends keyof BaseEvents>(event: Event, ...args: EventParameters<BaseEvents, Event>): void;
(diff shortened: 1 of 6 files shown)


src/lib/EventStream.ts
```diff
@@ -1896,6 +1896,41 @@ export class EventStream<EventTypes extends BaseEvents> {
...
+        // If you are seeing stack traces here, make sure to handle errors via either:
+        // - runner.on('error', () => ...)
+        // - await runner.done()
+        // - await runner.finalChatCompletion()
+        // - etc.
+        Promise.reject(error);
+      }
+      this.#rejectConnectedPromise(error);
+      this.#rejectEndPromise(error);
+      this._emit('end');
```
HAYDEN-OAI, line 1958: **[P1] Preserve terminal failures before emitting `end`.** When an `error` listener registered before an async iterator throws, listener dispatch stops before the iterator's internal `onFailure` listener runs. This newly guaranteed `end` emission then resolves a pending `next()` with `{ done: true }`; buffered iterators likewise drain and complete without surfacing the original error. The same issue exists for `abort` at line 1911 and affects Chat Completions, Assistants, and Responses. Please record the terminal error in iterator-visible state before emitting `end`, or guarantee internal failure listeners run even when user listeners throw.

HAYDEN-OAI (changes requested):
Requesting changes because a throwing terminal listener can silently convert a failed or aborted stream into successful async-iterator completion.

morgan-coded:
Recorded the terminal error as `#terminalFailure` in `#settleTerminalEvent` before `_emit('end')`, on both the `error` branch and the `abort` branch you named. `_createIterator` adopts it in `onEnd` and in a `next()` made after the stream ended, gated by each iterator's `rejectOnError`/`rejectOnAbort` so `events('error')`-style iterators never reject the events they yield, and `return()` marks the failure delivered so a consumer-returned iterator stays silently done.

Added three regression tests for your scenarios — a pending reader, a buffered drain-then-reject, and the abort variant — each failing on `d9b575ce` and passing on the revision. Checked the same probe, a throwing terminal listener registered before the iterator with one event delivered before the failure, against Chat Completions, Assistants, and Responses: `main` hangs, `d9b575ce` completes without surfacing the error, and the revision rejects with that first event still delivered.

One behavior change reaches past the throwing-listener case: an iterator or `toReadableStream()` created after the stream already failed now rejects once instead of completing as a silently truncated zero-chunk transfer, while streams that end without a failure, gated `error`/`abort` event iterators, and already-returned iterators are unchanged. Left two pre-existing behaviors alone: concurrently parked `next()` calls still deliver the failure only to the first, as on `d9b575ce`, and a throwing `end` listener still starves `onEnd`, though a fresh `next()` now surfaces the error where the prior head hung. `b7b8188d` sits on top of the reviewed head, so the re-review diff is only this change.

HAYDEN-OAI (approved):
Reviewed server-error conversion, listener-failure settlement, and pending/buffered iterators. The earlier terminal-error and inherited-error-property findings are addressed. No substantive introduced findings in the current revision.

PR #2504, opened 2026-08-27
feat(api): add default WebSocket User-Agent and audit log events

apcha-oai:
## Summary

Backend Realtime and Responses WebSocket connections now identify the SDK using `User-Agent: <ClientClass>/JS <SDK version>`. This covers the stable and beta Node `ws` transports. Caller-provided headers retain precedence, and browser-native WebSocket transports are unchanged.

The Admin Organization audit log request and response types also include `tenant.trusted_access.business_verification.started` and `tenant.trusted_access.application.submitted`.

## Validation

The Realtime regression failed for both stable and beta helpers before the fix and passed afterward. A separate handwritten Responses test mocks `ws` and uses fake timers to verify default User-Agent values and caller overrides on initial connections and reconnects for both stable and beta entrypoints; removing the header makes both default cases fail. All 423 focused Realtime, Responses, provider, and redirect tests passed, along with lint and TypeScript checking on Node 24.


jbeckwith-oai (approved):
Verified stable and beta Node Realtime/Responses WebSocket handshakes send the versioned SDK User-Agent while preserving caller overrides, reconnects, authentication, and browser transports. Both audit-log events are generated consistently in request and response unions, and the focused regressions cover the previously identified Responses gap.

PR #2506, opened 2026-08-27
fix(polling): treat a cancelled vector store file as terminal

dfedoryshchev:
`vectorStores.files.poll` never returns for a `cancelled` file, and re-requests it with no delay between calls.

`vector-store-polling.ts:23` lists `['failed', 'completed']` as terminal, but `VectorStoreFile.status` is `'in_progress' | 'completed' | 'cancelled' | 'failed'`. `pollWithResponse` sleeps only for intermediate statuses and returns only for terminal ones, so `cancelled` falls through both branches of the `while (true)` loop and the next request goes out immediately.

`pollVectorStoreFileBatch` in the same file already lists `cancelled` as terminal, and `openai-python`'s `poll_vector_store_file` returns on it too, so this is the file disagreeing with itself.

Reachability is narrow and worth stating: there is no per-file cancel endpoint, so a file reaches `cancelled` only via `fileBatches.cancel`, and the hang needs someone to then poll one of those files through the single-file helper.

Two things you would otherwise find yourself. `docs/uploads.md` said the helper "does not currently return for a `cancelled` file, so use an abort signal" - I have updated that paragraph, and I am flagging it because it means this changes documented behaviour rather than fixing an oversight. It also did not mention the tight request loop, which is the part I would want fixed.

The existing `preserves the distinct cancelled status behavior` test encoded the divergence directly (`name === 'file' ? 'completed' : 'cancelled'`), so it is replaced rather than extended; the new test asserts both helpers stop after one request and fails on the tenth re-request.

Happy to drop the doc edit or narrow this to the loop alone if you would rather keep the current return semantics.

### DIFF
--- a/docs/uploads.md
+++ b/docs/uploads.md
@@ -173,12 +173,10 @@ if (file.status !== 'completed') {
 }
 ```
 
-The helper returns files with `completed` or `failed` status; inspect `status`
-and `last_error` before assuming ingestion succeeded. It does not currently
-return for a `cancelled` file, so use an abort signal or retrieve the file
-directly when cancellation is possible. Vector-store polling uses the interval
-suggested by the API, or five seconds when none is provided; `pollIntervalMs`
-overrides that interval.
+The helper returns files with `completed`, `cancelled`, or `failed` status;
+inspect `status` and `last_error` before assuming ingestion succeeded.
+Vector-store polling uses the interval suggested by the API, or five seconds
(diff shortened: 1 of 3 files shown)


HAYDEN-OAI (approved):
Reviewed the polling helper, file/batch behavior, and public callers. Cancelled files return through the terminal-state path, and the documentation reflects the change. No substantive findings.

PR #2514, opened 2026-08-28
fix(azure): carry declared options through withOptions and fix deployment paths

morgan-coded:
`AzureOpenAI` overrides `withOptions` only to re-type it: `apiVersion` and `deployment` are instance fields that `OpenAI.withOptions` cannot see when it rebuilds the clone, and the inherited `baseURL` is always carried across. So a clone throws `The OPENAI_API_VERSION environment variable is missing or empty` when that variable is unset, silently rebuilds the deployment segment from `body.model` when it is set — a client configured for `deployment-a` starts sending `/openai/deployments/request-model/chat/completions`, which 404s — and rejects `withOptions({ endpoint })` with `baseURL and endpoint are mutually exclusive`, though `endpoint` is declared on the `Partial<AzureClientOptions>` the override accepts. Separately, the deployment-path suppression was a substring test on the base URL, so a host like `deployments.example.com` or a gateway path like `/azure/deployments-proxy/openai` dropped the deployment segment and Azure answered 404 with nothing to show the SDK had dropped it.

Following `src/bedrock.ts`, which overrides `withOptions` to re-inject a subclass-only field the base clone cannot see, the Azure override now re-injects `apiVersion` and `deployment` ahead of the caller's options and clears the inherited `baseURL` only when the caller passed an `endpoint` and no explicit `baseURL`, so caller-supplied values still win and `{ deployment: undefined }` still clears. The suppression now tests `new URL(baseURL).pathname` for a `deployments` path segment, so `/openai/deployments/existing` and its trailing-slash form still suppress and only the false matches change. One residual stays: with `OPENAI_BASE_URL` set, `withOptions({ endpoint })` still throws `baseURL and endpoint are mutually exclusive`, exactly as `new AzureOpenAI({ endpoint })` does in that environment.

I added 10 regression tests across `tests/lib/azure.test.ts` and `tests/lib/azure-deployment-path-safety.test.ts` — nine red before the fix, the tenth a parity pin on the boundary case — then ran lint, `tsc --noEmit`, and the full suite, with 7019 handwritten and 556 generated tests passing.

`AzureClientOptions` declares a `workloadIdentity` member that the Azure credential guard never consults, and I've deliberately left it untouched here: wiring it up would federate an Azure-only client's calls through OpenAI's own token exchange at the hardcoded `https://auth.openai.com/oauth/token`, so whether Azure should accept it or declare it `?: never` alongside `provider`/`dataResidency`/`credential`/`x509Transport` is your call, not mine — flagging it rather than changing it.

### DIFF
--- a/src/azure.ts
+++ b/src/azure.ts
@@ -3,7 +3,7 @@ import type { NullableHeaders } from './internal/headers';
 import { buildHeaders } from './internal/headers';
 import * as Errors from './error';
 import type { FinalRequestOptions } from './internal/request-options';
-import { isObj, readEnv } from './internal/utils';
+import { hasOwn, isObj, readEnv } from './internal/utils';
 import { path } from './internal/utils/path';
 import { OpenAI } from './client';
 import type { ClientOptions } from './client';
@@ -162,8 +162,9 @@ export class AzureOpenAI extends OpenAI {
     };
 
     // The inherited base URL is always carried into the clone, so an `endpoint` override would
-    // otherwise collide with it; let the endpoint rebuild the base URL instead.
-    if (options.endpoint !== undefined && options.baseURL === undefined) {
+    // otherwise collide with it; let the endpoint rebuild the base URL instead. Both tests read
+    // own properties, so an option bag that passed neither field keeps the inherited base URL.
+    if (hasOwn(options, 'endpoint') && options.endpoint !== undefined && !hasOwn(options, 'baseURL')) {
       azureOptions.baseURL = undefined;
     }
 
@@ -236,7 +237,13 @@ export class AzureOpenAI extends OpenAI {
  * such as `/deployments-proxy/`, or a host like `deployments.example.com`.
  */
 function hasDeploymentPathSegment(baseURL: string): boolean {
-  return new URL(baseURL).pathname.split('/').includes('deployments');
+  try {
+    return new URL(baseURL).pathname.split('/').includes('deployments');
+  } catch {
+    // A base URL that is not an absolute URL has no path segments to read. Report none, the way
+    // the previous substring test did, and let joining the request path reject it as it did before.
+    return false;
+  }
 }
 
 const _deployments_endpoints = new Set([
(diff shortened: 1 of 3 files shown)


morgan-coded:
Tightened the `withOptions` guard to test what the caller passed — `hasOwn(options, 'endpoint')` and `!hasOwn(options, 'baseURL')`, the `hasOwn` test `OpenAI.withOptions` already applies to `baseURL` at `src/client.ts:690` — rather than reading both fields through the prototype chain. The automated review flagged that at `src/azure.ts:167`. Also wrapped the parse in `hasDeploymentPathSegment`: a base URL like `https://` threw `TypeError: Invalid URL` out of `buildRequest` where the pre-PR substring test reported no match, so it reports no segment there instead. `144062a5` sits on top of the reviewed head, so the re-review diff is one commit.

The repro in that comment is accurate for `8679512b` — the clone does rebuild against the polluted origin with the key attached. Unmodified `main` never reaches that state through this path: it refuses the clone with `The OPENAI_API_VERSION environment variable is missing or empty` when that variable is unset and `baseURL and endpoint are mutually exclusive` when it is set, so nothing released is affected and the defect existed on this PR's head commit and nowhere else. What made the path reachable is this PR's own fix — the clone builds at all because `apiVersion` and `deployment` are now carried into it — and the constructor takes its options by destructuring (`src/azure.ts:86-96`), so an inherited `endpoint` is visible to it.

Two sentences in the description are stale now. Corrected:

- Tests: the branch adds 19 test blocks and 32 executed cases to the two azure test files, of which 16 blocks and 20 cases are red at the pre-PR parent; the remaining 3 blocks and 12 cases are parity pins that pass there by design. The full suite is 7041 handwritten and 556 generated passing.
- Deployment paths: on a base URL that parses, the answer moves in both directions — `deployments.example.com` and `/azure/deployments-proxy/openai` stop suppressing, and a base URL whose `/deployments` segment appears once WHATWG normalization runs (a backslash, an embedded tab, carriage return, or newline) starts suppressing where the raw substring test did not match, which now has its own test block. A base URL that does not parse always reports no segment, so the ones that textually carried `/deployments` insert where they used to suppress, and the request still ends at `TypeError: Invalid URL` once the path is joined, as it did before.

HAYDEN-OAI (approved):
Reviewed Azure clone defaults, caller overrides, deployment routing, and token-provider retention. The earlier inherited-endpoint finding is addressed. No substantive findings in the current revision.

PR #2516, opened 2026-08-30
fix: preserve package selection in parallel ecosystem tests

HAYDEN-OAI:
## Problem

The ecosystem CLI loses the value of `--fromNpm` when it starts parallel workers: it forwards only `--from-npm`. The parent deliberately skips building a local tarball when a package is selected, but each worker sees an empty selection and falls back to `.pack/openai.tgz`.

As a result, using `--fromNpm` with `--parallel` or `--jobs=2` fails when there is no local tarball, or can test a stale local tarball instead of the requested package.

## Fix

Forward the selected package as one `--from-npm=<value>` argument. This is a one-line runner change; no SDK runtime, generated code, dependencies, workflow permissions, or release policy change.

Added one table-driven integration regression covering:
- Sequential execution as an unchanged control.
- Both explicit worker counts and `--parallel`.
- Both `--fromNpm` and `--from-npm` spellings.
- A local package path containing spaces and `=`, with each real worker installing and executing the selected package.

The fixtures use local, synthetic packages and offline npm installation; no API credentials or live API requests are needed.

## Validation

Using Node.js 24.20.0 and the repository-pinned pnpm 11.5.1:
- Before the fix: sequential regression passed; both worker-mode regressions failed.
- `./scripts/test tests/ecosystem-cli.test.ts`: 23 passed.
- `pnpm install --frozen-lockfile`: passed.
- `pnpm exec tsc`: passed.
- `pnpm lint`: passed.
- `pnpm build`: passed.
- `./scripts/test`: 7,012 handwritten tests and 556 generated tests passed; 2 existing generated tests skipped.
- `git diff --check`: passed.

## Adversarial self-review

Reviewed argument boundaries, shell metacharacter handling, credential isolation, option aliases, sequential/parallel behavior, and test portability. Tightened the fixture to use the repository's pinned package manager, then repeated the review and full validation. The value remains a single argv element, and existing live/deploy gates and install/type-check credential filtering are unchanged.

Both changed files are absent from the pinned public generated snapshot (`65094c1637432c123c310d03e0cda1e7e6f1c997`). Open PRs were checked for overlapping work.

### DIFF
--- a/ecosystem-tests/cli.ts
+++ b/ecosystem-tests/cli.ts
@@ -844,7 +844,7 @@ async function main() {
                       ...(args.live ? ['--live'] : []),
                       ...(args.verbose ? ['--verbose'] : []),
                       ...(args.deploy ? ['--deploy'] : []),
-                      ...(args.fromNpm ? ['--from-npm'] : []),
+                      ...(args.fromNpm ? [`--from-npm=${args.fromNpm}`] : []),
                     ],
                     {
                       stdio: 'pipe',
--- a/tests/ecosystem-cli.test.ts
+++ b/tests/ecosystem-cli.test.ts
@@ -170,6 +170,73 @@ describe('ecosystem test CLI', () => {
   });
 
   test.each([
+    { name: 'sequential', options: [], packageOption: '--fromNpm' },
+    { name: 'explicit worker count', options: ['--jobs=2'], packageOption: '--fromNpm' },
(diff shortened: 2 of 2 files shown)


jbeckwith-oai (approved):
Reviewed the exact head. The package selection is preserved as a single argv element for parallel workers, including paths with spaces or `=`, and the focused regression coverage matches the requested scope. No blocking findings.

PR #2517, opened 2026-08-30
fix(types): distinguish parse-only Zod function tools

HAYDEN-OAI:
## Problem

`zodFunction()` currently marks every returned tool as runnable, even when its optional `function` callback is omitted:

```ts
const tool = zodFunction({
  name: 'lookup',
  parameters: z.object({ id: z.string() }),
});

client.chat.completions.runTools({
  model: 'gpt-4o',
  messages: [],
  tools: [tool],
});
```

This compiles on main, but both streaming and non-streaming runners reject before making a request because the tool has no associated function. The existing `standardFunction()` helper already distinguishes callback-free tools.

## Changes

- Give `zodFunction()` two overloads: a guaranteed callback produces a runnable tool; an omitted, undefined, or potentially undefined callback produces a parse-only tool.
- Preserve the existing argument/callback metadata and contextual argument inference, including calls with an explicit schema type argument.
- Remove the now-unnecessary `@ts-expect-error` from the implementation.
- Add compile-time regressions in the existing Zod test file for both `runTools()` modes, parse/stream acceptance, Zod v3/v4/Mini schemas, explicit generic calls, and promise-returning callbacks.

This is a typing-only correction in two handwritten files. The function body, generated schemas, parser behavior, callback execution, and emitted runtime JavaScript are unchanged. Both files are absent from the pinned generated snapshot; no generated or vendored parser code is modified, and the open Zod schema-conversion PRs are unaffected.

## Compatibility

Callback-bearing calls remain runnable. Callback-free tools remain valid for `parse()` and `stream()`; they now fail type-checking when passed to `runTools()`, matching its existing runtime rejection. If a callback's type includes `undefined`, narrow it before creating a tool intended for execution.

## Validation

On Node.js 24.20.0 with pinned pnpm 11.5.1:

- Final compile-time regressions against main: six expected diagnostics, including unused rejection assertions because invalid `runTools()` calls were accepted.
- `pnpm exec tsc`: passes with the fix.
- Focused Zod, Standard Schema, and tool-runner tests: 165 passed.
- `pnpm lint` and `pnpm build`: passed.
- Full `CI=true ./scripts/test`: 7,009 handwritten and 556 generated tests passed; two existing generated tests skipped.
- Isolated emitted-declaration fixtures with Zod v3: both CJS and ESM pass TypeScript 4.9.5 and 6.0.3, including negative `runTools()` assertions.
- Published source navigation check with TypeScript 4.9.5: passed.
- Comment-stripped CommonJS and ESM output compared with main: identical.
- `git diff --check`: passed.

## Adversarial self-review

Reviewed callback presence, explicit `undefined`, callback unions, contextual and explicit-generic inference, async callbacks, both streaming modes, declaration compatibility, and runtime/security effects. Tightened the callback metadata and compatibility controls, then repeated review and validation. No remaining findings; no new runtime logic or dependencies.

### DIFF
--- a/src/helpers/zod.ts
+++ b/src/helpers/zod.ts
@@ -333,18 +333,8 @@ export function zodTextFormat<ZodInput extends ZodTypeLike>(
   );
 }
 
-/**
- * Creates a chat completion `function` tool that can be invoked
- * automatically by the chat completion `.runTools()` method or automatically
- * parsed by `.parse()` / `.stream()`.
- *
- * Arguments are converted to strict JSON Schema and validated with the supplied
- * Zod schema before the optional callback receives them.
- *
- * @param options Model-visible function name, Zod parameter schema, description,
- * and optional callback used by `chat.completions.runTools()`.
- */
-export function zodFunction<Parameters extends ZodTypeLike>(options: {
+/** Model-facing settings and an optional execution callback for a Zod function tool. */
+interface ZodFunctionOptions<Parameters extends ZodTypeLike> {
   /** Model-visible function name used to identify matching tool calls. */
   name: string;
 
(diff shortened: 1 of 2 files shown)


jbeckwith-oai (approved):
Reviewed the exact head. The overloads correctly distinguish guaranteed callbacks from parse-only or optional callbacks while retaining inferred Zod argument types. No blocking findings.

PR #2518, opened 2026-08-30
fix(polling): honor zero custom polling intervals

HAYDEN-OAI:
## Summary

Honor an explicitly configured `pollIntervalMs: 0` in the shared handwritten polling helper.

The helper already sends `X-Stainless-Custom-Poll-Interval: 0`, but its truthiness check discards zero when selecting the local delay. As a result, assistant-run, vector-store-file, and vector-store-file-batch polling wait for the server's interval or the five-second default despite the explicit override.

The fix reads the interval once per iteration and accepts zero alongside the previously accepted values. It preserves existing `NaN` fallback, negative intervals, header parsing/precedence, caller options, status handling, and cancellation behavior. Zero still uses the existing asynchronous sleep path.

## Scope and regression coverage

- Only four handwritten files change; no generated/upstream code, dependencies, exports, or release policy changes.
- Correct the existing zero-interval expectations and cover both server and default interval overrides through all three public poll methods.
- Add fake-timer public-entrypoint regressions with a real `AbortSignal`, asserting completion without the fallback wait and no retained listeners or timers.
- This does not change cancelled-file status handling, which is addressed separately by #2506.

## Validation

- Before the fix: **12 regression failures**, with the remaining 79 focused tests passing.
- After the fix: **91 focused polling/cancellation tests passed**.
- Full `CI=true ./scripts/test`: **7,018 handwritten tests and 556 generated tests passed**; 2 existing generated tests skipped.
- Built CommonJS and ESM packages: all **12 public polling smoke cases passed on Node 22.0.0**, covering each method with and without a caller signal.
- Completed two adversarial self-review passes covering security boundaries, compatibility, numeric edge cases, option/header preservation, timer/listener cleanup, and regression completeness. Clarified zero handling in the helper JSDoc; no further changes needed.

### DIFF
--- a/src/lib/polling.ts
+++ b/src/lib/polling.ts
@@ -92,8 +92,8 @@ function sleepUntilAborted(milliseconds: number, signal: AbortSignal): Promise<v
 
 /**
  * Repeatedly retrieves a lifecycle resource using the existing polling-helper
- * headers. Intermediate states wait for the explicit interval, server interval,
- * or five-second default; terminal states return the same parsed object.
+ * headers. Intermediate states wait for the explicit interval (including zero),
+ * the server interval, or five-second default; terminal states return the same parsed object.
  * The caller's signal interrupts intermediate waits, unknown states retry
  * immediately, and retrieval errors propagate unchanged.
  *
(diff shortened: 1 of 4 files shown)


jbeckwith-oai (approved):
Reviewed the exact head. Explicit zero polling intervals now override server/default values without changing the established handling of other interval values, and cancellation/listener cleanup remains covered. No blocking findings.

PR #2519, opened 2026-08-30
fix(uploads): reject zero batch upload concurrency

HAYDEN-OAI:
## Summary

Reject zero `maxConcurrency` in `vectorStores.fileBatches.uploadAndPoll` before any upload or batch-creation request.

Previously, zero created no workers, so every supplied new file was silently skipped. If the caller also supplied existing `fileIds`, the helper could create and return a successful batch containing only those existing files. Without existing IDs, it submitted an empty batch instead.

The fix is a three-line guard in the handwritten upload orchestrator. It rejects with `RangeError: maxConcurrency must be greater than 0`, including for negative zero. Valid upload concurrency, default/null behavior, file-count capping, completion-order IDs, caller options, and the existing native errors for other invalid numeric limits are unchanged.

## Scope and regression coverage

- Three handwritten files only: the guard, focused regression tests, and upload documentation. No generated/upstream code, dependencies, or exported signatures change.
- Replace the old zero-concurrency compatibility expectation, which encoded the skipped-upload behavior, with public SDK regressions using the actual fetch boundary.
- Cover zero and negative zero with existing IDs, plus zero when IDs are omitted. All cases reject without any request or identifier mutation.
- Existing positive concurrency, upload ordering, error propagation, and cancellation coverage remains intact.

## Validation

- Before the fix: both initial public-entrypoint regressions failed because the promise resolved to a completed batch.
- After the fix and coverage tightening: **73 focused upload/cancellation tests passed**.
- Full `CI=true ./scripts/test`: **7,011 handwritten tests and 556 generated tests passed**; 2 existing generated tests skipped.
- Built CommonJS and ESM packages: **8 public upload smoke cases passed on Node 22.0.0**, checking zero rejection and successful uploads with concurrency 1 and Infinity.
- Completed two adversarial self-review passes for security, compatibility, numeric edge cases, async failure behavior, mutation/side effects, and missing regressions. Tightened the omitted-ID case and verified its exact optional-property typing before publication.

### DIFF
--- a/docs/uploads.md
+++ b/docs/uploads.md
@@ -203,6 +203,8 @@ if (batch.status !== 'completed' || batch.file_counts.failed > 0 || batch.file_c
 }
 ```
 
+A zero `maxConcurrency` is rejected before any upload or batch-creation request.
+
 Use `vectorStores.files.createAndPoll` or
 `vectorStores.fileBatches.createAndPoll` when files have already been uploaded
 and you only need to attach and process their IDs. These polling helpers accept
--- a/src/lib/vector-store-upload.ts
+++ b/src/lib/vector-store-upload.ts
@@ -8,7 +8,8 @@ type UploadOptions = RequestOptions & { pollIntervalMs?: number; maxConcurrency?
 
 /**
  * Uploads files with a shared iterator, then creates and polls the batch. Every
- * worker settles before upload failures are propagated.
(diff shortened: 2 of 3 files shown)


jbeckwith-oai (approved):
Reviewed the exact head. Zero concurrency is rejected before network activity, including negative zero, while existing invalid-concurrency behavior and file ID immutability remain intact. No blocking findings.

PR #2520, opened 2026-08-30
fix(runners): suppress content events for input history

HAYDEN-OAI:
## Summary

Make non-streaming `runTools()` honor the existing `emit = false` flag while loading input history.

`AbstractChatCompletionRunner` deliberately adds initial messages without emitting events. The `ChatCompletionRunner` override forwarded that flag to the base class but ignored it for `content` events. A conversation with an earlier assistant reply therefore emitted both the old reply and the new reply as fresh content, even though `message` events correctly contained only the new response.

Structured assistant history made this worse: a content-part array could reach a listener whose public parameter type is `string`.

The runtime fix is one condition: emit content only when `emit` is true. Input history remains in the request and runner state, and newly received assistant replies retain their existing events. This matches the documented content-event contract.

## Scope and regressions

- Two handwritten files only; no generated/upstream code, dependencies, public signatures, or streaming implementation changes.
- Add public-entrypoint regressions for text history in default and explicit `stream: false` modes, plus structured content-part history.
- Check unchanged request history, retained message identity, normal new-response events, and final-result/lifecycle behavior through the existing runner assertions.
- The existing streaming-error PR #2502 addresses different files and behavior.

## Validation

- Built-package reproduction before the fix: `content` received `["Earlier reply", "New reply"]` while `message` received only `["New reply"]`.
- Both initial regressions failed before the fix, including the array-valued history case.
- After coverage tightening: **79 focused runner/stream tests passed**.
- Full `CI=true ./scripts/test`: **7,012 handwritten tests and 556 generated tests passed**; 2 existing generated tests skipped.
- **8 built CommonJS/ESM public runner cases passed on Node 22.0.0**, covering both non-streaming forms and both history representations.
- Completed two adversarial self-review passes for security, compatibility, event ordering, callback argument types, history identity/mutation, and missing regression coverage. Added the explicit non-streaming case and clarified the event JSDoc; no further changes needed.

### DIFF
--- a/src/lib/ChatCompletionRunner.ts
+++ b/src/lib/ChatCompletionRunner.ts
@@ -11,7 +11,7 @@ import type { AutoParseableTool } from '../lib/parser';
 
 /** Events emitted while executing a non-streaming chat completion tool run. */
 export interface ChatCompletionRunnerEvents extends AbstractChatCompletionRunnerEvents {
-  /** Called when an assistant message with nonempty text content is received. */
+  /** Called when a new assistant message with nonempty text is received; input history is not emitted. */
   content: (content: string) => void;
 }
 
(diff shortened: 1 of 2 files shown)


jbeckwith-oai (approved):
Reviewed the exact head. Content events now respect the existing emit flag, so input history is retained without being replayed while newly received assistant content still emits. No blocking findings.

PR #2521, opened 2026-08-30
fix(types): include missing parsed response stream events

HAYDEN-OAI:
## Summary

Derive `ParsedResponseStreamEvent` from the canonical `ResponseStreamEvent` union instead of maintaining a second event list.

The handwritten union had drifted to 36 variants while the API union has 58. It excluded valid reasoning, image-generation, MCP, queued-response, annotation, and custom-tool events. Consumers importing the parsed-event type could not assign those events or narrow them by their valid discriminators.

The replacement excludes the two raw deltas that receive SDK snapshots and substitutes their existing enriched types. All other event shapes come directly from the canonical API union, so future additions cannot silently drift out of this declaration.

## Scope and compatibility

- Two handwritten files only, with **41 fewer lines overall**. No generated/upstream code or dependencies change.
- This corrects the public type union; it does not change which events the runtime emits.
- Existing event shapes remain intact, and text/function-argument deltas still require a string `snapshot`.
- Consumers with exhaustive switches may now be prompted to handle valid event variants the old declaration omitted.

## Regression tests and adversarial review

- Added compile-time parity checks for every canonical event discriminator and every non-enriched event shape.
- Added exact enriched-delta checks, tightened during self-review with explicit required-string snapshot assertions.
- Completed two adversarial passes for security, declaration compatibility, event-shape preservation, snapshot requirements, module formats, and future schema additions.

## Validation

- The new parity assertions failed against the old source declaration.
- Isolated `.cts` and `.mts` consumers failed against the old published declarations, including image-generation narrowing, then passed after rebuilding.
- **149 focused response-stream/accumulator tests passed**.
- Full `CI=true ./scripts/test`: **7,011 handwritten tests and 556 generated tests passed**; 2 existing generated tests skipped.
- Published `.d.ts` and `.d.mts` consumers passed on **TypeScript 4.9.5 and 6.0.3**, including negative checks for missing snapshots.
- Published source navigation typechecked on TypeScript 4.9.5.
- CommonJS and ESM JavaScript emission is **byte-identical to the base**.

### DIFF
--- a/src/lib/responses/EventTypes.ts
+++ b/src/lib/responses/EventTypes.ts
@@ -1,40 +1,7 @@
 import type {
-  ResponseAudioDeltaEvent,
-  ResponseAudioDoneEvent,
-  ResponseAudioTranscriptDeltaEvent,
-  ResponseAudioTranscriptDoneEvent,
-  ResponseCodeInterpreterCallCodeDeltaEvent,
-  ResponseCodeInterpreterCallCodeDoneEvent,
-  ResponseCodeInterpreterCallCompletedEvent,
-  ResponseCodeInterpreterCallInProgressEvent,
-  ResponseCodeInterpreterCallInterpretingEvent,
-  ResponseCompletedEvent,
-  ResponseContentPartAddedEvent,
-  ResponseContentPartDoneEvent,
-  ResponseCreatedEvent,
-  ResponseErrorEvent,
-  ResponseFailedEvent,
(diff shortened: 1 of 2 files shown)


jbeckwith-oai (approved):
Reviewed the exact head. The parsed stream union now tracks every generated response event while replacing only the two enhanced delta variants with snapshot-bearing types. No blocking findings.

PR #2526, opened 2026-08-31
fix(types): preserve custom tool factory metadata

HAYDEN-OAI:
## Summary

- Pass the complete options metadata to `AutoParseableTool` and `AutoParseableResponseTool` instead of passing the parsed argument value type a second time.
- Preserve explicit parser-output, callback-argument, and function-name types in both public custom-tool factories.
- Derive Chat Completions runnable metadata from the supplied callback: a guaranteed callback is runnable, `undefined` is parse-only, and an optional callback remains unguaranteed. Callback parameters are checked against the parser's declared output.

## Reproduction

A custom factory call with metadata such as `{ name: 'lookup'; arguments: { city: string }; function: (args: { city: string }) => string }` currently returns a tool whose `$parseRaw()` result and name metadata are `unknown`. The argument object is incorrectly treated as another metadata object. A valid callback-backed Chat Completions tool is also incorrectly marked non-runnable and rejected by both public `runTools()` overloads.

A strict public-import consumer reproduced these errors before the change. The new compile-time regressions fail on the base and pass with the fix. They also ensure correcting the metadata cannot accidentally make a callback-free or optional-callback tool runnable, even if the declared metadata contains a function or the parsed argument object has its own `name`, `arguments`, and `function` properties.

## Validation

- Complete `CI=true ./scripts/test`: 7,012 handwritten tests and 556 generated tests pass (2 existing generated skips), with all 12 generated snapshots passing.
- Focused factory/parser suites: 46 tests pass.
- `pnpm exec tsc --noEmit`, `pnpm lint`, and `pnpm build` pass. CI's existing type-check step includes the new compile-time assertions.
- Strict emitted-declaration consumer passes TypeScript 4.5.5 (CommonJS), 4.9 (CommonJS/ESM), and 6 (CommonJS/ESM), all with `skipLibCheck: false`. TypeScript 4.9 also checks distributed source.
- Built CJS/ESM factory checks pass on Node 22.0.0, 24.20.0, and 26.7.0: 12 cases per runtime covering missing/sync/async callbacks, parser and callback identity, frozen definitions, non-enumerable metadata, and unchanged JSON serialization.
- Compared both factories' emitted CJS/ESM code with main using the pinned compilers to normalize comments/formatting; runtime code is unchanged.

Reviewed the complete return-type invariant in both factories, not just the initially failing property access. Verified callback presence is determined by the actual callback argument rather than unrelated parser-output properties or a possibly inaccurate metadata declaration. Covered parse-only and optional tools in both runner modes, sync/async callbacks, metadata without a function field, string/array/nested parser outputs, and incompatible callback argument types. Rechecked the final diff after validation.

This is a types-only change in handwritten code. No runtime parser, schema conversion, request serialization, generated file, dependency, custom-code budget, or runtime policy changes. All three paths are absent from the pinned generated snapshot. Open-PR checks found no overlap; #2517 changes the Zod convenience helper, not these custom factories.

### DIFF
--- a/src/lib/ResponsesParser.ts
+++ b/src/lib/ResponsesParser.ts
@@ -220,9 +220,9 @@ export function makeParseableResponseTool<OptionsT extends ToolOptions>(
     /** Converts the raw JSON argument string into the function's typed argument value. */
     parser: (content: string) => OptionsT['arguments'];
     /** Optional callback available to helpers that execute the parsed function. */
-    callback: ((args: any) => any) | undefined;
+    callback: ((args: OptionsT['arguments']) => any) | undefined;
   },
-): AutoParseableResponseTool<OptionsT['arguments']> {
+): AutoParseableResponseTool<OptionsT> {
   const obj = { ...tool };
 
   Object.defineProperties(obj, {
@@ -240,7 +240,7 @@ export function makeParseableResponseTool<OptionsT extends ToolOptions>(
     },
   });
 
-  return obj as AutoParseableResponseTool<OptionsT['arguments']>;
+  return obj as AutoParseableResponseTool<OptionsT>;
 }
 
 /** Returns whether a Responses API tool carries the SDK's argument-parser marker. */
--- a/src/lib/parser.ts
+++ b/src/lib/parser.ts
@@ -214,19 +214,43 @@ export type AutoParseableTool<
   $parseRaw(args: string): OptionsT['arguments'];
 };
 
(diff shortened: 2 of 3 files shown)


jbeckwith-oai (approved):
Reviewed the exact head. The factory overloads correctly preserve the full metadata type, derive runnable status from the supplied callback, and keep callback argument types aligned with parser output. Focused type tests and `tsc --noEmit` pass.

PR #2527, opened 2026-08-31
fix(types): account for optional event arguments

HAYDEN-OAI:
## Summary

- Correct `.emitted()` result types in the core `EventEmitter`, library `EventEmitter`, and `EventStream` for optional and rest parameters.
- Share one small type-only mapping instead of maintaining three copies of the return-type conditional.
- Preserve fixed zero-, one-, and multi-argument results, required rest prefixes/suffixes, and the existing `unknown` fallback for union event names. No runtime code changes.

## Reproduction

With a public emitter event declared as `(value?: { text: string }) => void`, `.emitted('value')` currently has type `Promise<[value?: { text: string }]>`. Emitting one object actually resolves with that object, and emitting no arguments resolves with `undefined`. Valid access to the returned value is rejected, while the declared tuple shape does not match the runtime.

A strict consumer importing the built public core emitter, library emitter, and event stream reproduced this compiler failure on all three classes. The committed regression assertions also fail on the base revision. Optional second parameters and rest-only events have the same mismatch when a call contains fewer than two arguments.

The correction includes the applicable unwrapped-value and empty-call cases while retaining each declared tuple/array branch for multi-argument calls. It does not change event dispatch, promise settlement, listener cleanup, buffering, or serialization.

## Validation

- Complete final `CI=true ./scripts/test`: **7,036 handwritten tests and 556 generated tests passed**, with 2 existing generated skips and all 12 snapshots passing.
- The new shared matrix covers **27 public-emitter cases** across all three implementations. Existing event/listener and waiter-performance suites pass as well.
- `pnpm exec tsc --noEmit`, `pnpm lint`, and `pnpm build` pass.
- Strict built-declaration consumers pass **TypeScript 4.5.5** (CommonJS), **4.9.5** (CommonJS/ESM), and **6.0.3** (CommonJS/ESM), with `skipLibCheck: false` and `exactOptionalPropertyTypes: true`. TypeScript 4.9.5 also checks distributed source.
- **180 built-package runtime checks pass** across Node **22.0.0**, **24.20.0**, and **26.7.0**, covering all three public classes in CJS and ESM, zero/one/multiple arguments, explicit `undefined`, nulls, array-valued single arguments, and object identity.
- Comparing base and final TypeScript emission with comments removed produces identical runtime code for all three source files in both module formats. Type-only imports add no runtime dependency.

Reviewed optional arity, explicit `undefined`, fixed tuples, homogeneous rest arrays, required prefixes/suffixes, compiler-floor behavior, and module isolation. Compatibility review caught an overly narrow `never` result for union event names in the first formulation; the final mapping preserves the prior `unknown` fallback and has a regression for it. Repeated the type checks, full tests, and runtime comparisons after that correction.

Only handwritten source and one regression file change. No generated/upstream files, parser algorithms, credentials, dependencies, budgets, runtime policy, or listener behavior change. Checked all open PRs: #2502 also touches `EventStream.ts`, but only changes independent terminal-failure handling, not `.emitted()` result typing.

### DIFF
--- a/src/core/EventEmitter.ts
+++ b/src/core/EventEmitter.ts
@@ -1,3 +1,5 @@
+import type { EmittedEventResult } from '../lib/EventEmitter';
+
 /** Listener callback associated with one event name in a typed event map. */
 type EventListener<Events, EventType extends keyof Events> = Events[EventType];
 
@@ -145,13 +147,7 @@ export class EventEmitter<EventTypes extends Record<string, (...args: any) => an
    */
   emitted<Event extends keyof EventTypes>(
     event: Event,
-  ): Promise<
-    EventParameters<EventTypes, Event> extends [infer Param]
-      ? Param
-      : EventParameters<EventTypes, Event> extends []
-        ? void
-        : EventParameters<EventTypes, Event>
-  > {
+  ): Promise<EmittedEventResult<EventParameters<EventTypes, Event>>> {
     return new Promise((resolve, reject) => {
       const listeners = {
         onError: (error: unknown) => {
--- a/src/lib/EventEmitter.ts
+++ b/src/lib/EventEmitter.ts
@@ -14,6 +14,20 @@ export type EventParameters<Events, EventType extends keyof Events> = Record<
   EventListener<Events, EventType> extends (...args: infer P) => any ? P : never
 >[EventType];
 
+/** Result of awaiting an event: no value, one unwrapped value, or its argument tuple. */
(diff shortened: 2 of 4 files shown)


jbeckwith-oai (approved):
Reviewed the exact head. The shared emitted-result mapping matches the runtime’s zero/single/multiple-argument behavior across fixed, optional, and rest parameter forms while preserving the union-event fallback. The 27-case focused suite and `tsc --noEmit` pass.

PR #2528, opened 2026-08-31
fix(examples): keep tracking fine-tuning file validation

HAYDEN-OAI:
## Summary

Include `validating_files` in the fine-tuning example's existing polling condition. It is a nonterminal status in the public `FineTuningJob` contract, alongside `queued` and `running`.

Previously, a create response in `validating_files` caused the example to exit successfully without retrieving the job again or fetching any progress events. The example now keeps tracking until the job succeeds, fails, or is cancelled.

The production change is one condition in `examples/fine-tuning/fine-tuning.ts`. Request parameters, model selection, file handling, event reporting, polling interval, and terminal-state behavior are unchanged. No SDK runtime or generated code changes.

## Reproduction and regressions

Executed the actual example with the public SDK and an injected synthetic fetch transport. The real training-file stream was consumed as multipart data, while every API request was handled offline; no live fine-tuning job was created.

- Before: an initial `validating_files` response produced **zero job polls and zero event requests**, with no error.
- After: the same reproduction followed `queued` → `running` → `succeeded`, making **three job polls and three event requests**.
- Eight regressions cover repeated validation, queueing/training, all three terminal statuses, initially queued/running jobs, and failure/cancellation during validation. The three validation cases failed on the base; the five existing-behavior controls passed. All eight pass with the change.
- The tests run the transpiled example itself rather than a copied polling loop, preserve its timer callback behavior without real five-second waits, and assert that only one fine-tuning job is created.

## Validation

- `pnpm exec tsc --noEmit` and `pnpm lint` pass.
- Complete `CI=true ./scripts/test`: **7,017 handwritten tests and 556 generated tests pass**, with 2 existing generated skips and all 12 snapshots passing.
- `pnpm build` passes for both module formats.
- The actual-example reproduction passes on **Node 22.0.0, 24.20.0, and 26.7.0** using the built public SDK and the offline fetch transport.

Checked the complete six-status public contract: validation, queueing, and running continue; succeeded, failed, and cancelled stop. Verified repeated validation does not terminate early, terminal transitions do not issue another job poll, the existing five-second interval is retained, and event polling and single-job creation stay unchanged. The regression transport cannot make live API requests.

Both changed files are handwritten and absent from the pinned generated snapshot. No open PR overlaps this example. No new polling helper, state machine, dependency, timeout policy, API surface, or credential handling is introduced.

### DIFF
--- a/examples/fine-tuning/fine-tuning.ts
+++ b/examples/fine-tuning/fine-tuning.ts
@@ -49,7 +49,11 @@ async function main() {
 
   const events: Record<string, FineTuningJobEvent> = {};
 
-  while (fineTune.status === 'running' || fineTune.status === 'queued') {
+  while (
+    fineTune.status === 'validating_files' ||
+    fineTune.status === 'queued' ||
+    fineTune.status === 'running'
+  ) {
     fineTune = await client.fineTuning.jobs.retrieve(fineTune.id);
     console.log(`${fineTune.status}`);
 
--- a/tests/lib/fine-tuning-example.test.ts
+++ b/tests/lib/fine-tuning-example.test.ts
@@ -1,0 +1,146 @@
+import fs from 'node:fs';
+import { runInNewContext } from 'node:vm';
+import ts from 'typescript';
+import { vi } from 'vitest';
+import OpenAI from 'openai';
+import type { Fetch } from 'openai/internal/builtin-types';
+import type { FileObject } from 'openai/resources/files';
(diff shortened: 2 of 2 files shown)


jbeckwith-oai (approved):
Reviewed the exact head. Adding `validating_files` to the existing nonterminal polling condition covers the missing public status without changing terminal behavior or the example’s polling/event flow. The 8-case example regression suite and `tsc --noEmit` pass.

PR #2529, opened 2026-08-31
fix(uploads): accept cross-realm ArrayBuffers

HAYDEN-OAI:
## Summary

- Fix `toFile()` and streaming multipart uploads rejecting native `ArrayBuffer` values created in another JavaScript realm (for example, an iframe or Node VM context).
- Share a small native-brand check between the buffered and streaming paths. The captured `ArrayBuffer.prototype.byteLength` getter recognizes real buffers across realms without trusting `Symbol.toStringTag` or accepting `SharedArrayBuffer` as an `ArrayBuffer`.
- Add 27 public-boundary regression/control cases and an offline browser test using an actual iframe. No generated files, dependencies, upload metadata rules, or public API signatures change.

## Reproduction and behavior

Before this change, a VM-created `new Uint8Array([0, 1, 127, 255]).buffer` is accepted by the native `File` constructor but rejected by `toFile()` with `Unexpected data type: object; constructor: ArrayBuffer`. Supplying the same buffer as a streamed file chunk makes `files.create()` fail with an `Invalid streaming file chunk` cause.

The two upload paths used `instanceof ArrayBuffer`, which only recognizes the current realm's prototype. Ten new foreign-buffer cases fail on the base revision while the 17 existing-behavior controls pass; all 27 pass with the fix.

Coverage includes nonempty and empty foreign buffers, native buffers, foreign typed-array and DataView slices, direct and async-chunked `toFile()`, `toStreamingFile()`, raw async iterables, and readable streams. Streaming tests inspect the actual multipart bytes at a synthetic `fetch` boundary. Invalid string-tag lookalikes and `SharedArrayBuffer` inputs remain rejected.

## Validation

- `CI=true ./scripts/test`: 7,036 handwritten tests and 556 generated tests passed (two existing generated skips; 12 snapshots passed).
- Four focused upload suites: 134 tests passed.
- `pnpm lint`, `pnpm exec tsc --noEmit`, and `pnpm build` passed.
- Existing packed-package `browser-direct-import` workflow ran offline: fixture type-check, native browser ESM imports, API-key protection, and iframe buffer regression all passed. Re-ran its type-check and browser tests after the final lint-only fixture edit.
- Built CJS and ESM public entrypoints: 27 upload cases per format passed on Node 22.0.0, 24.20.0, and 26.7.0 (162 checks total).
- Published upload declarations passed strict TypeScript 4.5.5 CJS and 4.9.5 CJS/ESM consumers; distributed source passed TypeScript 4.9.5.

Adversarial review covered native-brand spoofing, empty buffers, byte-range preservation, metadata compatibility, module/runtime compatibility, and the unchanged browser credential guard. Tests are offline and use synthetic credentials. All four changed paths are absent from the pinned generated snapshot; current open PRs have no overlapping upload fix.

### DIFF
--- a/ecosystem-tests/browser-direct-import/public/index.js
+++ b/ecosystem-tests/browser-direct-import/public/index.js
@@ -22,7 +22,13 @@ async function runTests() {
     pre.textContent = JSON.stringify(results, null, 2);
   }
   for (const { path, run, timeout } of tests) {
-    if (!live && path[0] !== 'browser API-key protection remains enabled by default') {
+    if (
+      !live &&
+      ![
+        'browser API-key protection remains enabled by default',
+        'toFile accepts ArrayBuffers from another realm',
+      ].includes(path[0])
+    ) {
       continue;
     }
     console.log('running', ...path);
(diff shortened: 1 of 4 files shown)


jbeckwith-oai (approved):
Reviewed the exact head. The captured native `ArrayBuffer.prototype.byteLength` getter is a sound cross-realm brand check, rejects `SharedArrayBuffer` and tag lookalikes, and is applied consistently to buffered and streaming uploads. The 27-case focused suite and `tsc --noEmit` pass.

PR #2530, opened 2026-08-31
fix(examples): cancel Edge streams when requests abort

HAYDEN-OAI:
## Summary

- Pass the incoming `Request.signal` to the SDK stream in the Next.js Edge streaming example.
- Stop upstream work when an authenticated caller aborts, including before response headers arrive and after streaming starts.
- Keep the existing Pages Router handler, authorization/origin checks, prompt limits, request parameters, and response encoding unchanged. No SDK source, public types, dependencies, or generated files change.

## Reproduction

The example previously created its SDK stream without the incoming abort signal. Running the actual handler with the public SDK and an offline synthetic transport showed `request.signal.aborted === true` while the upstream signal remained un-aborted and the upstream body was not canceled.

The fix delegates cancellation to the SDK's existing signal handling rather than adding new listeners or lifecycle logic to the example. The same reproduction now aborts the upstream signal, cancels its body exactly once, releases the reader, and surfaces `APIUserAbortError` to the downstream reader.

## Regression coverage

Four tests exercise the actual example handler with the public SDK, standard Web `Request`/`Response` objects, and synthetic `fetch` responses:

- Already-aborted requests never reach the upstream transport.
- Cancellation propagates while upstream response headers are pending, without retries.
- Cancellation after the first SSE chunk closes the upstream body once and releases its reader and incoming abort listener.
- Normal completion preserves request parameters and response bytes; a later caller abort has no effect.

The three cancellation cases fail on unchanged main; the normal-completion control passes. Existing authorization tests now also assert that the exact incoming signal is forwarded.

## Validation

- `CI=true ./scripts/test`: 7,013 handwritten tests and 556 generated tests passed (two existing generated skips; 12 snapshots passed).
- Three focused example security/cancellation suites: 61 tests passed.
- `pnpm lint`, `pnpm exec tsc --noEmit`, and `pnpm build` passed.
- Standalone actual-handler reproduction passed with built CJS and ESM SDK entrypoints on Node 22.0.0, 24.20.0, and 26.7.0.

Adversarial review covered pre-aborted requests, cancellation before/after headers, custom abort reasons, reader/listener cleanup, retry behavior, normal completion, and the unchanged authentication boundary. All testing was offline; this does not claim a deployed Next.js end-to-end test. The three changed paths are absent from the pinned generated snapshot, and current open PRs have no overlapping fix.

### DIFF
--- a/examples/chat-completions/stream-to-client-next.ts
+++ b/examples/chat-completions/stream-to-client-next.ts
@@ -124,11 +124,14 @@ export default async function handler(request: Request): Promise<Response> {
   }
 
   const openai = new OpenAI();
-  const stream = openai.chat.completions.stream({
-    model: 'gpt-3.5-turbo',
-    stream: true,
-    messages: [{ role: 'user', content: prompt }],
-  });
+  const stream = openai.chat.completions.stream(
+    {
+      model: 'gpt-3.5-turbo',
+      stream: true,
+      messages: [{ role: 'user', content: prompt }],
+    },
+    { signal: request.signal },
+  );
 
   return new Response(stream.toReadableStream());
 }
--- a/tests/lib/next-streaming-example-cancellation.test.ts
+++ b/tests/lib/next-streaming-example-cancellation.test.ts
@@ -1,0 +1,172 @@
+import { getEventListeners } from 'node:events';
+import { afterEach, beforeEach, describe, expect, test, vi } from 'vitest';
+
+import { APIUserAbortError } from 'openai';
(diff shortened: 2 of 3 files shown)


jbeckwith-oai (approved):
Reviewed the exact head. Forwarding the incoming `Request.signal` delegates cancellation to the SDK’s existing lifecycle handling and preserves the example’s authentication, prompt, request, and response behavior. The focused cancellation/security suites and `tsc --noEmit` pass.

PR #2533, opened 2026-08-31
fix(examples): close multi-agent WebSockets on terminal responses

HAYDEN-OAI:
## Summary

Make the multi-agent Responses WebSocket example finish on the coordinator's terminal response and close its connection on every exit path.

The existing loop closes only for `response.completed`. A root `response.failed` or `response.incomplete` leaves it waiting indefinitely on the still-open socket. API errors and exceptions while writing output leave the iterator without closing the connection; the public WebSocket iterator explicitly does not own socket closure. A child agent's completion also currently closes the example before the coordinator can finish.

This change:

- Wraps the request and iteration in `try/finally`, with one `ws.close()` cleanup site.
- Rejects failed or incomplete coordinator responses with a fixed diagnostic containing only the terminal event type.
- Ignores child-agent terminal events and continues displaying the coordinator's response. Missing/null agent metadata retains the existing `/root` convention.
- Preserves the request, beta header, model, streamed text, and agent/coordinator formatting.

Only the handwritten example and its regression tests change. No generated SDK files, public declarations, dependencies, policies, or custom-code budget changes are involved. Neither file appears in the verified generated snapshot, and no other open PR covers this example.

## Regression coverage

The tests execute the actual example with the public `OpenAI` and beta `ResponsesWS` implementations against an ephemeral localhost WebSocket server. Only client configuration and stdout are injected; socket transport, protocol dispatch, iteration, and the close handshake are real. All credentials and events are synthetic.

The first six regressions were run before the fix: the completed-response control passed and all five failure/cleanup cases failed because the connection stayed open. The adversarial review then added child-agent termination cases, which exposed premature completion and prevented the initial failure handler from terminating the coordinator on a child's failure.

The final 11 tests cover success, failed/incomplete coordinator responses before and after partial text, each child terminal status followed by coordinator output, both API error shapes, and output-write failures. They verify a clean close handshake, unchanged request settings, preserved output formatting, and propagation of existing errors.

## Validation

- Focused example and existing stable/beta WebSocket suites: **66 passed**.
- `CI=true ./scripts/test`: **7,020 handwritten + 556 generated tests passed**, with the existing two generated test skips; all 12 snapshots passed.
- `pnpm lint`, `pnpm exec tsc --noEmit`, and `pnpm build`: passed.
- Built CJS and ESM public SDK entrypoints on Node **22.0.0, 24.20.0, and 26.7.0**: **126 local WebSocket example scenarios passed**, including omitted/null/explicit-root metadata, child terminal events, API errors, and output failures.
- The changed example compiles against built public declarations with **TypeScript 4.5.5 and 4.9.5**, with `skipLibCheck: false`.

## Adversarial self-review

Repeated the review after adding the child-agent safeguard. The final change has a single connection owner and cleanup path, does not add retries or parser state, preserves streamed output and existing error objects, and adds no response-body data to failure diagnostics. The request URL and beta header were checked in the compatibility smoke tests. API behavior and runtime requirements outside this example are unchanged.

### DIFF
--- a/tests/lib/multi-agent-websocket-example.test.ts
+++ b/tests/lib/multi-agent-websocket-example.test.ts
@@ -1,0 +1,240 @@
+import { once } from 'node:events';
+import { readFileSync } from 'node:fs';
+import { runInNewContext } from 'node:vm';
+
+import OpenAI from 'openai';
+import type { BetaResponse, BetaResponsesServerEvent } from 'openai/resources/beta/responses/responses';
+import { ResponsesWS } from 'openai/resources/beta/responses/ws';
+import ts from 'typescript';
+import { expect, test, vi } from 'vitest';
+import { WebSocketServer } from 'ws';
+
+const filename = 'examples/responses/multi-agent-websocket.ts';
+const source = ts.transpileModule(readFileSync(filename, 'utf-8'), {
+  compilerOptions: {
+    module: ts.ModuleKind.CommonJS,
+    target: ts.ScriptTarget.ES2022,
+    esModuleInterop: true,
+  },
+}).outputText;
+
+type TerminalStatus = 'completed' | 'failed' | 'incomplete';
+type TerminalEvent = Extract<BetaResponsesServerEvent, { type: `response.${TerminalStatus}` }>;
+
+function terminalEvent(status: TerminalStatus): TerminalEvent {
+  const response: BetaResponse = {
+    id: 'resp_example',
+    object: 'response',
+    created_at: 1,
+    status,
+    error: status === 'failed' ? { code: 'server_error', message: 'Synthetic server failure' } : null,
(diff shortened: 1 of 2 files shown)


tests/lib/multi-agent-websocket-example.test.ts
```diff
@@ -0,0 +1,240 @@
+import { once } from 'node:events';
+import { readFileSync } from 'node:fs';
+import { runInNewContext } from 'node:vm';
+
```
mccrearydamion69-crypto: My apples products I have not received them

mccrearydamion69-crypto: Apparently a report

jbeckwith-oai (approved):

PR #2549, opened 2026-09-01
fix(embeddings): preserve numeric output with body overrides

HAYDEN-OAI:
When primary embeddings parameters omit `encoding_format`, a supported `RequestOptions.body` override can return numeric vectors that the SDK incorrectly decodes as base64, turning `[1.25, -2.5]` into `[]`.

Preserve the existing numeric SDK result for default-format calls: leave numeric override embeddings intact and continue decoding base64 override responses. Explicit primary encodings retain their existing behavior. The helper keeps request-option precedence, single getter evaluation, and raw/parsed response accessors.

The three override regressions cover omitted, float, and base64 wire encodings, exact `number[]` inference, a working `.map()` call, immutable caller objects, headers/security precedence, one fetch, and response identity. Only the handwritten helper and its existing test file change.

Validation:

- The reported base64-override `.map()` regression fails before the fix; all 20 embeddings tests pass afterward.
- `pnpm lint`, `pnpm exec tsc --noEmit`, and `pnpm build` pass with Node 24.20.0 and the frozen dependency lockfile.
- Canonical `./scripts/test` passes using a separately managed local Steady server: 7,012 handwritten tests, 556 generated tests, and 12 snapshots; existing generated skips remain.
- Packed npm checks pass for CommonJS, ESM, ES2020 browser bundling, TypeScript consumers, and 1,286 source maps.
- Built CJS/ESM entrypoints pass 48 mock-fetch cases across Node 22.0.0, 24.20.0, and 26.7.0.
- Independent review found no additional correctness or parsing issues.

### DIFF
--- a/src/lib/embeddings.ts
+++ b/src/lib/embeddings.ts
@@ -34,16 +34,16 @@ export function createEmbedding(
     ...options,
     __security: { bearerAuth: true },
   };
-  const shouldDecode = !hasUserProvidedEncodingFormat && requestOptions.body === optimizedBody;
+  const hasBodyOverride = requestOptions.body !== optimizedBody;
   const response: APIPromise<CreateEmbeddingResponse> = client.post('/embeddings', requestOptions);
 
-  // Only decode the optimized body we supplied, not a caller's body override.
-  if (!shouldDecode) {
+  // Explicit encodings return the original response promise unchanged.
+  if (hasUserProvidedEncodingFormat) {
     return response;
   }
 
-  // The default request uses base64 on the wire, but returns the API's default
-  // numeric embedding representation to the caller.
+  // Preserve numeric output for default requests, including body overrides that
+  // change the encoding used on the wire.
   loggerFor(client).debug('embeddings/decoding base64 embeddings from base64');
 
   return response._thenUnwrap((data) => {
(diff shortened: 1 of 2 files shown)


jbeckwith-oai (approved):

marcuswood-oai:
Thanks for fixing the float-override corruption! Clarifying my earlier feedback: base64 pass-through follows the encoding requested in the effective body. The [OpenAPI request contract](https://github.com/openai/openai-openapi/blob/cf79040afd73deda323d2cc869199d4f463bd7ad/openapi.yaml#L34421) allows `float` or `base64`, defaulting to `float`; preserving the previous decoded output is an SDK compatibility choice, not an OpenAPI requirement.

The compatibility/type issue remains: when primary params omit `encoding_format`, a base64 `options.body` override now returns a string instead of the previous `number[]`, while the public overload still promises `number[]`.

Minimal repro (mock fetch, no API call):

```ts
import OpenAI from 'openai';

const client = new OpenAI({
  apiKey: 'test-key',
  fetch: async () => Response.json({
    data: [{ embedding: 'AACgPwAAIMA=' }], // base64 for [1.25, -2.5]
  }),
});
const params = { model: 'text-embedding-3-small', input: 'hello' };
const result = await client.embeddings.create(params, {
  body: { ...params, encoding_format: 'base64' },
});
result.data[0]!.embedding.map(x => x * 2);
// Base eea2292: [2.5, -5]
// PR e352eda: TypeError: ...map is not a function
```

Could we resolve the runtime/type mismatch and make an explicit compatibility decision here? Preserving the old output is one option; if the change is intentional, please document the break and add coverage for the agreed runtime and type behavior. Widening the existing return type would also be a compatibility change.

HAYDEN-OAI:
@marcuswood-oai Fixed in [75974ae3](https://github.com/openai/openai-node/commit/75974ae31c69d381b9e9efa44b603b694930c2e2).

The compatibility decision is to preserve the existing decoded `number[]` result when the primary parameters omit `encoding_format`, including a base64 `options.body` override. Numeric arrays returned by float/default overrides now pass through intact, so the original corruption fix is retained. Explicit primary encodings and the public overloads keep their existing behavior.

The override regression now asserts exact `number[]` inference and executes your `.map()` example. It fails on the previous PR head and passes with this fix; `.asResponse()` still exposes the original wire response.

Validation passed: all 20 embeddings tests; 7,012 handwritten and 556 generated tests; lint, type-check, build, packed-package checks; and built CJS/ESM mock-fetch checks on Node 22.0.0, 24.20.0, and 26.7.0.

marcuswood-oai (approved):

PR #2550, opened 2026-09-01
fix(examples): report multi-agent streaming failures

HAYDEN-OAI:
## Summary

Make the multi-agent SSE example report unsuccessful coordinator responses instead of silently exiting successfully.

The example currently prints message/text events but ignores `response.failed` and `response.incomplete`. Both are valid beta Responses stream events. Unlike named SSE `error` frames, the SDK correctly yields these ordinary response events to the caller. When the stream ends, the example prints a newline and exits 0—even if the coordinator failed before producing any text or after partial output.

## Change

Add a five-line branch in `examples/responses/multi-agent-streaming.ts` that throws a concise diagnostic for failed/incomplete coordinator responses.

- Treat omitted/null ownership and explicit `/root` ownership as the coordinator.
- Keep child-agent terminal events non-fatal so the coordinator can recover and continue.
- Preserve successful output, agent labels, request parameters, and the existing SDK handling of named SSE errors.
- Report only the terminal event type, not response/error payload details.

This is separate from #2533 and #2544, which address WebSocket examples. The SSE example needs no WebSocket-style close handling: its existing iterator closes normally on completion and performs its existing abort cleanup when the loop throws.

The shared ecosystem CLI fixtures also persist offline npm settings in each temporary project’s `.npmrc`. pnpm strips `npm_config_*` environment variables when starting workers, which allowed Node 22’s nested npm installs to make audit requests and time out. The dependency-free wrapper also disables pnpm’s automatic root installation, preventing parallel workers from racing on shared pnpm state. The fixtures retain real npm package installations and the existing 15-second timeout; failures now include subprocess output.

Only this handwritten example and two handwritten test files change. No generated code, SDK APIs, dependencies, or budget rules change.

## Validation

The ecosystem fixture correction passes all three install modes on Node 22.23.2 and all 31 ecosystem CLI/SSE tests on Node 24.20.0. Lint, TypeScript, and the CJS/ESM build pass.

Validation of the streaming change:

- Four executable regressions fail on main because failed/incomplete root responses exit 0; all eight final cases pass with the fix.
- Tests run the actual documented `ts-node` entrypoint and public SDK against a local synthetic HTTP/SSE server. They cover failures before/after partial text, omitted/null/explicit-root ownership, all three child terminal statuses followed by coordinator success, and named SSE error handling. They also check the actual request path/body, exact output, exit status, and absence of credentials/private failure details in new diagnostics.
- Complete `CI=true ./scripts/test`: 7,017 handwritten tests across 164 files, 556 generated tests across 82 suites, and 12 snapshots pass. Existing generated skips are unchanged.
- `pnpm lint`, `pnpm exec tsc --noEmit`, and `pnpm build` pass.
- The changed example compiles against the built SDK with strict TypeScript 4.5.5 and 4.9.5 checks (`skipLibCheck: false`).
- An additional 138-case runtime matrix passes across Node 22.0.0, 24.20.0, and 26.7.0, loading actual built CJS/ESM SDK classes with the transpiled example. It covers root/child terminal events, partial output, SSE errors, original stdout-error identity, and request-signal abort on failure. The same matrix confirms the corresponding baseline failures.

Repeated adversarial self-review and an independent bounded review found no remaining defect. Fixtures are synthetic and require no live API calls or credentials.


jbeckwith-oai (approved):

PR #2551, opened 2026-09-01
fix: parse JSON media types case-insensitively

HAYDEN-OAI:
## Summary

Parse JSON responses regardless of the capitalization of their media type.

[RFC 9110 §8.3.1](https://www.rfc-editor.org/rfc/rfc9110.html#section-8.3.1) defines media-type type/subtype tokens as case-insensitive. The handwritten response parser currently compares them case-sensitively: a valid `Application/JSON` or `application/vnd.openai+JSON` response is returned as a string instead of parsed JSON. Public `models.list()` then silently produces an empty page from that string.

## Change

Add `.toLowerCase()` to the already-extracted, trimmed media type before the existing JSON checks. This is a one-line production fix; it does not add a MIME parser or change the existing classification rules.

The original response header and parameter values remain untouched. Existing text, binary, streaming, 204, JSON parsing, empty-body, error, and request-ID behavior is preserved.

Add public-client regressions for mixed-case JSON and structured `+JSON` suffixes, parameter/header preservation, response accessors and identity, non-enumerable request IDs, pagination, empty JSON, and malformed JSON. Only `src/internal/parse.ts`, the existing handwritten `tests/client-behavior.test.ts`, and the handwritten ecosystem CLI test change. No generated files, dependencies, API declarations, or budget rules change.

The ecosystem CLI fixtures persist offline npm settings in each temporary project’s `.npmrc`, since pnpm drops the corresponding environment variables in workers. The dependency-free wrapper also disables automatic pnpm initialization so parallel workers cannot race on shared state. Real local package installs and the 15-second timeout remain in place; failures include subprocess output.

## Validation

The fixture correction passes all 73 ecosystem CLI, client behavior, and response tests on Node 22.23.2, plus lint and TypeScript checks.

Validation of the parser change:

- Seven new cases fail against the original parser: four return strings, one drops paginated items, one returns an empty string instead of `undefined`, and one returns malformed JSON as text instead of rejecting. Confirmed again after adopting the repository's preferred `Response.json()` fixtures.
- Both focused response suites pass with the fix: 50 tests.
- Complete `CI=true ./scripts/test`: 7,017 handwritten tests across 163 files, 556 generated tests across 82 suites, and 12 snapshots pass. Existing generated skips are unchanged.
- Final `pnpm lint`, `pnpm exec tsc --noEmit`, and `pnpm build` pass.
- Packed npm artifact passes CommonJS, ESM, ES2020 browser bundling, TypeScript consumer checks, optional-peer checks, and 1,286 source-map checks.
- Actual built CJS/ESM SDK entrypoints pass 426 synthetic cases across Node 22.0.0, 24.20.0, and 26.7.0. The matrix covers seven media-type forms, JSON value types, public model/page responses, malformed/empty JSON, text and missing-header fallbacks, raw binary responses, 204 responses, one fetch, accessor identity, and unchanged original headers. The same matrix confirms affected baseline behavior and unchanged controls.
- Emitted parser `.d.ts` and `.d.mts` files are byte-identical to baseline.

Repeated adversarial self-review and an independent bounded final review found no remaining defect. Fixtures are synthetic; no live API calls or credentials are required.


jbeckwith-oai (approved):

markstuart-oai (approved):
The media-type normalization stays local to the existing response-classification boundary and preserves the original header. The focused coverage exercises mixed-case standard and structured JSON types together with pagination, empty-body, malformed-body, and raw-response behavior. I found no substantive correctness or maintainability issues in the current diff.

PR #2560, opened 2026-09-01
fix: preserve alternate Headers inputs

HAYDEN-OAI:
## Summary

A genuine `Headers` object from another fetch implementation, such as `undici.Headers`, is not an instance of the runtime's global constructor. The shared header normalizer treated it as a plain record and silently discarded its entries. This affects both client `defaultHeaders` and per-request `headers`, including request overrides.

- Consume header collections through their shared iteration protocol.
- Read the iterator once and invoke it with its original receiver.
- Retain the existing plain-object overwrite rules, repeated-value handling, explicit-null removals, and header-name validation.
- Keep non-callable iterator metadata on ordinary objects on the existing record path.

The production change is confined to the shared normalizer: seven added lines and five removed lines, including a private iterator-entry type. No new dependency or public API change.

## Regression coverage

Eight tests make real loopback HTTP requests through `client.models.list()`: native Headers, Undici Headers, header records, and header tuples, each supplied through client defaults and request options. They verify the endpoint, exact header values, override precedence, repeated values, null removal, and literal headers named `entries` and `get`.

Five additional regressions cover non-callable iterator metadata and a single iterator lookup with correct receiver binding.

The shared ecosystem CLI fixtures also persist offline npm settings in each temporary project’s `.npmrc`. pnpm strips `npm_config_*` environment variables when starting workers, which allowed Node 22’s nested npm installs to make audit requests and time out. The dependency-free wrapper also disables pnpm’s automatic root installation, preventing parallel workers from racing on shared pnpm state. The fixtures retain real npm package installations and the existing 15-second timeout; failures now include subprocess output.

## Verification

The ecosystem fixture correction passes all 49 selected header and ecosystem CLI tests on Node 22.23.2. Lint, TypeScript, and the CJS/ESM build pass.

- **Before the fix:** both Undici cases fail because headers are missing or defaults win incorrectly; all six native/object/tuple controls pass on Node 22, 24, and 26. [Test-only reproduction](https://github.com/openai/openai-node/actions/runs/33552166587).
- **Original implementation:** full CI passes on Node 22, 24, and 26, including 7,022 handwritten tests, 556 generated tests, and 12 snapshots on each line. Lint, formatting, TypeScript, builds, package checks, ecosystem tests, and benchmarks are green. Existing generated-suite skips are unchanged. [Final CI](https://github.com/openai/openai-node/actions/runs/33553501267).
- **Exact runtime floor:** eight additional packed-package HTTP cases pass on Node 22.0.0 across CommonJS/ESM, native/Undici Headers, and both option locations. The tested normalizer blob is identical to this PR. [Minimum-runtime verification](https://github.com/openai/openai-node/actions/runs/33553308754/job/100007978373).

Adversarial self-review checked merge semantics, nullable metadata, inherited iteration, getter reads, receiver binding, validation, credential isolation in the fixtures, and supported runtime/type compatibility.

Only four handwritten files change. Their paths are absent from the current verified generated snapshot. No workflows, generated-owned files, dependency manifests, or custom-code policy are changed in this PR; the temporary verification workflow is separate. Open PRs were checked for overlap.


jbeckwith-oai (approved):

PR #2561, opened 2026-09-01
fix(pagination): follow next cursors through empty pages

HAYDEN-OAI:
## Summary

`NextCursorPage` has an explicit server-provided `next` cursor, but its `hasNextPage()` delegates to the base implementation, which rejects every empty data array before checking continuation metadata. An empty first or intermediate page can therefore silently stop both item auto-iteration and `iterPages()`, omitting subsequent results.

After the existing `has_more === false` guard, consult `nextPageRequestOptions()` directly. Missing/empty cursors still stop pagination. Other pagination classes are unchanged.

## Regression coverage

Add eight real loopback-HTTP cases through `client.admin.organization.groups.list()`: empty first page, empty intermediate page, ordinary populated pages, and terminal empty page, each exercised through item and page iteration.

The tests use a positive page limit and distinct opaque cursors, and verify yielded items/pages, exact request counts, cursor encoding, retained query parameters, admin authentication, and request headers. Unknown/repeated unexpected requests fail within the synthetic fixture; retries are disabled and servers/connections close in `finally`.

## Validation

- **Before the fix:** the [test-only commit](https://github.com/openai/openai-node/commit/3c37675e9e53fc1242efba044d8b87e0e326211c) produces exactly four regression failures and four passing controls on Node 22, 24, and 26 ([CI run](https://github.com/openai/openai-node/actions/runs/33555440586)). The formatted final test also reproduced the same four failures locally before the production change.
- **After the fix:** [CI on `96b30860`](https://github.com/openai/openai-node/actions/runs/33556279606) passes all 7,017 handwritten tests, 556 generated tests, and 12 generated snapshots on each of Node 22, 24, and 26. Existing skipped tests are unchanged.
- All 20 focused pagination tests pass locally, together with the complete handwritten suite, canonical lint, TypeScript checking, and package build.
- The installed npm tarball passes **16 CJS/ESM cases and 32 native loopback HTTP requests on exact Node 22.0.0**, covering both iteration modes and all four page shapes. This check ran locally without optional peers.
- CI also passes published-source checks with TypeScript 4.9 and 6, package/export checks, ecosystem tests, and benchmarks.

Adversarial self-review covered termination guards, subclass dispatch, cursor/query encoding, request-option and authentication preservation, response ordering, and fixture cleanup. An independent closing review found no actionable bypasses or regressions.

## Scope

One production line plus one focused test file. Both paths are absent from the current verified public generated tree (`65094c1637432c123c310d03e0cda1e7e6f1c997`). No generated resources, public types, dependencies, workflow files, custom-code budgets, or pagination limits change.

### DIFF
--- a/src/core/pagination.ts
+++ b/src/core/pagination.ts
@@ -309,7 +309,7 @@ export class NextCursorPage<Item> extends AbstractPage<Item> implements NextCurs
       return false;
     }
 
-    return super.hasNextPage();
+    return this.nextPageRequestOptions() != null;
   }
 
   nextPageRequestOptions(): PageRequestOptions | null {
--- a/tests/lib/next-cursor-pagination.test.ts
+++ b/tests/lib/next-cursor-pagination.test.ts
@@ -1,0 +1,126 @@
+import { once } from 'node:events';
+import { createServer, type IncomingHttpHeaders } from 'node:http';
+import OpenAI from 'openai';
+import type { NextCursorPageResponse } from 'openai/core/pagination';
+import type { Group } from 'openai/resources/admin/organization/groups/groups';
+
+const firstGroup: Group = {
+  id: 'group_first',
+  created_at: 0,
+  group_type: 'group',
+  is_scim_managed: false,
+  name: 'First group',
+};
+const secondGroup: Group = { ...firstGroup, id: 'group_second', name: 'Second group' };
+
+const cases: Array<{ name: string; pages: NextCursorPageResponse<Group>[] }> = [
+  {
+    name: 'empty first page with a continuation cursor',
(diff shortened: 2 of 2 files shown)


jbeckwith-oai (approved):

sylvesterkaczmarek (commented):
Looks right. `NextCursorPage` is different from the other cursor pages here: continuation comes from the server's `next`, so an empty `data` array isn't enough to stop.

`has_more === false` still wins, and a missing/empty `next` still terminates through `nextPageRequestOptions()`. The tests cover empty first/intermediate pages through both item iteration and `iterPages()`. No blocker from me.

PR #2464, opened 2026-08-25
fix: publish Castiron checks for external-fork pull requests

HAYDEN-OAI:
## Summary

GitHub omits `workflow_run.pull_requests` for external-fork runs, and querying the upstream repository's commit-association endpoint returns no pull requests for those fork commits. That leaves the required `Castiron / budget-only change` and `Castiron / custom-code budget` contexts permanently expected even when the candidate workflow succeeds.

- Centralize trusted Python pull-request association in `custom_code_report.py` and reuse it from trusted report generation, comment publication, and budget evaluation.
- Resolve missing associations from the authenticated source run's `head_repository`, including legitimately renamed forks, then independently re-fetch every candidate PR from `openai/openai-node`.
- Apply equivalent validation in both privileged JavaScript publishers: required commit statuses and failure-report comments.
- Refresh the candidate workflow's pinned reporter SHA-256 after the trusted reporter change.

## Security model

Fork-side associations and PR numbers are discovery hints, never authorization. The trusted paths:

1. Re-fetch the workflow run from the upstream Actions API and verify its repository, immutable candidate SHA, workflow path, completion, and run attempt as applicable.
2. Strictly validate the source repository's owner/name syntax and consistency with authenticated `head_repository` metadata; reject traversal-like components and spoofed identities.
3. Re-fetch each hinted PR from the upstream repository and require an open PR with the exact candidate SHA, the exact source head repository, the intended upstream repository and `main` base ref, and exactly one valid current association.
4. Require the current `main` base SHA wherever budget evaluation or status publication needs freshness; preserve existing stale-run behavior and merge-group validation.

The existing trusted `workflow_run`/main-checkout boundary, bare Git object store, candidate-artifact isolation, least-privilege job permissions, merge-queue protections, and exact required status names remain unchanged. No candidate workflow definition, mutable ref, contributor artifact, or fork-supplied PR number is trusted.

## Affected contributor PRs

- #2463: live fork workflow run `32877584723` has `pull_requests: []`; the updated resolver correctly finds its upstream PR through `cmun2/openai-node`.
- #2444 and #2431: independently reproduced the same fork-only association behavior; both PRs merged while this fix was being prepared.

## Verification

- `env PYTHONDONTWRITEBYTECODE=1 python3 -m unittest discover -s scripts/castiron -p 'test_custom_code*.py'` — 60 tests pass, with one pre-existing skip.
[...]

PR #2562, opened 2026-09-01
chore: raise custom-code budget to 4,000 lines

apcha-oai:
Raise the custom-code budget from 3,000 to 4,000 to unblock SDK generation. Main currently uses 2,983 lines; the current generation candidate uses 3,497, leaving 503 lines under the new limit.

These remaining customizations are technical debt. We do not have time to address that debt at the moment, so this raises the budget to unblock the work and defers the underlying cleanup.

Usage was 1,770 when the original budget landed in #2446. The increase on main came from:

- Reporter and test changes: +696 lines in #2447 and #2464.
- X.509 client integration: +513 lines in #2472, #2473 and #2479.
- Webhook verification wiring: +4 lines in #2475.

Only `.castiron-ratchet.json` changes.

### DIFF
--- a/.castiron-ratchet.json
+++ b/.castiron-ratchet.json
@@ -1,4 +1,4 @@
 {
   "schema_version": 1,
-  "max_custom_patch_lines": 3000
+  "max_custom_patch_lines": 4000
 }


jbeckwith-oai (approved):
so its fine when Alex does it

PR #2567, opened 2026-09-02
fix(realtime): preserve base URL routing components

HAYDEN-OAI:
## Summary

Fix Realtime endpoint construction when a custom `baseURL` contains routing query parameters or a fragment.

- Append `realtime` to the parsed URL pathname in both stable and beta Realtime helpers.
- Preserve routing queries, repeated query values, and escaped path segments; remove fragments before opening the WebSocket.
- Keep the existing Azure GA path/query normalization, explicit session-parameter precedence, and custom `buildRealtimeURL` callback behavior unchanged.

## Reproduction

Given `baseURL: 'https://example.com/v1?route=tenant'`, string concatenation currently puts the endpoint suffix inside the routing parameter:

```text
Before: wss://example.com/v1?route=tenant%2Frealtime&model=gpt-realtime
After:  wss://example.com/v1/realtime?route=tenant&model=gpt-realtime
```

The old URL requests the wrong path and corrupts the routing value. A fragment in the base URL likewise absorbs the endpoint suffix and leaves a fragment on the WebSocket URL. Both public native-WebSocket and Node `ws` wrappers use these helpers.

## Regression coverage

Add 21 cases covering query-bearing base URLs with and without trailing slashes, root paths, encoded path segments, repeated routing parameters, fragments, transcription intent, and stable/beta public constructors and asynchronous sideband factories with rotating credentials. Nineteen cases fail before the fix; the two query-free controls already pass.

## Validation

- 178 focused Realtime tests passed.
- Full local suite: 7,030 handwritten tests, 556 generated tests, and 12 snapshots passed.
- Formatting, lint, root typechecking, build, and published source checks with TypeScript 4.9 and 6 passed.
- 84 offline checks through built public native-WebSocket exports passed across Node 22.0.0, 24.20.0, and 26.7.0, each in CommonJS and ESM. These capture actual constructor arguments with synthetic credentials; they do not make live API connections.
- [Full CI passed on this commit](https://github.com/openai/openai-node/actions/runs/33574064054): Node 22/24/26 test matrix, packed-package checks on Node 22/24, exact Node 22.0 X.509 compatibility fixture, all 14 ecosystem fixtures, build, lint, and performance benchmarks.
- Adversarial self-review and independent review covered security, compatibility, URL edge cases, and regression coverage; no unresolved findings.

Scope is four handwritten files only. Generated HTTP-client URL construction, dependencies, and CI policies are unchanged.

### DIFF
--- a/src/beta/realtime/internal-base.ts
+++ b/src/beta/realtime/internal-base.ts
@@ -186,8 +186,9 @@ export function buildRealtimeURL(
   const hasModel = !!config.model;
   const hasCallID = !!config.callID;
 
-  const path = '/realtime';
-  const url = new URL(baseURL + (baseURL.endsWith('/') ? path.slice(1) : path));
+  const url = new URL(baseURL);
+  url.pathname += url.pathname.endsWith('/') ? 'realtime' : '/realtime';
+  url.hash = '';
 
   if (hasModel === hasCallID) {
     throw new Error('Pass exactly one of `model` or `callID` when opening a Realtime WebSocket.');
--- a/src/realtime/internal-base.ts
+++ b/src/realtime/internal-base.ts
@@ -281,19 +281,17 @@ export function buildRealtimeURL(
     return url;
   }
 
-  let url: URL;
+  const url = new URL(baseURL);
   if (azure) {
-    url = new URL(baseURL);
     const basePath = url.pathname.replace(/\/+/g, '/').replace(/\/+$/, '');
     const versionedPath = basePath.endsWith('/v1') ? basePath : `${basePath}/v1`;
     url.pathname = `${versionedPath}/realtime`;
     url.search = '';
-    url.hash = '';
   } else {
-    const path = '/realtime';
-    url = new URL(baseURL + (baseURL.endsWith('/') ? path.slice(1) : path));
+    url.pathname += url.pathname.endsWith('/') ? 'realtime' : '/realtime';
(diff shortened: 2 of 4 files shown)


sylvesterkaczmarek (commented):
I checked the stable and beta URL builders against the routing/precedence boundaries here. Parsing `baseURL` before appending the endpoint keeps the pathname mutation separate from the query and fragment, so routing parameters are no longer absorbed into the `/realtime` suffix. Assigning through `URL.pathname` also preserves the encoded path cases covered by the regression matrix.

The later `searchParams.set(...)` calls still give explicit session parameters (`model`, `call_id`, `intent`) precedence over same-named values from the base URL while leaving unrelated and repeated routing parameters intact. The Azure path remains intentionally separate: it normalizes `/v1`, clears the inherited search, and then applies its GA parameters. Clearing `hash` for both paths is also correct before opening the WebSocket.

The browser/Node and stable/beta constructor coverage exercises the public call paths rather than only the helper. I don't see a blocking URL-construction regression in the current head.

jbeckwith-oai (approved):
Reviewed the exact head. The pathname-based construction preserves routing queries, repeated values, encoded path segments, and explicit session-parameter precedence while stripping fragments; Azure and custom callback behavior remain intact. Focused Realtime tests passed locally (178), root typechecking passed, and CI is green.

PR #2568, opened 2026-09-02
fix(format): preserve file-list scope

HAYDEN-OAI:
## Problem

`scripts/fast-format` accepts one file path per line, but sends that input through whitespace- and quote-parsing `xargs`.

- A file containing only blank lines passes the existing nonempty-file check. Oxfmt is then invoked without any paths and formats the entire working directory, including unrelated files.
- Paths containing spaces are split and silently skipped; unmatched quotes cause the command to fail.

## Change

Normalize CRLF and empty lines, guard against an empty normalized selection, and pass NUL-delimited paths to both tools with an explicit end-of-options separator. Existing file-type selection, xargs batching, generated-file cleanup, and pipeline failure propagation remain intact.

The production change is six added and four removed lines in the handwritten shell script. No SDK runtime code, generated files, dependencies, or CI policies change.

## Regression coverage

Add 11 isolated cases that run the actual script with the repository's pinned Oxlint and Oxfmt. The fixtures require both the `var` → `const` lint fix and formatting to reach each selected file, while checking that an unlisted file remains byte-for-byte unchanged.

Coverage includes empty/LF/CRLF-only lists, ordinary and spaced paths, single and double quotes, leading dashes, blank lines mixed with CRLF, a final path without a newline, JSON-only selections, and deleted paths. The initial baseline reproducer had six failures and three passing controls before the fix.

## Validation

- Refreshed onto current main `3cf755ec`; range-diff confirmed the reviewed patch was unchanged.
- Full local suite: 7,020 handwritten tests, 556 generated tests, and 12 snapshots passed.
- Focused formatting/lint suites: 20 tests passed.
- Canonical lint, TypeScript checking, build, Bash syntax, and diff checks passed.
- [Full CI passed on this commit](https://github.com/openai/openai-node/actions/runs/33588967078): Node 22/24/26 tests, packed-package checks on Node 22/24, the exact Node 22.0 compatibility fixture, all 14 ecosystem fixtures, build, lint, and benchmarks.
- Adversarial self-review and independent closing review found no unresolved blockers.

Execution validation used Linux/GNU tools. macOS/BSD portability was reviewed but not executed; the temporary Bash/symlink fixture is skipped on native Windows.

### DIFF
--- a/scripts/fast-format
+++ b/scripts/fast-format
@@ -28,12 +28,14 @@ echo "==> Fixing generated lint issues"
 node ./scripts/lint-generated.cjs --fix --file-list "$FILE_LIST"
 
 echo "==> Running oxlint --fix"
-LINT_FILES="$(grep -E '\.(ts|tsx|mts|cts|js|jsx|mjs|cjs)$' "$FILE_LIST" || true)"
+# Match the generated linter's line-based input, without letting xargs split paths.
+FILES="$(sed -e $'s/\r$//' -e '/^$/d' "$FILE_LIST")"
+LINT_FILES="$(printf '%s\n' "$FILES" | grep -E '\.(ts|tsx|mts|cts|js|jsx|mjs|cjs)$' || true)"
 if [ -n "$LINT_FILES" ]; then
-   echo "$LINT_FILES" | xargs ./node_modules/.bin/oxlint --fix --no-error-on-unmatched-pattern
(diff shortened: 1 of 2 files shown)


jbeckwith-oai (approved):
Reviewed the exact head. The normalized, NUL-delimited path flow keeps the formatter scoped to the line-based file list, handles CRLF/blank inputs and shell-significant filenames, and retains the existing lint/format selection and failure behavior. Bash syntax and all 11 focused regression cases passed locally; CI is green.

PR #2569, opened 2026-09-02
fix(ci): propagate Node matrix check failures

HAYDEN-OAI:
## Problem

The `Read policy matrix` step embeds the policy checker inside `echo`. If the checker fails, `echo` still returns zero and appends an empty `matrix=` output. The producer step is marked successful even though the downstream test job cannot parse its matrix.

## Change

Capture the matrix in a standalone assignment before appending it to `GITHUB_OUTPUT`. The existing [default Bash `-e` behavior](https://docs.github.com/en/actions/reference/workflows-and-actions/workflow-syntax#jobsjob_idstepsshell) then preserves the checker failure at the policy step, with its original diagnostic and without publishing a matrix. Successful output is unchanged.

This is a CI failure-reporting fix, not a new policy requirement or a security-bypass fix.

## Regression coverage

Three isolated process-level tests execute the actual workflow command with the real policy checker and copied repository inputs:

- Successful policy output is appended byte-for-byte, preserving an existing output and a path containing spaces.
- Drift between package metadata and the policy fails with the checker diagnostic and leaves the output unchanged.
- A missing policy document preserves the read failure without publishing a matrix.

Before the fix, both error cases returned zero; the success control passed. All three pass after the fix. The Bash fixture is skipped on native Windows because this workflow runs on Ubuntu.

## Validation

- Final local suite: 7,012 handwritten tests, 556 generated tests, and 12 snapshots passed.
- [Full CI on this commit passed](https://github.com/openai/openai-node/actions/runs/33590460969), including the same suites on Node 22/24/26, all 14 ecosystem fixtures, packed-package checks on Node 22/24, the exact Node 22.0 compatibility fixture, build, lint, and benchmarks.
- Root TypeScript checking, canonical lint, package build, and diff checks passed locally.
- Adversarial self-review and independent closing review found no unresolved blockers.

Only the handwritten CI workflow and one regression test change. SDK runtime code, generated files, dependencies, workflow permissions/triggers, and supported Node versions are unchanged.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -106,7 +106,9 @@ jobs:
 
       - name: Read policy matrix
         id: policy
-        run: echo "matrix=$(node --experimental-strip-types scripts/check-node-version-policy.ts --matrix)" >> "$GITHUB_OUTPUT"
+        run: |
+          matrix="$(node --experimental-strip-types scripts/check-node-version-policy.ts --matrix)"
+          echo "matrix=$matrix" >> "$GITHUB_OUTPUT"
 
   test:
     name: test (${{ matrix.node-version }})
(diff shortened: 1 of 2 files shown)


jbeckwith-oai (approved):
Reviewed the exact head. A standalone assignment correctly propagates the policy checker's exit status under the workflow's Bash `-e` semantics and avoids publishing a partial matrix, while leaving successful output unchanged. The process-level regression coverage exercises both failure modes and the success path; CI is green.

PR #2570, opened 2026-09-02
fix(examples): report audio demo failures

HAYDEN-OAI:
## Summary

The speech-to-text and text-to-speech examples log rejected API requests or audio-tool failures, but `main().catch(console.error)` leaves the process exit status at zero. Shell scripts and smoke checks therefore report success even when recording, transcription, synthesis, or playback failed.

Wrap each entrypoint in the existing async `run()` pattern and set `process.exitCode = 1` after logging the error. Successful runs still exit zero, and output can flush normally without forcing `process.exit()`.

## Regression coverage

Add six process-level cases covering HTTP rejection, external audio-tool failure, and success for both examples. They run the actual TypeScript entrypoints and public SDK against a loopback HTTP server, with synthetic `ffmpeg`/`ffplay` executables instead of hardware or live API calls. Assertions verify the exit status, preserved diagnostics, request/recording/playback boundaries, synthetic audio bytes, and cleanup. The fixture does not depend on recording-device arguments and is compatible with the separate ALSA fix in #2564.

Before the fix, the four failure cases fail because the examples exit zero; both success controls pass. After the fix, all six pass. Unix-only executable fixtures are explicitly skipped on native Windows.

## Verification

- Focused audio suites: 47 tests passed.
- Full canonical `./scripts/test`: 7,015 handwritten tests, 556 generated tests, and 12 snapshots passed (existing generated skips unchanged).
- `pnpm exec tsc --noEmit`, canonical `./scripts/lint`, and `./scripts/build` passed.
- Direct before/after CLI reproductions on the exact Node 22.0.0 floor confirm API and recording failures change from exit zero to one while preserving diagnostics.
- [Branch CI](https://github.com/openai/openai-node/actions/runs/33601168381) passed on Node 22, 24, and 26, including all 14 ecosystem integrations, packed-package checks on supported LTS lines, and performance benchmarks.
- Adversarial self-review and independent closing review found no remaining security, compatibility, edge-case, or regression-test blockers.

Only handwritten examples and a new regression test change. No SDK API, generated code, dependency, or lockfile changes.

### DIFF
--- a/examples/audio/speech-to-text.ts
+++ b/examples/audio/speech-to-text.ts
@@ -16,4 +16,13 @@ async function main(): Promise<void> {
   console.log(transcription.text);
 }
 
-main().catch(console.error);
+async function run() {
+  try {
+    await main();
+  } catch (error) {
+    console.error(error);
+    process.exitCode = 1;
+  }
+}
+
+void run();
--- a/examples/audio/text-to-speech.ts
+++ b/examples/audio/text-to-speech.ts
@@ -20,4 +20,13 @@ async function main(): Promise<void> {
   await playAudio(response);
 }
 
-main().catch(console.error);
+async function run() {
+  try {
+    await main();
+  } catch (error) {
+    console.error(error);
+    process.exitCode = 1;
+  }
+}
+
+void run();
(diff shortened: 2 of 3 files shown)


jbeckwith-oai (approved):
Reviewed the exact head. Both examples now preserve their diagnostics while returning a failing process status for API, recording, synthesis, and playback errors; successful runs remain unchanged and can flush normally. The process-level tests cover each boundary with loopback HTTP and synthetic tools, and exact-head CI is green.

PR #2572, opened 2026-09-02
fix(examples): find structured-output tool calls

HAYDEN-OAI:
## Summary

Find the first function call in the structured-output tool example instead of assuming that it is the first response output item.

Responses can place a message before a function call. The example currently throws `Expected function call` for that valid response even though `responses.parse()` has already parsed the tool arguments. An empty output also produces an accidental `TypeError`.

## Changes

- Select the first output item whose type is `function_call`.
- Keep the existing `Expected function call` error when no function call exists, including an empty output.
- Add four regression cases that execute the actual example CLI and public SDK against a synthetic loopback HTTP server: function call first, message before the function call, message only, and empty output.

The production change is two lines in one handwritten example. It does not change SDK behavior, generated code, dependencies, tool schemas, or handling of multiple function calls.

## Validation

- Before the fix, the two regression cases fail and the two controls pass; after the fix, all four pass.
- The real example also passes the success cases on the exact minimum supported Node.js runtime, 22.0.0.
- Full branch CI passes on Node 22, 24, and 26: 7,013 handwritten tests, 556 generated tests, and 12 snapshots per runtime. Existing generated skips are unchanged.
- All 14 ecosystem projects pass, including Deno, Bun, browser, Cloudflare Worker, and Vercel Edge checks.
- Build, lint, type checking, published-source checks with TypeScript 4.9 and 6, packed-package checks, the minimum-runtime authentication check, and benchmarks pass.
- Deliberately adversarial self-review and an independent source review found no remaining blocking issue in this bounded change. Tests use synthetic responses and credentials, make no live OpenAI API calls, verify the parsed result and exit behavior, and clean up child processes and connections.

Branch validation: https://github.com/openai/openai-node/actions/runs/33658578443

### DIFF
--- a/examples/responses/structured-outputs-tools.ts
+++ b/examples/responses/structured-outputs-tools.ts
@@ -46,9 +46,9 @@ async function main() {
 
   console.log(rsp);
 
-  const functionCall = rsp.output[0]!;
+  const functionCall = rsp.output.find((item) => item.type === 'function_call');
 
-  if (functionCall.type !== 'function_call') {
+  if (functionCall?.type !== 'function_call') {
     throw new Error('Expected function call');
   }
 
--- a/tests/lib/structured-output-tool-example.test.ts
+++ b/tests/lib/structured-output-tool-example.test.ts
@@ -1,0 +1,162 @@
+import { spawn } from 'node:child_process';
+import { once } from 'node:events';
+import { createServer } from 'node:http';
+import path from 'node:path';
+import { inspect } from 'node:util';
+import type {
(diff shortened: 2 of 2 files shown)


jbeckwith-oai (approved):
Reviewed the exact head. Selecting the first discriminated `function_call` handles valid message-first output without changing multiple-call semantics, and the optional guard preserves the intended error for message-only and empty responses. The process-level coverage exercises all four shapes through the public SDK; exact-head CI is green.

PR #2573, opened 2026-09-02
fix(examples): serialize missing book lookup results

HAYDEN-OAI:
## Summary

Return `null` when a book lookup misses in the five handwritten Chat Completions function/tool-call examples.

These examples advertise `get(id: string)`, but `db.find(...)!` can still return `undefined` at runtime. The next request serializes the result with `JSON.stringify`, so an unknown ID causes the function/tool message's `content` field to be omitted. Returning `null` makes that result the explicit JSON string `"null"`; successful lookups are unchanged.

The production change is one line in each example. No generated SDK code, dependencies, or public SDK APIs change.

## Regression coverage

- Ten cases run the actual example entrypoints against a loopback HTTP server: one existing and one missing ID for each of the five examples.
- Check both outgoing requests, function/tool association, successful book serialization, and the missing-result `"null"` content.
- Cover ordinary JSON responses and streaming responses with function arguments split across SSE events. Fixtures and credentials are synthetic; the tests make no live API calls.

## Validation

[Branch CI for commit `18778d2`](https://github.com/openai/openai-node/actions/runs/33676058452) passed:

- Node 22, 24, and 26: 7,019 handwritten tests, 556 generated tests, and 12 snapshots on each version, including all ten new regression cases.
- All 14 ecosystem projects, build, lint, TypeScript compatibility checks, and performance benchmarks.
- Packed-package checks and the exact Node 22.0.0 minimum-runtime check.

### DIFF
--- a/examples/chat-completions/function-call-diy.ts
+++ b/examples/chat-completions/function-call-diy.ts
@@ -140,7 +140,7 @@ async function search(name: string) {
 }
 
 async function get(id: string) {
-  return db.find((item) => item.id === id)!;
+  return db.find((item) => item.id === id) ?? null;
 }
 
 main();
--- a/examples/chat-completions/function-call-stream-raw.ts
+++ b/examples/chat-completions/function-call-stream-raw.ts
@@ -183,7 +183,7 @@ async function search(name: string) {
 }
 
 async function get(id: string) {
-  return db.find((item) => item.id === id)!;
+  return db.find((item) => item.id === id) ?? null;
 }
 
 main();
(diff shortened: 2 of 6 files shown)


jbeckwith-oai (approved):
Reviewed the exact head. Replacing the non-null assertion with `?? null` gives all five examples an explicit JSON-serializable miss value while leaving successful lookups unchanged. The process-level matrix covers existing and missing IDs across ordinary, raw-streaming, accumulated-streaming, and tool-call flows. A full branch run passed on this commit; the latest exact-head checks are still rerunning.

PR #2574, opened 2026-09-02
fix(responses): parse namespaced function tools

HAYDEN-OAI:
## Summary

Resolve Responses function tools by both namespace and function name when parsing their arguments.

The API supports [function tools grouped in namespaces](https://developers.openai.com/api/docs/guides/function-calling#defining-namespaces), but the parsing helper only searches top-level tools. Namespaced strict or helper-created tools therefore receive `parsed_arguments: null`, or use a same-named top-level tool's parser.

The shared lookup now searches only the matching namespace's immediate function children when a namespace is present. The stream-finalization detection also recognizes those functions. Existing argument parsers, callback behavior, response-status handling, and exported signatures are unchanged. No generated files or dependencies change.

## Regression coverage

- 21 loopback HTTP cases exercise JSON parsing, stream creation, and stream retrieval through the public SDK with synthetic responses.
- Verify same-name isolation, unknown namespaces, custom/non-strict tools, parser receiver and callback behavior, incomplete responses, sanitized errors, and parser-error identity.
- Six helper cases cover nullish versus empty namespaces, optional namespace-child fields, and parser-error identity.

## Validation

[Branch CI for commit `60060a7`](https://github.com/openai/openai-node/actions/runs/33678739685) passed:

- Node 22, 24, and 26: 7,036 handwritten tests, 556 generated tests, and 12 snapshots on each version, including all 27 new cases.
- All 14 ecosystem projects, build, lint, TypeScript 4.9/6 compatibility checks, and performance benchmarks.
- Packed-package checks and the exact Node 22.0.0 minimum-runtime check.

### DIFF
--- a/src/lib/ResponsesParser.ts
+++ b/src/lib/ResponsesParser.ts
@@ -2,6 +2,7 @@ import { OpenAIError } from '../error';
 import type { ChatCompletionTool } from '../resources/chat/completions';
 import type {
   FunctionTool,
+  NamespaceTool,
   ParsedContent,
   ParsedResponse,
   ParsedResponseFunctionToolCall,
@@ -97,7 +98,8 @@ export function maybeParseResponse<
 }
 
 /**
- * Parses completed response text and strict function-tool arguments.
+ * Parses completed response text and strict function-tool arguments, matching
+ * namespaced functions by both namespace and name.
  *
  * Incomplete or nonterminal responses keep their parsed values as `null`, and
  * `output_parsed` returns the first successfully parsed output-text item.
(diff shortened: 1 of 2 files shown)


jbeckwith-oai (approved):
Reviewed the exact head. The shared lookup now keys namespaced function calls by both namespace and function name, preserves nullish top-level behavior, and avoids cross-namespace or custom-tool fallback. Auto-parse detection and stream finalization use the same matching semantics, with parser identity/callback, incomplete-response, and sanitized-error behavior covered. Exact-head CI is green.

PR #2578, opened 2026-09-03
feat(api): allow omitted function-call argument event names

markstuart-oai:
## Summary

Function-call argument completion event types allow name to be absent.

## Changes

- Make name optional on stable and beta function-call argument completion events.
- Check for undefined before using name as a string, or read the corresponding function-call item's name.


HAYDEN-OAI (approved):
Reviewed the schema and stable, beta, and WebSocket event types. The optional name is consistent across these surfaces, and stream handling does not assume it is present. No substantive findings.

api_reference/openapi.transformed.yml
```diff
@@ -53305,7 +53305,6 @@ components:
       required:
       - type
```
mccrearydamion69-crypto: 👋🏾

PR #2585, opened 2026-09-03
fix(examples): annotate Responses WebSocket turn result

sylvesterkaczmarek:
## Summary

Follow-up to the legacy TypeScript compatibility issue noted in #2544.

The standard Responses WebSocket example still reports `TS7022` on the `turnResult` initializer under TypeScript 4.5.5 and 4.9.5. `runTurn()` already has an explicit `Promise<RunTurnResult>` return type, so the local result can be annotated directly without changing runtime behavior.

This adds the explicit `RunTurnResult` annotation to the initializer, avoiding the legacy compiler's circular inference failure while leaving the example logic unchanged.

## Validation context

The failure is documented in the validation notes for #2544 on current main for both TypeScript 4.5.5 and 4.9.5. This change is compile-time only and does not alter the WebSocket flow.


HAYDEN-OAI (approved):
Reviewed the RunTurnResult definition, runTurn's return path, and the multi-turn call site. The explicit annotation matches the existing Promise<RunTurnResult> contract and changes no runtime behavior or API payload. No substantive correctness, security, or compatibility findings.

Static review; no tests run.

PR #2587, opened 2026-09-03
chore: migrate to forked steady

apcha-oai:
migrate to forked steady

### DIFF
--- a/.github/CONTRIBUTING.md
+++ b/.github/CONTRIBUTING.md
@@ -30,7 +30,7 @@ PowerShell instead, add the Git for Windows `bin` directory that contains `bash.
 To set up the repository, run:
 
 ```sh
-$ pnpm install
+$ ./scripts/bootstrap
 $ pnpm build
 ```
 
--- a/.github/CONTRIBUTING.md
+++ b/.github/CONTRIBUTING.md
@@ -142,6 +142,20 @@ $ pnpm link --global openai
 
 ## Running tests
 
+The mock server uses [the OpenAI Steady fork](https://github.com/openai-oss-forks/steady).
+`scripts/steady/settings` pins its full Git commit and Deno 2.7.11 runtime
+checksums. `./scripts/steady/install` fetches that source, verifies the runtime,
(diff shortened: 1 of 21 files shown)


.github/CONTRIBUTING.md
mccrearydamion69-crypto: 👋🏾

tests/mock-launcher.test.ts
mccrearydamion69-crypto: ✌️

jbeckwith-oai (dismissed):
Two in-scope lifecycle issues need addressing before approval:\n\n1. The atomic publisher is not safe under concurrent Windows/Git Bash bootstraps.  only treats  and  as contention, but Node's directory rename can report  or  when another process wins. Handle those codes as contention only when the destination now exists, then let  authenticate the winner; add focused coverage.\n2. The revision/version-keyed source, runtime, and dependency caches have no invalidation or bound. This repository's cache policy requires an explicit lifetime. Please add a safe bounded cleanup policy (without replacing/deleting active Windows executables) or another enforced lifetime mechanism.

jbeckwith-oai (changes requested):
Two in-scope lifecycle issues need addressing before approval. First, the atomic publisher is not safe under concurrent Windows and Git Bash bootstraps. The publish_directory helper only treats EEXIST and ENOTEMPTY as contention, but Node directory rename can report EPERM or EACCES when another process wins. Handle those codes as contention only when the destination now exists, then let steady_verify authenticate the winner, and add focused coverage. Second, the revision and version-keyed source, runtime, and dependency caches have no invalidation or bound. This repository cache policy requires an explicit lifetime. Please add a safe bounded cleanup policy, without replacing or deleting active Windows executables, or another enforced lifetime mechanism.

jbeckwith-oai (approved):
Re-reviewed the requested fixes. Concurrent publication is guarded and covered; the revision-keyed cache now has a documented 30-day idle lifetime with serialized pruning and process leases that preserve active generations. HTTPS-only redirects are enforced and exact-head checks are green.

PR #2590, opened 2026-09-04
fix(embeddings): preserve numeric responses after request hooks

HAYDEN-OAI:
## Summary

Default-format `embeddings.create()` requests can receive valid float embeddings after a protected `prepareOptions` hook or a request's `toJSON()` changes the wire encoding. The helper only preserved numeric arrays when it detected a direct `RequestOptions.body` override before dispatch, so later customization incorrectly sent those arrays through the base64 decoder.

Reproduced through the public client with synthetic fetch: the serialized request asks for `float`, the raw response contains `[1.25, -2.5, 3.75, 4.5]`, but the SDK returned `[1.551560886955465e-36]`.

- Preserve already-numeric arrays at the decoding boundary; remove the stale body-identity flag.
- Cover in-place hook mutation, hook body replacement, and JSON serialization, including raw/parsed response accessors and unchanged caller input.
- Extend the existing sparse-parser test with a frozen numeric entry beside a base64 entry, protecting identity and mixed decoding.

This is a small follow-up to merged #2549. Explicit float/base64 requests, base64 decoding, request precedence, public types, and promise behavior remain unchanged.

## Ownership and scope

Only the handwritten `src/lib/embeddings.ts` and its existing handwritten test change. Neither file belongs to the pinned Castiron generated snapshot. No generated files, dependencies, exports, or parsing infrastructure change.

## Verification

- The three new wire-customization regressions fail on unchanged main and pass after the fix.
- Both embeddings suites: **23 passed** on Node 24.19.0 and 22.22.2.
- Built CommonJS/ESM entrypoints: **36 cases passed** across Node **22.0.0**, 24.19.0, and 26.7.0, including default/explicit-format controls and all three customization paths.
- `tsc --noEmit`, canonical `./scripts/build`, changed-file lint/format checks, and `git diff --check` pass.
- Full handwritten suite with the host's unrelated `AWS_BEDROCK_BASE_URL` unset: **7,409 passed, 1 failed**. The remaining `oxlint-config` package-command test fails with `ERR_PNPM_VERIFY_DEPS_BEFORE_RUN`; the same failure reproduces in an unchanged-main worktree.
- Two independent adversarial reviews covered security, compatibility, malformed/out-of-contract inputs, sparse/frozen/custom-parser values, response identity, and regression gaps. The suggested frozen-entry coverage was added and the final review was clean.

### Local tooling limitation

A fresh `pnpm install --frozen-lockfile` was blocked because the configured registry lacks publication-time metadata for `qs@6.16.0`. Verification used already-available local dependencies (Vitest 4.1.10, oxlint 1.75.0, oxfmt 0.60.0; TypeScript 6.0.3). Full repository formatting also reports two unchanged files with that older formatter, reproduced on main. CI remains the authoritative check with the repository's pinned toolchain. No registry, lockfile, or supply-chain safeguards were changed.

### DIFF
--- a/src/lib/embeddings.ts
+++ b/src/lib/embeddings.ts
@@ -34,7 +34,6 @@ export function createEmbedding(
     ...options,
     __security: { bearerAuth: true },
   };
-  const hasBodyOverride = requestOptions.body !== optimizedBody;
   const response: APIPromise<CreateEmbeddingResponse> = client.post('/embeddings', requestOptions);
 
   // Explicit encodings return the original response promise unchanged.
@@ -55,8 +54,8 @@ export function createEmbedding(
         if (index in embeddings) {
           const embeddingBase64Obj = embeddings[index] as Embedding;
           const { embedding } = embeddingBase64Obj;
-          // A body override can request float embeddings or omit the format.
-          if (hasBodyOverride && Array.isArray(embedding)) {
+          // Request hooks and serialization can also select float embeddings.
+          if (Array.isArray(embedding)) {
             continue;
           }
           embeddingBase64Obj.embedding = toFloat32Array(embedding as unknown as string);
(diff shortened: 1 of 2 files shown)


jbeckwith-oai (approved):
Validated the fix at the exact head. Numeric embedding arrays are preserved after request-hook mutation, replacement, or serialization, while encoded strings still decode. The tests also preserve APIPromise response identity, sparse entries, immutable numeric entries, and single-fetch behavior. Exact-head checks are green.

PR #2592, opened 2026-09-04
fix(examples): stop polling failed training files

HAYDEN-OAI:
## Summary

The fine-tuning example waits indefinitely when an uploaded file reaches the terminal `error` status: it sleeps for another second and retrieves the failed file again instead of stopping.

Stop that loop with a concise error and reuse the example's existing error-reporting/nonzero-exit path. The message includes only the already-logged file ID, not file contents or `status_details`. Successful processing is unchanged, and a failed file never reaches fine-tuning job creation.

The implementation is **two lines** in the handwritten example. Its existing offline regression harness now models file-status progression and checks:

- Failure on the first processing check.
- An uploaded file that subsequently fails.
- Exact retrieval and wait counts, error reporting, exit status 1, and no job creation after failure.
- An uploaded file that processes successfully and does start a job.

Only the example and its handwritten test change; no generated files, SDK behavior, dependencies, or polling framework changes.

## Verification

- The two new failure regressions fail on unchanged main because it polls again after the terminal error.
- **27 related tests pass** on Node 24.19.0 and 22.22.2, including all existing fine-tuning-job lifecycle tests.
- `tsc --noEmit`, changed-file formatting (Oxfmt 0.62.0), lint, and `git diff --check` pass.
- The tests execute the actual example with the real SDK and injected fetch; no API key, live request, or real timer delay is required.
- Adversarial review covered privacy, terminal-state handling, extra polling/job creation, and test fidelity. Tightened the upload fixture to start at `uploaded`, then repeated review with no remaining findings.

Local tests used cached Vitest 4.1.10 and Oxlint 1.76.0. Fresh frozen installation is still blocked by missing `qs@6.16.0` publication metadata in the configured registry; CI will verify the pinned toolchain. No registry or supply-chain settings were changed.

### DIFF
--- a/examples/fine-tuning/fine-tuning.ts
+++ b/examples/fine-tuning/fine-tuning.ts
@@ -32,6 +32,8 @@ async function main() {
 
     if (file.status === 'processed') {
       break;
+    } else if (file.status === 'error') {
+      throw new Error(`File processing failed for ${file.id}`);
     } else {
       await new Promise((resolve) => setTimeout(resolve, 1000));
     }
--- a/tests/lib/fine-tuning-example.test.ts
+++ b/tests/lib/fine-tuning-example.test.ts
@@ -18,13 +18,18 @@ const source = ts.transpileModule(fs.readFileSync(filename, 'utf-8'), {
   fileName: filename,
 }).outputText;
 
-async function runExample(initialStatus: Status, followingStatuses: Status[]) {
+async function executeExample(
+  initialStatus: Status,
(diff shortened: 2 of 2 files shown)


jbeckwith-oai (approved):
Validated the exact-head fix. Terminal file processing errors now stop before another poll or fine-tuning job creation, while uploaded-to-processed success remains unchanged. The focused harness exercises the real example, exact retrieval and wait counts, error and exit behavior, and successful job creation. Exact-head checks are green.

PR #2593, opened 2026-09-04
fix(examples): use the default audio recording input

HAYDEN-OAI:
## Summary

Use `recordAudio`'s documented default input (index 0) in the speech-to-text example instead of hardcoding input 4. The example currently fails before transcription on a machine with only input zero available.

This is a one-line executable-example change. The SDK helper, five-second timeout, transcription model, and error/exit behavior are unchanged.

## Regression coverage

Strengthen the existing audio subprocess harness so its synthetic `ffmpeg` accepts only input zero and records the selected input. The three speech-to-text cases now verify the selected input while preserving successful multipart upload, HTTP rejection, and recorder-failure/no-request behavior. The three playback controls are unchanged.

All three recording regressions fail on the original example: it selects `:4` and never reaches the expected transcription request. They pass after removing the override. Tests use synthetic audio/credentials and a loopback API, not microphone hardware or a live API.

## Verification

- 58 focused audio tests pass on Node 24.19.0 and 22.22.2.
- Independent executable-example smoke checks pass on exact Node 22.0.0 and Node 26.7.0, including the actual multipart request and one transcription call.
- TypeScript `--noEmit`, changed-file lint, pinned Oxfmt 0.62.0, and `git diff --check` pass.
- Adversarial security, compatibility, lifecycle, and regression-test review found no blockers.
- Both changed files are wholly handwritten and absent from the current pinned Castiron snapshot `efa35b8b45ad152d9b5f9c327c48e35bca65c7ce`.

Local checks used the available cached Vitest 4.1.10, TypeScript 6.0.3, and Oxlint 1.76.0. A fresh pinned install remains blocked by the configured registry's missing publication-time metadata for `qs@6.16.0`; no dependency, lockfile, registry, or install-policy changes are included. CI will run the pinned repository toolchain.

### DIFF
--- a/examples/audio/speech-to-text.ts
+++ b/examples/audio/speech-to-text.ts
@@ -5,7 +5,7 @@ const openai = new OpenAI();
 
 async function main(): Promise<void> {
   console.log('Recording for 5 seconds...');
-  const response = await recordAudio({ timeout: 5000, device: 4 });
+  const response = await recordAudio({ timeout: 5000 });
 
   console.log('Transcribing...');
   const transcription = await openai.audio.transcriptions.create({
--- a/tests/lib/audio-example-exit-status.test.ts
+++ b/tests/lib/audio-example-exit-status.test.ts
@@ -16,8 +16,10 @@ const audioTool = [
   "const tool = require('node:path').basename(process.argv[1]);",
   "const failed = process.env.AUDIO_TOOL_FAILURE === 'true';",
   "if (tool === 'ffmpeg') {",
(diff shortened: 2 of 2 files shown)


jbeckwith-oai (approved):
Validated the exact-head example fix. Removing the hardcoded device restores the documented default input on Darwin and ALSA platforms, and the strengthened synthetic ffmpeg harness verifies the selected input across success, HTTP rejection, and recorder failure without changing playback behavior. Exact-head checks are green.

PR #2595, opened 2026-09-04
fix(docs): await headers in Next.js webhook examples

HAYDEN-OAI:
## Summary

Await `headers()` in both README Next.js webhook handlers before passing the returned headers to SDK verification.

[Next.js documents `headers()` as asynchronous](https://nextjs.org/docs/app/api-reference/functions/headers). The current snippets pass its Promise directly to `unwrap()` or `verifySignature()`, so even a correctly signed webhook returns HTTP 400 with `Missing required header: webhook-signature`.

The documentation fix adds two `await` keywords. It also works with older synchronous `headers()` implementations. Raw-body handling, signature verification, event processing, and invalid-signature responses are unchanged; no SDK implementation or generated files are modified.

## Regression coverage

Extend the existing webhook documentation test to execute both actual README snippets with the real public SDK and HMAC verification. Eight cases cover asynchronous/synchronous headers and valid/invalid signatures for both methods.

Valid requests must return 200 and process the expected event. Invalid requests must return 400 with the typed signature error and no event processing. Fixtures use synthetic credentials and a signed whitespace-preserving payload; network requests are forbidden.

Four asynchronous cases fail before the README correction and pass afterward. The existing assertion that webhook verification itself is awaited is retained.

## Verification

- 39 focused webhook tests pass on Node 24.19.0 and Node 22.22.2.
- 16 independent checks of the actual fixed README snippets pass on exact Node 22.0.0 and Node 26.7.0, without any in-memory source correction.
- TypeScript `--noEmit`, changed-file lint, pinned Oxfmt 0.62.0, and `git diff --check` pass.
- Adversarial review of security, compatibility, raw-body handling, and regression coverage found no remaining blockers.
- Both changed files are absent from the current pinned Castiron snapshot `efa35b8b45ad152d9b5f9c327c48e35bca65c7ce`.

Local checks used cached Vitest 4.1.10, TypeScript 6.0.3, and Oxlint 1.76.0. The configured registry's missing publication-time metadata for `qs@6.16.0` still prevents a fresh pinned install; no dependency, lockfile, registry, or install-policy changes are included. CI runs the pinned repository toolchain.

### DIFF
--- a/README.md
+++ b/README.md
@@ -318,7 +318,7 @@ const client = new OpenAI({
 });
 
 export async function webhook(request: Request) {
-  const headersList = headers();
+  const headersList = await headers();
   const body = await request.text();
 
   try {
@@ -358,7 +358,7 @@ const client = new OpenAI({
 });
 
 export async function webhook(request: Request) {
-  const headersList = headers();
+  const headersList = await headers();
   const body = await request.text();
 
   try {
--- a/tests/lib/webhooks.test.ts
+++ b/tests/lib/webhooks.test.ts
@@ -1,6 +1,9 @@
 import { readFileSync } from 'node:fs';
+import { runInNewContext } from 'node:vm';
+import ts from 'typescript';
 import { vi } from 'vitest';
 import OpenAI, { InvalidWebhookSignatureError } from 'openai';
+import type { ClientOptions } from 'openai';
 
 const timestamp = 1_750_861_210;
 const webhookID = 'wh_685c059ae39c8190af8c71ed1022a24d';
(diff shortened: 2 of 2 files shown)


jbeckwith-oai (approved):
Validated the exact-head documentation fix. Both real Next.js webhook snippets now await the asynchronous headers helper while remaining compatible with synchronous values. The executable regression covers both verification methods, valid and invalid signatures, exact raw-body handling, no network access, and no event processing on failure. Review is complete; CI is still running.

PR #2600, opened 2026-09-04
fix(examples): reject unsuccessful background responses

HAYDEN-OAI:
## Summary

The background streaming example treats an initial `response.failed` or `response.incomplete` event as an interruption, unnecessarily retrieves the response again, and can finish with exit code 0. An unsuccessful resumed response is also printed as a successful result.

Add six lines to the example:

- Reject failed/incomplete events in the initial stream before the demonstration's sequence-10 cutoff can trigger a resume.
- Require the resumed `finalResponse()` status to be `completed` before printing its result.

The final-status check is necessary because `starting_after: 10` can filter an early terminal event from iteration while the accumulated response still records its failed/incomplete status. No cursor, replay, SDK, parsing, retry, or timeout behavior changes.

Only the handwritten example and its existing test harness change; both are absent from the pinned Castiron generated snapshot. Open PRs were checked and none covers this issue.

## Regression coverage and validation

- Reused the existing real-CLI/public-SDK subprocess and loopback-HTTP harness. Its four original completion/resumption scenarios now also cover failed and incomplete responses: before and at the initial cutoff, after explicit resumption, and after early EOF with a filtered terminal event.
- **8 new regressions failed before the fix; all 4 original success controls passed.**
- **49 focused tests passed on Node 22.22.2 and Node 24.19.0** via `./scripts/test tests/lib/background-streaming-example.test.ts tests/lib/ResponseStream.test.ts`.
- **36 independent actual CLI cases passed**—12 each on Node 22.0.0, 24.19.0, and 26.7.0. Verified exit codes, one POST versus POST+GET, resumption behavior, and successful output. No live API calls or real credentials were used.
- `tsc --noEmit`, pinned Oxfmt 0.62.0, available Oxlint 1.76.0, and `git diff --check` passed.

Local focused tests used cached Vitest 4.1.10/Oxlint 1.76.0 rather than the repository's 4.1.11/1.79.0 pins. Fresh frozen installation remains blocked by missing publication-time metadata for `qs@6.16.0` in the configured registry; no safeguards, dependencies, lockfiles, or configuration were changed. CI will verify the pinned toolchain.

## Adversarial self-review

Reviewed cutoff ordering, filtered replay events, terminal status, existing iterator abort/cleanup, API/transport error precedence, minimum-runtime compatibility, and missing regressions. Independent review of the actual diff found no actionable issue. New thrown errors contain only event type/status, and tests exclude synthetic private server details from stderr; the example's existing explicit event logging is unchanged.

### DIFF
--- a/examples/responses/stream_background.ts
+++ b/examples/responses/stream_background.ts
@@ -20,6 +20,9 @@ async function main() {
     }
 
     console.log('event', event);
+    if (event.type === 'response.failed' || event.type === 'response.incomplete') {
+      throw new Error(`Response ended with ${event.type}.`);
+    }
     if (event.type === 'response.completed') {
       completed = true;
     }
@@ -46,6 +49,9 @@ async function main() {
   }
 
   const result = await runner2.finalResponse();
+  if (result.status !== 'completed') {
+    throw new Error(`Response ended with status ${result.status}.`);
+  }
   console.log(result);
 }
 
--- a/tests/lib/background-streaming-example.test.ts
+++ b/tests/lib/background-streaming-example.test.ts
@@ -6,7 +6,10 @@ import path from 'node:path';
 import type { Response, ResponseStreamEvent } from 'openai/resources/responses/responses';
 import { expect, test } from 'vitest';
 
-function responseEvents(chunks: readonly string[]): ResponseStreamEvent[] {
+type TerminalStatus = 'completed' | 'failed' | 'incomplete';
+const privateErrorDetail = 'Synthetic private response error detail';
+
+function responseEvents(chunks: readonly string[], status: TerminalStatus): ResponseStreamEvent[] {
   const response: Response = {
     id: 'resp_background_example',
     object: 'response',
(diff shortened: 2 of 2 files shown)


jbeckwith-oai (approved):
Validated the exact-head background-response fix. Failed and incomplete terminal events now reject before the demonstration cutoff can trigger an unnecessary resume, and resumed final responses are required to be completed before output is treated as successful. The focused cases cover all four cutoff and EOF paths for completed, failed, and incomplete outcomes. CI is green.

tests/lib/background-streaming-example.test.ts
```diff
@@ -168,9 +177,16 @@ test.each(scenarios)('background example: $name', async ({ chunks, resumed, part
   try {
```
mccrearydamion69-crypto: Looks

examples/responses/stream_background.ts
```diff
@@ -46,6 +49,9 @@ async function main() {
   }
 
   const result = await runner2.finalResponse();
+  if (result.status !== 'completed') {
+    throw new Error(`Response ended with status ${result.status}.`);
+  }
   console.log(result);
 }
```
mccrearydamion69-crypto: Hi

PR #2601, opened 2026-09-04
test: stabilize Node CI subprocess fixtures

markstuart-oai:

## Changes being requested

Reduce repeated setup work in handwritten tests that intermittently exceed the Node 22 CI deadlines:

- Run the real `pnpm lint` → `pnpm format` → `pnpm lint` integration in a small temporary project containing the repository's actual aliases, scripts, and configs. Give the fixture its own offline pnpm state and link the installed tools. This avoids whole-repository scans and writes into other tests' temporary fixtures while preserving generated-import assertions and checking handwritten formatting.
- Select ts-node's existing SWC backend in the eight executable-example subprocess suites. Each child still runs the actual example against SDK source and synthetic local services, but avoids repeatedly compiling the SDK import graph with TypeScript. SWC is already a repository dependency.
- Disable npm's update notifier in the existing offline ecosystem fixtures. Main already contains their project-local offline configuration; preserve that configuration and its current smoke and credential coverage.

The existing 15-second child deadlines, 30-second test deadline, assertions, and worker modes remain. No production code, dependencies, lockfile, workflow policy, or generated files change. Main's newer LF and audio-input regression tests are preserved.

## Validation

- The equivalent test changes passed the full Node 22/24/26 suites: 7,329 handwritten tests and 556 generated tests per runtime, retaining the existing two generated skips; all 14 ecosystem projects and packaging/runtime-floor checks passed. Previously timing-out Node 22 WebSocket/audio cases completed in approximately two seconds under the same 15-second deadline.
- Independent review verified the public port preserves newer main-branch coverage and carries the same fixture and subprocess behavior. `git diff --check` passes.
- A fresh local frozen installation on the latest public base is blocked because the configured registry omits publication metadata for `qs@6.16.0`, also reported in [PR 2589](https://github.com/openai/openai-node/pull/2589). A fresh metadata cache gives the same missing-version result. The lockfile and dependency policies remain unchanged; this PR's CI will validate the current base.

### DIFF
--- a/tests/ecosystem-cli.test.ts
+++ b/tests/ecosystem-cli.test.ts
@@ -201,10 +201,11 @@ describe('ecosystem test CLI', () => {
       for (const name of projects) {
         const project = path.join(fixture, 'ecosystem-tests', name);
         mkdirSync(project, { recursive: true });
-        // pnpm does not forward npm_config_* variables to its worker scripts.
+        // pnpm workers strip npm_config_* from the environment. Keep local installs offline
+        // even with npm versions that audit linked packages.
         writeFileSync(
           path.join(project, '.npmrc'),
-          'audit=false\nfund=false\noffline=true\npackage-lock=false\n',
+          'audit=false\nfund=false\noffline=true\npackage-lock=false\nupdate-notifier=false\n',
         );
         writeFileSync(
           path.join(project, 'package.json'),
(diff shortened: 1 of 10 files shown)


HAYDEN-OAI (approved):
Reviewed the complete 10-file test-only diff. The isolated lint/format fixture still exercises the real package entrypoints and generated-import behavior without scanning neighboring fixtures, while the SWC substitutions preserve the executable-example coverage using the repository's existing backend. I found no substantive issues.

PR #2603, opened 2026-09-04
fix(examples): report nonstreaming request failures

HAYDEN-OAI:

## Problem

Three executable examples use `main().catch(console.error)`: raw responses, structured outputs, and Bedrock Responses. API, connection, response-body, and parsing failures are printed, but the process exits successfully. Scripts or CI invoking the examples therefore cannot distinguish those failures from a successful run.

## Changes

Use the existing audio-example pattern in each file: an `async run()` wrapper awaits `main()`, logs the original error, and sets `process.exitCode = 1` when it fails. Successful output and exit status remain unchanged. This does not force an early process exit, so output can drain normally.

Add one shared synthetic subprocess suite for these three examples. It executes the actual CLI files through the public SDK, covering HTTP rejection, malformed response JSON, success, and rejection of the raw-response example's second request. It checks exact request counts/routes/models, diagnostics, successful and partial output, exit codes, lack of signals, and credential non-disclosure.

The scope is three handwritten example entrypoints and one regression file. No SDK, authentication, transport, parsing, generated files, dependencies, or configuration changes.

## Validation

- **Before:** 7 intended regressions failed specifically with expected exit `1`, received `0`; all 3 success controls passed.
- **After:** all 10 subprocess cases pass. **209 related tests pass on Node 22.22.2 and 24.19.0.**
- **51 independent actual-CLI cases pass on the final source:** 17 each on exact minimum Node **22.0.0**, Node **24.19.0**, and forward-tested Node **26.7.0**. Each runtime covers three successes, thirteen corrected asynchronous failures, and an unchanged constructor-failure control. Extra cases include exhausted connection retries, structured JSON/Zod failures, HTTP authentication errors, and Bedrock's pre-dispatch invalid-header rejection.
- TypeScript 6.0.3 `--noEmit`, pinned Oxfmt 0.62.0, cached Oxlint 1.76.0, and `git diff --check` pass. The first inline-callback draft was replaced by the repository's existing async/await pattern to satisfy lint; final tests and independent review were repeated afterward.
- Local verification uses cached Vitest 4.1.10 and Oxlint 1.76.0 rather than the pinned 4.1.11/1.79.0; fresh frozen installation remains blocked by the configured registry's missing publication metadata for `qs@6.16.0`. No install safeguards were changed. Repository CI will validate the pinned toolchain.

All services and credentials are synthetic. Bedrock uses its documented endpoint/bearer-token configuration with the existing origin guards; child environments do not inherit credentials or preload hooks, AWS profile files are directed to the platform null device, and metadata lookup is disabled. No live OpenAI or AWS calls.

## Adversarial review

Independent review of the final wrappers and complete regression file found no actionable security, compatibility, cleanup, output-flushing, or coverage gaps. Error diagnostics, natural output draining, partial raw-response output, successful exits, and failures occurring before `main()` are preserved.

### DIFF
--- a/examples/bedrock/responses.ts
+++ b/examples/bedrock/responses.ts
@@ -21,4 +21,13 @@ async function main() {
   console.log(response.output_text);
 }
 
-main().catch(console.error);
+async function run() {
+  try {
+    await main();
+  } catch (error) {
+    console.error(error);
+    process.exitCode = 1;
+  }
+}
+
+void run();
--- a/examples/client/raw-response.ts
+++ b/examples/client/raw-response.ts
@@ -31,4 +31,13 @@ async function main() {
   }
 }
 
-main().catch(console.error);
+async function run() {
+  try {
+    await main();
+  } catch (error) {
+    console.error(error);
+    process.exitCode = 1;
+  }
+}
+
+void run();
--- a/examples/responses/structured-outputs.ts
+++ b/examples/responses/structured-outputs.ts
@@ -29,4 +29,13 @@ async function main() {
   console.log('answer:', rsp.output_parsed?.final_answer);
 }
 
-main().catch(console.error);
+async function run() {
+  try {
+    await main();
+  } catch (error) {
+    console.error(error);
+    process.exitCode = 1;
+  }
+}
+
+void run();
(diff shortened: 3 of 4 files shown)


jbeckwith-oai (approved):
Validated the exact-head example exit-status fix. Each nonstreaming example now preserves the original diagnostic and natural output draining while setting a failing process status for asynchronous request and parsing errors. The subprocess suite covers success, HTTP failure, malformed JSON, partial raw-response output, request boundaries, and credential non-disclosure. Review is complete; CI is still running.

PR #2607, opened 2026-09-04
fix: preserve chat runner abort reasons

### LINKED ISSUE #2606: Non-streaming chat runners drop caller abort reasons
## Summary

The chat completion convenience runners lose the reason supplied to a caller `AbortSignal` on the non-streaming path.

A direct SDK request preserves `signal.reason` as `APIUserAbortError.cause`, and the streaming runner path preserves it as well. `runTools(..., { signal })` with `stream: false` instead bridges the caller signal through `EventStream._listenForAbort()`, which aborts the runner's internal controller without forwarding the reason.

## Reproduction

On current `main` at `1662d1e`, extend the existing `flow with abort` regression in `tests/lib/ChatCompletionRunFunctions.test.ts`:

```ts
const controller = new AbortController();
const reason = new Error('stop after assistant message');

const runner = openai.chat.completions.runTools(params, {
  signal: controller.signal,
});

runner.on('message', (message) => {
  if (message.role === 'assistant') controller.abort(reason);
});

const error = await runner.done().catch((error) => error);
expect(error).toMatchObject({ cause: reason });
```

Current result: the non-streaming case fails because the final `APIUserAbortError` has no caller reason. The corresponding streaming `flow with abort` case passes the same assertion.

## Root cause

`EventStream._listenForAbort()` currently does:

```ts
if (signal.aborted) {
  this.controller.abort();
  return;
}

const listener = () => this.controller.abort();
```

The internal controller therefore receives a default abort reason rather than `signal.reason`. The request layer can only preserve the reason it receives from that internal signal.

## Expected behavior

Forward the caller's abort reason when bridging into the runner-owned controller, so convenience runners retain the same `APIUserAbortError.cause` semantics as direct requests. This should apply both when the signal is already aborted and when it aborts later.

sylvesterkaczmarek:

## Changes being requested

Preserve caller-provided abort reasons across chat completion convenience runners.

The direct request layer already attaches `signal.reason` to `APIUserAbortError.cause`, but chat completion runners could lose or misclassify that reason at internal cancellation boundaries. This change:

- forwards `signal.reason` when `EventStream` bridges an external signal into its runner-owned controller;
- attaches the runner controller's reason to abort errors raised at tool-callback cancellation checkpoints;
- normalizes a sequential tool callback that throws the aborted runner signal's exact reason into `APIUserAbortError`, preserving that reason as `cause`;
- covers both later aborts and already-aborted source signals;
- verifies parity with the existing streaming runner behavior.

Fixes #2606.

## Review follow-up

The current head addresses the sequential-callback classification issue raised in review. Both parsed and unparsed tool callback paths now catch a rejection equal to `runner.controller.signal.reason` after cancellation and translate it through the same `userAbortError()` helper. The regression exercises `parallel_tool_calls: false`, cancels with a custom `Error`, has the active callback call `throwIfAborted()`, and verifies `runner.done()` rejects with `APIUserAbortError`, preserves the custom cause, emits the abort path, and leaves `runner.aborted === true`.

Fresh code/security review is running on commit `69e2710cce37947e2f51476d8b57ab3f05c836fb`.

## Validation

Previous-head validation:

- `pnpm test tests/lib/ChatCompletionRunFunctions.test.ts tests/lib/runner-abort-listeners.test.ts` — 48 passed
- `pnpm exec tsc --noEmit` — passed
- repository-wide Ultracite formatting/lint check — passed, 0 warnings/errors
- `pnpm build` — passed
- `git diff --check` — passed

Pushed-head CI/review is required for the new regression.

## Additional context & links

The change does not alter cancellation timing or add a new API; it preserves and consistently classifies the same abort-cause information already exposed by direct requests.

### DIFF
--- a/src/lib/EventStream.ts
+++ b/src/lib/EventStream.ts
@@ -1444,19 +1444,37 @@ export class EventStream<EventTypes extends BaseEvents> {
     this.controller.abort();
   }
 
+  protected _userAbortError(): APIUserAbortError {
+    const error = new APIUserAbortError();
+    Object.defineProperty(error, 'cause', {
+      value: this.controller.signal.reason,
+      writable: true,
+      configurable: true,
+    });
+    return error;
+  }
+
+  #abortFromSignal(signal: AbortSignal) {
+    try {
+      this.controller.abort(signal.reason);
+    } catch {
+      this.controller.abort();
+    }
+  }
+
   protected _listenForAbort(signal: AbortSignal | null | undefined) {
     if (!signal || this.ended) {
       return;
     }
     if (signal.aborted) {
-      this.controller.abort(signal.reason);
+      this.#abortFromSignal(signal);
       return;
     }
     if (this.#abortListeners.some((registration) => registration.signal === signal)) {
       return;
     }
 
-    const listener = () => this.controller.abort(signal.reason);
+    const listener = () => this.#abortFromSignal(signal);
     signal.addEventListener('abort', listener, { once: true });
     this.#abortListeners.push({ signal, listener });
   }
@@ -1897,7 +1915,7 @@ export class EventStream<EventTypes extends BaseEvents> {
   #handleError(this: EventStream<EventTypes>, error: unknown) {
     this.#errored = true;
     if (error instanceof Error && error.name === 'AbortError') {
-      error = new APIUserAbortError();
+      error = this._userAbortError();
     }
     if (error instanceof APIUserAbortError) {
       this.#aborted = true;
--- a/src/lib/EventStream.ts
+++ b/src/lib/EventStream.ts
@@ -1449,14 +1449,14 @@ export class EventStream<EventTypes extends BaseEvents> {
       return;
     }
     if (signal.aborted) {
-      this.controller.abort();
+      this.controller.abort(signal.reason);
       return;
     }
     if (this.#abortListeners.some((registration) => registration.signal === signal)) {
       return;
     }
 
-    const listener = () => this.controller.abort();
+    const listener = () => this.controller.abort(signal.reason);
     signal.addEventListener('abort', listener, { once: true });
     this.#abortListeners.push({ signal, listener });
   }
--- a/src/lib/AbstractChatCompletionRunner.ts
+++ b/src/lib/AbstractChatCompletionRunner.ts
@@ -443,15 +443,7 @@ export class AbstractChatCompletionRunner<
     };
 
     let allowBufferedToolCall = false;
-    const userAbortError = () => {
-      const error = new APIUserAbortError();
-      Object.defineProperty(error, 'cause', {
-        value: this.controller.signal.reason,
-        writable: true,
-        configurable: true,
-      });
-      return error;
-    };
+    const userAbortError = () => this._userAbortError();
 
     const runToolCall = async (toolCall: ChatCompletionMessageToolCall): Promise<ToolCallResult> => {
       const bufferedToolCall = allowBufferedToolCall;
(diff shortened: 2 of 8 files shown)


sylvesterkaczmarek (commented):
Reason propagation is now consistent across both cancellation boundaries: external signals transfer their original reason into the runner controller, and tool checkpoints wrap that same reason in `APIUserAbortError.cause`. Covering already-aborted signals is important because that path bypasses listener delivery entirely.

src/lib/EventStream.ts
```diff
@@ -1449,14 +1449,14 @@ export class EventStream<EventTypes extends BaseEvents> {
...
-      this.controller.abort();
+      this.controller.abort(signal.reason);
       return;
     }
     if (this.#abortListeners.some((registration) => registration.signal === signal)) {
       return;
     }
 
-    const listener = () => this.controller.abort();
+    const listener = () => this.controller.abort(signal.reason);
```
HAYDEN-OAI: **[P2] Preserve cancellation classification for tool callbacks**

With `parallel_tool_calls: false` or a named `tool_choice`, a tool callback that calls `runner.controller.signal.throwIfAborted()` after the caller cancels with `abort(new Error('stop'))` now throws that custom error. The callback rejection propagates through `await runToolCall(...)`, bypassing the new abort checkpoint, and `EventStream.#handleError` classifies it as an ordinary error because its name is not `AbortError`. Consequently, `runner.done()` rejects with `OpenAIError`, the runner emits `error` instead of `abort`, and `runner.aborted` remains `false`. Before this change, the runner controller's default `AbortError` was correctly converted into `APIUserAbortError`. Please normalize rejections matching the aborted runner signal's reason into `APIUserAbortError`, preserving `cause`.

HAYDEN-OAI (changes requested):
Requesting changes for a cancellation classification regression in sequential tool callbacks. Forwarding the caller's reason also requires normalizing callback rejections that match that reason into APIUserAbortError, preserving cause.

sylvesterkaczmarek:
Addressed the requested cancellation-classification case. Sequential tool callbacks now normalize a rejection that is exactly the aborted runner signal reason into APIUserAbortError while preserving that reason as cause. Added a focused parallel_tool_calls:false regression where the callback observes the caller abort via throwIfAborted(); it now emits the abort path and the focused test passes.

sylvesterkaczmarek:
Addressed the new cancellation findings. Abort propagation now tolerates structural signals whose `reason` getter throws, stream helpers preserve the runner controller reason in `APIUserAbortError.cause`, and sequential callback matching uses `Object.is` so values such as `NaN` retain user-abort classification. Added regressions for the structural signal, `NaN`, and Responses helper cause propagation. Focused validation: 117 tests passed across `runner-abort-listeners`, `ChatCompletionRunFunctions`, and `streaming-core`.

HAYDEN-OAI (approved):
Approved at 4d333bd5. All four review threads are resolved, the branch includes current main without conflicts, and independent review found no remaining actionable correctness issue.

Validation: 675 relevant tests pass on Node 22 and 24; lint, typecheck, build, and CJS/ESM pre-abort and mid-body cancellation smoke checks pass on Node 22.0.0 and 24. Coverage includes parsed/unparsed and named tool callbacks, NaN reasons, all three streaming helpers, and throwing structural-signal getters with cleanup.

GitHub CI is currently waiting for approval of the fork workflow runs; this review does not indicate those checks have passed.

sylvesterkaczmarek:
Thanks for the review and merge. I appreciate the careful pass over the abort-reason propagation and cancellation cases.

PR #2611, opened 2026-09-04
feat(api): correct function argument completion event fields (openapi-545)

markstuart-oai:
## Summary

Removes the name interface field from response.function_call_arguments.done to match the API response.

## Changes

- Remove name from stable and beta Responses function argument completion interfaces.
- Update code that accesses the removed property to read name from the function-call item matched by item_id.


sylvesterkaczmarek (commented):
The generated TypeScript surface now matches the wire contract: the completion event carries the finalized arguments plus `item_id`, while function identity remains on the corresponding function-call item. Removing `name` from both stable and beta avoids promising consumers a field the server does not send, and the OpenAPI/interface edits move together. No blocker from this SDK projection.

HAYDEN-OAI (approved):
Reviewed all four changed files and reconciled the complete schema and generated-interface diff. Stable and beta Responses completion events consistently remove name, including their inherited WebSocket types; function-call item names and Realtime contracts remain unchanged. The stream accumulator and parser use arguments and function-call items without relying on this removed event field. No actionable correctness findings. Validation was static inspection and exact-file/diff checks; project tests and consumer compilation were not run.

PR #2612, opened 2026-09-04
fix(examples): report unsuccessful Azure Realtime responses

HAYDEN-OAI:
## Summary

Make the Azure Realtime `ws` and native `websocket` examples report a nonzero exit status when `response.done` contains a failed, cancelled, or incomplete response. Previously both examples closed and exited 0 for these terminal failures.

Copy the existing canonical Realtime examples' terminal-status guard: emit a fixed diagnostic, set `process.exitCode = 1` for non-completed responses, and keep the existing orderly close for every terminal status. No SDK, authentication, URL, transport, dependency, or generated-code changes.

Extend the existing real TLS/WebSocket subprocess regression table to include both Azure examples. Only Azure credential acquisition is stubbed with a synthetic token; the actual examples, public SDK, and socket transports run. The child environment is isolated from developer `.env` files.

## Verification

- Before the fix, six new Azure failure cases failed and ten controls passed on exact Node 22.0.0 and Node 24.19.0.
- After the fix, all 16 real TLS/WebSocket cases passed independently on exact Node 22.0.0, 24.19.0, and 26.7.0 (48 checks), plus Node 22.22.2.
- An independent actual-example harness reproduced 18 incorrect successes across those three runtimes before the fix; all 24 success/failure cases passed afterward.
- Final example/Realtime privacy suites: 68/68 tests passed. Tests verify expected Azure route and bearer header, outbound messages, streamed text, socket closure, exit codes, and absence of synthetic credentials or private response metadata in output.
- Full TypeScript check, pinned Oxfmt 0.62.0, cached Oxlint 1.76.0, and whitespace checks pass. The local cache uses Vitest 4.1.10 and Oxlint 1.76.0 rather than pinned 4.1.11/1.79.0; pinned CI remains authoritative.

## Adversarial review

Independent reviews checked failure-status handling, successful exit compatibility, diagnostic privacy, credential/endpoint scope, subprocess isolation, and socket cleanup. Test isolation was tightened to prevent developer `.env` files from supplying credentials. The handlers match the canonical examples exactly; no blocking findings remain.

### DIFF
--- a/examples/azure/realtime/ws.ts
+++ b/examples/azure/realtime/ws.ts
@@ -52,7 +52,14 @@ async function main() {
   rt.on('response.output_text.delta', (event) => process.stdout.write(event.delta));
   rt.on('response.output_text.done', () => console.log());
 
-  rt.on('response.done', () => rt.close());
+  // response.done also covers failed, cancelled, and incomplete responses.
+  rt.on('response.done', (event) => {
+    if (event.response.status !== 'completed') {
+      console.error('Response did not complete successfully.');
+      process.exitCode = 1;
+    }
+    rt.close();
+  });
 
   rt.socket.on('close', () => console.log('\nConnection closed!'));
 }
--- a/examples/azure/realtime/websocket.ts
+++ b/examples/azure/realtime/websocket.ts
@@ -52,7 +52,14 @@ async function main() {
   rt.on('response.output_text.delta', (event) => process.stdout.write(event.delta));
   rt.on('response.output_text.done', () => console.log());
 
-  rt.on('response.done', () => rt.close());
+  // response.done also covers failed, cancelled, and incomplete responses.
+  rt.on('response.done', (event) => {
+    if (event.response.status !== 'completed') {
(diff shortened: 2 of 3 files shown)


jbeckwith-oai (approved):
Reviewed exact head 890b2660680b3010266a4d518afdc514a241e4bb. The Azure ws and native-websocket examples now match the canonical Realtime terminal-status behavior: completed responses retain exit 0, failed/cancelled/incomplete responses emit the fixed non-sensitive diagnostic and set exit 1, and every terminal status still closes the socket. The expanded subprocess matrix runs the actual examples and transports, isolates dotenv/credentials, verifies the Azure route and bearer header, and checks output privacy. Focused Node 24 validation passed all 16 cases; exact-head CI has no failures observed.

examples/azure/realtime/ws.ts
mccrearydamion69-crypto, line 2: Hi

mccrearydamion69-crypto (commented):
I hope life gets better sooner than later

PR #2613, opened 2026-09-04
fix(streaming): abort before early-exit cleanup

HAYDEN-OAI:
## Summary

Fix a response-backed SSE iterator lifecycle hang: after receiving an event, breaking from `for await` or returning the iterator could wait indefinitely for asynchronous body cancellation before the request controller was aborted.

Abort before the locally owned SSE-message iterator begins nested cleanup. This reuses the existing abort-aware cleanup and completion-sentinel state; normal EOF and `[DONE]`, unused/completed iterators, original errors, and tee sibling ownership are preserved.

This is a nine-line production change plus focused public-API regressions. No parser, buffering, timeout, generated code, public declaration, dependency, or retry changes. Native async-generator return queued behind an unresolved next remains outside this fix.

## Review follow-up (`626aa57b`)

Moved the nine public Responses early-exit lifecycle cases and their setup into `tests/lib/streaming-early-exit.test.ts`, addressing the maintainability review. All executable test code and helper declarations are unchanged. `tests/lib/streaming-core.test.ts` is byte-identical to the PR base (961 lines), so it is no longer part of the overall PR diff. Production code is unchanged from the initial implementation.

CI identified that the deliberate one-event `for await` loop had moved outside an existing filename-scoped lint exception. A single documented next-line `no-unreachable-loop` exception preserves that regression without changing the loop or broadening lint policy. The initial local check used the pinned linter binary with an older shared Ultracite preset; final validation below uses a clean, frozen installation of the complete pinned toolchain.

- The relocated suite retained its original before-fix detection on base `31bf7e79`: four expected deferred-cancellation failures and five passing controls.
- Final full canonical handwritten unit run: **7,540/7,540 tests passed across 191 files**, using Node 24.19.0 and pinned Vitest 4.1.11. The earlier cached-formatter failure is absent with the complete pinned installation.
- The combined core and early-exit suites passed all **75 tests** (66 core + 9 early-exit) on Node 22.22.3, 24.19.0, and 26.7.0 with Vitest 4.1.11.
- Full canonical `./scripts/lint`, TypeScript 6.0.3 checking, and whitespace checks passed with Ultracite 7.10.5, Oxfmt 0.62.0, and Oxlint 1.79.0. Normal `pnpm 11.5.1 install --frozen-lockfile` retained all dependency and supply-chain policies; no dependency or configuration files changed.
- Repeated adversarial self-review and independent equivalence reviews confirmed unchanged assertions, error identity, cancellation ordering, listener cleanup, imports, and test discovery. Both test paths are wholly handwritten.

The updated head is also subject to the repository's full CI. The implementation-level results below were collected before this test-only extraction.

## Initial implementation verification (`194a49a6`)

- Four new public `responses.create({ stream: true })` regressions failed on clean main before the fix; all nine new lifecycle cases now pass.
- 177 focused streaming tests pass on Node 22.22.2 and 24.19.0, including tee, SSE sentinel/error/privacy, and Bedrock Runtime coverage. The original combined streaming-core suite passed 75/75.
- 150 independent built-package checks pass across CJS/ESM and exact Node 22.0.0, 24.19.0, and 26.7.0. They cover public Responses/chat, ordinary/thread/synthesized SSE, pending/rejecting/throwing cleanup, error identity, locks/listeners, ordinary/nested tees, and real native-fetch loopback disconnects. All 30 baseline hang cases were reproduced before rebuilding.
- Canonical build, full TypeScript check, pinned Oxfmt 0.62.0, cached Oxlint 1.76.0, and whitespace checks pass. Root and core-streaming CJS/ESM declarations are byte-identical to main. All 24 streaming benchmarks completed successfully; no performance comparison is claimed.
- Final complete handwritten unit run: **7,539 passed / 7,540 total**. The sole failure also reproduces on untouched main: the cached formatter fixture leaves an assignment unformatted. Local dependencies contain Vitest 4.1.10 rather than pinned 4.1.11, and Oxlint 1.76.0 rather than pinned 1.79.0; pinned CI remains authoritative.

## Adversarial review

Independent reviews checked security, compatibility, cancellation/error ordering, `[DONE]`, unstarted/unused/completed iterators, cleanup rejection observation, and tee ownership. Test cleanup was tightened to guarantee listener removal on failures and assert exactly-once cancellation after delayed cleanup settles. No blocking findings remain.

### DIFF
--- a/tests/lib/streaming-core.test.ts
+++ b/tests/lib/streaming-core.test.ts
@@ -1,3 +1,4 @@
+import { setImmediate as nextTurn } from 'node:timers/promises';
 import { vi } from 'vitest';
 import OpenAI from 'openai';
 import { APIError, APIUserAbortError, OpenAIError } from 'openai/core/error';
@@ -7,9 +8,27 @@ import { ReadableStreamFrom } from 'openai/internal/shims';
 import type { ResponseTextDeltaEvent } from 'openai/resources/responses/responses';
 
 const encoder = new TextEncoder();
+const responseEvent: ResponseTextDeltaEvent = {
+  type: 'response.output_text.delta',
+  sequence_number: 0,
+  item_id: 'msg_test',
+  output_index: 0,
+  content_index: 0,
+  delta: 'hello',
+  logprobs: [],
+};
 
 function responseForSSE(value: string): Response {
   return new Response(ReadableStreamFrom([encoder.encode(value)]));
+}
+
+async function publicSSEStream(body: ReadableStream<Uint8Array>) {
+  const client = new OpenAI({
+    apiKey: 'test-key',
+    maxRetries: 0,
+    fetch: async () => new Response(body, { headers: { 'content-type': 'text/event-stream' } }),
+  });
+  return await client.responses.create({ model: 'gpt-4o', input: 'hello', stream: true });
 }
 
 async function collect<T>(iterable: AsyncIterable<T>): Promise<T[]> {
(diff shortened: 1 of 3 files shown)


tests/lib/streaming-core.test.ts
```diff
@@ -237,6 +256,148 @@ describe('Stream.fromSSEResponse', () => {
   });
 });
 
+describe('public Responses SSE iterator early exit', () => {
```
markstuart-oai: [P2] Extract the public early-exit lifecycle suite before expanding this file past 1k lines. This PR grows streaming-core.test.ts from 961 to 1,113 lines, adding a distinct 142-line suite to a file that already covers the Stream APIs and SSE decoder. The new public Responses lifecycle cases and their publicSSEStream setup have a clear separate responsibility: move them into a focused test file such as streaming-early-exit.test.ts, preserving the regression coverage. That keeps this fix small without further expanding the existing multipurpose suite.

markstuart-oai (commented):
The early-exit fix reproduces against fresh main and resolves the hang. I found no functional regression, but the new test suite should be decomposed as noted inline before approval.

Validation at `194a49a6bf58938c3007f18140b6722d5a0bce74`: 166 focused streaming tests passed on both Node 22.23.2 and 24.19.0. All four new deferred-cancellation cases failed against fresh main and pass with this change. The canonical build, eight built-package CJS/ESM public Responses checks, changed-file Oxfmt/Oxlint checks, and diff whitespace checks passed. Full type-checking was limited by unavailable example dependencies (including Azure Identity, dotenv, and Express); the SDK build type-checks passed. I did not rerun the complete unit suite, ecosystem matrix, or real-network tests.

tests/lib/streaming-core.test.ts
HAYDEN-OAI: Confirmed on head `194a49a6bf58938c3007f18140b6722d5a0bce74`. This change adds a distinct 142-line public Responses early-exit lifecycle suite to the existing 961-line multipurpose streaming core test file. The behavior coverage is valuable and separable from the Stream/SSE decoder cases. Valid; extract it into a focused lifecycle test module with shared setup while preserving the assertions. Keeping unresolved until a fix is pushed.

HAYDEN-OAI: Addressed in 3e7a4f72383376600b4489666988fe52aaf78e6a. The nine public Responses early-exit lifecycle cases and their setup now live in `tests/lib/streaming-early-exit.test.ts`; `streaming-core.test.ts` is restored byte-for-byte to the 961-line PR base and disappears from the overall PR diff. Production code is unchanged.

Exact declaration-text comparisons and independent review confirm all assertions and cleanup are preserved. The moved suite still produces the same four deferred-cancellation failures and five passing controls on the old production code. Updated validation passes 92 related tests on Node 24, plus all 75 combined core/early-exit tests on Node 22 and 26, TypeScript checking, and the pinned formatter/linter. Fresh CI is running for the signed follow-up. Leaving the thread for your verification.

HAYDEN-OAI: Follow-up 626aa57bd46c7333e90191e2b571868d608aefe6 adds the narrowly documented lint exception needed by the deliberate one-event `for await` loop after moving outside the old file's exception. No executable test code or production code changed.

CI exposed a gap in the earlier local lint validation: the pinned binary was loading an older shared Ultracite preset. I repeated validation with a clean frozen installation of all pinned tools. Full canonical lint, TypeScript, all 7,540 handwritten tests, and the 75 focused cases across Node 22/24/26 now pass. The signed correction is pushed and fresh CI is running; the PR description now distinguishes this verified toolchain from the historical local results.

markstuart-oai (approved):
Re-reviewed the complete change at 626aa57b and the test reorganization since my prior review. The earlier decomposition finding is addressed: public Responses early-exit coverage now lives in a focused 176-line suite, and streaming-core.test.ts is back below 1,000 lines. The production change is unchanged from the previously inspected version; abort-before-cleanup ordering, unused-iterator ownership, completion handling, and error identity remain covered. No substantive correctness or maintainability issues remain.

Validation: 110 focused tests passed on Node 24.19.0 through the canonical scripts/test entrypoint, including the extracted early-exit suite, core streaming tests, and line decoder tests. Verified production source and dependency files are unchanged since my prior review; git diff --check passed. I did not repeat the full package build, complete test suite, or browser/runtime matrix.

PR #2614, opened 2026-09-04
docs(streaming): reject unsuccessful background responses

HAYDEN-OAI:
## Summary

Fix the background-resume snippet in the streaming guide so unsuccessful responses are not treated as successful output.

- Stop on initial `response.failed` or `response.incomplete` events before attempting a resume.
- Require a completed final snapshot before printing its final output, including when a resumed stream ends cleanly while still queued or in progress.
- Use fixed diagnostics that do not include private response error details.

This brings the guide in line with the executable example fixed in #2600. It does not change SDK behavior: `finalResponse()` intentionally returns the accumulated snapshot. Only the handwritten guide and its existing regression test change.

## Verification

- Extended the existing HTTP/subprocess suite to execute the guide's exact TypeScript code block through the public SDK, alongside the executable example; no replacement snippet or mock helper is used.
- Ten new guide failure cases failed before the fix; 16 existing/completed controls passed. All 26 scenarios now pass.
- Verified the suite on exact Node 22.0.0, 24.19.0, and 26.7.0. Final queued-response fixture tightening was rechecked on the minimum/forward runtimes and with the full suite on Node 24.
- Covered initial and resumed completion/failure, completion at the interruption cutoff, clean EOF, queued/in-progress final snapshots, expected POST/GET counts, and diagnostic privacy.
- Full TypeScript check, pinned Oxfmt 0.62.0, cached Oxlint 1.76.0, and whitespace checks pass. Local Vitest/Oxlint are 4.1.10/1.76.0 rather than pinned 4.1.11/1.79.0; pinned CI remains authoritative.

## Adversarial review

Independent reviews checked terminal-event ordering, resume/cursor compatibility, clean EOF, diagnostic privacy, exact-snippet execution, child-process isolation, and cleanup. The queued fixture was tightened to emit only `response.created` and preserve its ID through both requests. Final review found no blocking issues. No SDK, generated code, dependency, or public API changes.

### DIFF
--- a/docs/streaming.md
+++ b/docs/streaming.md
@@ -142,6 +142,10 @@ for await (const event of initial) {
     responseId = event.response.id;
   }
 
+  if (event.type === 'response.failed' || event.type === 'response.incomplete') {
+    throw new Error('The background response did not complete successfully.');
+  }
+
   if (event.type === 'response.completed') {
     completed = true;
   }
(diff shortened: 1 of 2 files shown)


jbeckwith-oai (approved):
Reviewed the exact head thoroughly. The guide now distinguishes terminal failures from resumable interruptions, validates the final snapshot before using output, and exercises the exact markdown snippet across completed, failed, incomplete, queued, in-progress, interrupted, and clean-EOF paths. Focused verification passed on Node 24: 26/26 tests in tests/lib/background-streaming-example.test.ts. No blocking findings.

PR #2615, opened 2026-09-04
fix(embeddings): reject incomplete float32 vectors

HAYDEN-OAI:
## Summary

Reject base64 embedding responses whose decoded byte length is not a multiple of four. The Node decoder currently passes a fractional element count to `Float32Array`, which silently drops trailing bytes: one to three bytes become an empty vector, and five to seven bytes become a one-element vector. The browser decoder already rejects these inputs with `RangeError`.

This adds a three-line guard at the existing conversion boundary and focused regressions in two existing handwritten test files. Empty/aligned vectors, pooled-buffer offsets, explicit base64/float requests, and the current decoding fast path are preserved. The fixed diagnostic includes no response contents. No grammar parser, size limit, buffer copy, dependency, generated-code, or public API changes.

## Verification

- Before the fix, 12 new Node-helper/public-client cases failed; browser rejection and all 29 controls passed. All 57 focused helper, embeddings, and request-compatibility tests now pass on Node 22.22.2 and 24.19.0.
- Independent built CJS/ESM checks passed all 84 scenarios on exact Node 22.0.0, 24.19.0, and 26.7.0, including 156 real loopback HTTP requests. Coverage includes malformed lengths 1/2/3/5/6/7/8193, empty and aligned vectors beyond the buffer-pool threshold, isolated browser decoding, explicit encodings, numeric body overrides, and `asResponse()`/`withResponse()` behavior.
- Canonical CJS/ESM build, full TypeScript check, pinned Oxfmt 0.62.0 and Oxlint 1.79.0, and whitespace checks pass. All emitted declaration files are byte-for-byte unchanged.
- Full local handwritten suite: 7,581/7,583 passed. The mock-launcher timeout passed on its isolated rerun (all four launcher tests); the formatter-entrypoint fixture fails identically on clean main with the existing local dependency cache. Local Vitest is 4.1.10 rather than pinned 4.1.11; CI will exercise the pinned environment. Neither unrelated fixture nor dependency configuration was changed.

## Adversarial review

Three independent reviews checked malformed and large inputs, diagnostic privacy, Buffer/atob parity, pooled-buffer ownership, response-accessor behavior, explicit encodings, declarations, and regression coverage. Final review found no blocking issues. All three changed files are absent from the verified pure-generated baseline.

### DIFF
--- a/src/internal/utils/base64.ts
+++ b/src/internal/utils/base64.ts
@@ -46,6 +46,9 @@ export const toFloat32Array = (base64Str: string): Array<number> => {
   if (typeof Buffer !== 'undefined') {
     // for Node.js environment
     const buf = Buffer.from(base64Str, 'base64');
+    if (buf.length % Float32Array.BYTES_PER_ELEMENT !== 0) {
+      throw new RangeError('Invalid base64-encoded float32 array: byte length must be a multiple of 4');
+    }
     return Array.from(
       new Float32Array(buf.buffer, buf.byteOffset, buf.length / Float32Array.BYTES_PER_ELEMENT),
     );
--- a/tests/base64.test.ts
+++ b/tests/base64.test.ts
@@ -1,4 +1,14 @@
-import { fromBase64, toBase64 } from 'openai/internal/utils/base64';
+import { fromBase64, toBase64, toFloat32Array } from 'openai/internal/utils/base64';
+
+const incompleteVectors = [1, 2, 3, 5, 6, 7].map((byteLength) => ({
+  byteLength,
+  encoded: Buffer.alloc(byteLength).toString('base64'),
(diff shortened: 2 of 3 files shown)


jbeckwith-oai (approved):
Reviewed the exact head against the stated malformed-float32-vector scope. The guard correctly rejects decoded byte lengths not divisible by Float32Array.BYTES_PER_ELEMENT before constructing the typed view, preserves empty/aligned values and pooled-buffer offsets, and keeps explicit base64 responses unchanged. The fixed diagnostic does not expose response content. Focused verification passed on Node 24: 41/41 tests across tests/base64.test.ts and tests/lib/embeddings.test.ts; diff check clean. No blocking findings.

PR #2616, opened 2026-09-04
fix(examples): report unsuccessful fine-tuning jobs

HAYDEN-OAI:
## Summary

Make the fine-tuning example exit unsuccessfully when a job finishes as `failed` or `cancelled`. Its polling loop already stops for both states, but the script then returns normally and reports exit code 0.

Add a four-line check after the existing loop: only `succeeded` is successful. Other terminal outcomes raise a fixed diagnostic through the existing `main().catch(...)` handler and exit with code 1. The message does not include `fineTune.error` or training data.

Polling states, intervals, event reporting, file handling, model parameters, and successful behavior are unchanged. Only the handwritten example and its existing regression suite change; there are no SDK, generated-code, dependency, or public API changes.

## Verification

- Updated the existing actual-example/public-SDK harness rather than introducing a replacement polling loop. Eight failure cases failed before the fix, while seven successful/file-processing controls passed; all 15 scenarios now pass on Node 22.22.2 and 24.19.0.
- Cover initially failed/cancelled jobs and transitions from each active state (`validating_files`, `queued`, `running`), retaining exact request, event, and timer assertions.
- Independent real-subprocess checks passed all 18 scenarios on exact Node 22.0.0, 24.19.0, and 26.7.0. Initial and queued terminal states preserve their three/five requests and the real five-second polling wait. Before the fix, all 12 failed/cancelled cases incorrectly exited 0; succeeded controls remain unchanged.
- All uploads and job requests were synthetic and loopback-only. Both output channels were checked for private error details; no live fine-tuning job was created.
- Normal full-project `tsc --noEmit`, pinned Oxfmt 0.62.0 and Oxlint 1.79.0, and whitespace checks pass. The focused suite uses cached Vitest 4.1.10; CI will run the pinned 4.1.11 environment.

## Adversarial review

Three independent reviews checked the complete status contract, initial and polled terminal outcomes, existing file failures, request counts, timing, privacy, and compatibility. Review tightened the privacy assertion to cover stdout and stderr and kept indexed test access compatible with the repository's ES2020 type library. The final diff has no blocking findings and introduces no new polling state or helper infrastructure.

### DIFF
--- a/examples/fine-tuning/fine-tuning.ts
+++ b/examples/fine-tuning/fine-tuning.ts
@@ -76,6 +76,10 @@ async function main() {
 
     await new Promise((resolve) => setTimeout(resolve, 5000));
   }
+
+  if (fineTune.status !== 'succeeded') {
+    throw new Error('Fine-tuning job did not complete successfully.');
+  }
 }
 
 main().catch((err) => {
--- a/tests/lib/fine-tuning-example.test.ts
+++ b/tests/lib/fine-tuning-example.test.ts
@@ -1,4 +1,5 @@
 import fs from 'node:fs';
+import { inspect } from 'node:util';
 import { runInNewContext } from 'node:vm';
 import ts from 'typescript';
 import { vi } from 'vitest';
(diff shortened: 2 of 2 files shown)


jbeckwith-oai (approved):
Reviewed the exact head against the stated fine-tuning example scope. All terminal job states are covered: only succeeded returns normally, while initially or subsequently failed/cancelled jobs flow through the existing catch handler with exit 1 and a fixed privacy-safe diagnostic. Polling states, request/event behavior, waits, file handling, and the success path remain unchanged. Focused Node 24 verification passed 15/15 tests; diff check clean. No blocking findings.

PR #2617, opened 2026-09-04
fix(examples): reject incomplete image streams

HAYDEN-OAI:
## Summary

Make the image-streaming example reject clean EOF when it has not received and saved a final image. An empty stream or a stream containing only partial images currently exits with code 0 despite producing no final image.

Add one local completion flag, set only after the `image_generation.completed` event's file write succeeds, and check it after iteration. Incomplete streams now use the existing catch handler and exit with code 1, with the fixed diagnostic `Image stream ended without a final image.` Partial files remain available.

Completion does not depend on receiving all three requested partial images: the API contract allows the final image to arrive earlier. SDK EOF semantics, request parameters, image formats, filenames, and existing HTTP/SSE/filesystem errors are unchanged. Only the handwritten example and its existing test file change; no generated code or dependencies are modified.

## Verification

- Added empty-EOF and partial-only-EOF regressions plus a completed-without-partials success control to the existing real-HTTP/subprocess suite. Before the fix, the two new failure cases failed and all five controls passed.
- All seven scenarios passed on exact Node 22.0.0, 24.19.0, and 26.7.0: 21 real-process checks using the actual TypeScript example and public SDK. Existing HTTP errors, SSE errors, write failures, partial/final file bytes, and successful completion remain covered.
- Independent public-example checks also confirmed that receiving all three partial images without a final event still fails, while completion with zero or one partial succeeds.
- New failure cases verify a nonzero exit, no final-image file, retained partial output, and diagnostics without encoded image data or credentials. All network traffic and image fixtures were synthetic and local; no live image generation occurred.
- Normal full-project TypeScript checking, pinned Oxfmt 0.62.0 and Oxlint 1.79.0, and whitespace checks pass. Local Vitest is 4.1.10; CI will exercise the pinned 4.1.11 environment.

## Adversarial review

Three independent reviews checked completion ordering, early final events, existing error preservation, output-file behavior, privacy, runtime compatibility, and subprocess/server cleanup. The flag is set only after a successful final write and adds no parser, count requirement, size limit, cancellation policy, or shared state. Final review found no blocking issues.

### DIFF
--- a/examples/images/image-stream.ts
+++ b/examples/images/image-stream.ts
@@ -16,6 +16,7 @@ const main = async () => {
     partial_images: 3,
   });
 
+  let receivedFinalImage = false;
   for await (const event of stream) {
     let filename: string;
     let imageBuffer: Buffer;
@@ -39,6 +40,7 @@ const main = async () => {
         filename = 'final_image.png';
         imageBuffer = Buffer.from(event.b64_json, 'base64');
         fs.writeFileSync(filename, imageBuffer);
+        receivedFinalImage = true;
         console.log(`    Saved to: ${path.resolve(filename)}`);
         break;
       }
@@ -46,6 +48,10 @@ const main = async () => {
         console.log(`❓ Unknown event: ${event}`);
       }
     }
+  }
+
+  if (!receivedFinalImage) {
+    throw new Error('Image stream ended without a final image.');
   }
 };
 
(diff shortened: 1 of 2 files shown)


jbeckwith-oai (approved):
Reviewed the exact head against the stated image-stream completion scope. The completion flag is set only after the final image write succeeds, clean EOF without a final image now reaches the existing error boundary with exit code 1, and partial files are intentionally retained. Completion correctly does not depend on receiving every requested partial; HTTP, SSE, filesystem, request, filename, and successful-final behavior remain unchanged. Focused Node 24 subprocess verification passed 7/7 scenarios, including empty EOF, partial-only EOF, and completion without partials; diff check clean. No blocking findings.

PR #2620, opened 2026-09-05
fix(build): postprocess the selected package manifest

HAYDEN-OAI:
## Summary

Make the package postprocessor read and write `package.json` inside its already-resolved output directory.

The script honors `DIST_PATH` (or its script-relative `dist` default) when rewriting files and discovering exports, but its final manifest operations use the literal, cwd-relative `dist/package.json`. An alternate output directory or invocation from another working directory therefore leaves the intended package's exports stale and can overwrite an unrelated package's exports. If the unrelated path does not exist, postprocessing fails even though the selected output is valid.

Resolve one manifest path from `distDir` and reuse it for both operations. This is a three-line production change; export-map construction, file rewriting, and the canonical build command are unchanged. No generated SDK files, public SDK APIs, dependencies, or generation metadata change.

## Regression coverage

The tests copy the actual script byte-for-byte into disposable project fixtures and execute it in a native Node subprocess. They cover:

- Absolute and relative `DIST_PATH` values, including spaces in paths.
- The script-relative default when invoked from another working directory.
- A valid selected output without any `cwd/dist/package.json`.
- A missing selected manifest: fail with `ENOENT` and leave an unrelated manifest untouched.
- The normal repository-root build as a compatibility control.

All five failure modes were reproduced before the fix; the canonical control already passed. The tests verify the selected manifest's metadata and exports, unrelated manifest bytes, existing CommonJS/ESM private-state rewrites, declaration padding, and unchanged source-map/source files. No repository output or shared dependencies are used as writable fixtures.

## Verification

- All six focused subprocess tests pass on exact Node 22.0.0, 24.19.0, and 26.7.0.
- Independent before/after artifact checks cover the same six scenarios across all three runtimes; an additional review covers symlinked output directories.
- Fresh canonical builds on Node 24.19.0 are byte-for-byte identical before and after the change: all 2,953 output files, including package metadata and declarations.
- The existing packed-package suite passes on Node 22.22.3 and 24.19.0, including CommonJS, ESM, ES2020 browser bundling, type/export checks, and 1,302 source maps.
- Packed CommonJS and ESM authentication checks pass on the exact Node 22.0.0 floor with the supported Undici 7 dependency line.
- Normal full-project `tsc --noEmit`, pinned Oxfmt 0.62.0 and Oxlint 1.79.0, JavaScript syntax checking, and whitespace checks pass.

Local focused tests use cached Vitest 4.1.10; repository CI will verify the pinned 4.1.11 environment.

## Adversarial review

Independent reviews checked default and overridden directory selection, relative paths, symlink behavior, metadata preservation, private export exclusions, error propagation, and temporary-fixture cleanup. A final review added the missing-selected-manifest failure control and rechecked it. Both manifest operations now share the same directory as the existing file transformations, without changing parsing or export-map logic.

### DIFF
--- a/scripts/utils/postprocess-files.cjs
+++ b/scripts/utils/postprocess-files.cjs
@@ -5,6 +5,7 @@ const path = require('node:path');
 const distDir = process.env['DIST_PATH']
   ? path.resolve(process.env['DIST_PATH'])
   : path.resolve(__dirname, '..', '..', 'dist');
+const packageJsonPath = path.join(distDir, 'package.json');
 
 async function* walk(dir) {
   for await (const d of await fs.promises.opendir(dir)) {
@@ -138,11 +139,11 @@ async function postprocess() {
     }
   }
   await fs.promises.writeFile(
-    'dist/package.json',
+    packageJsonPath,
     JSON.stringify(
       Object.assign(
         /** @type {Record<String, unknown>} */ (
-          JSON.parse(await fs.promises.readFile('dist/package.json', 'utf-8'))
+          JSON.parse(await fs.promises.readFile(packageJsonPath, 'utf-8'))
         ),
         {
           exports: newExports,
--- a/tests/postprocess-files.test.ts
+++ b/tests/postprocess-files.test.ts
@@ -1,0 +1,124 @@
+import { execFile } from 'node:child_process';
+import { copyFile, mkdir, mkdtemp, readFile, rm, writeFile } from 'node:fs/promises';
+import { tmpdir } from 'node:os';
+import path from 'node:path';
+import { promisify } from 'node:util';
+
+const execute = promisify(execFile);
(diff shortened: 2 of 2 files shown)


jbeckwith-oai (approved):
Reviewed exact head bedf78e1b74bd1c29f33c118e87e9deb2cee4434 against the requested outcome: both manifest reads and writes must target the same resolved distDir already used for traversal and rewriting. I checked absolute and relative DIST_PATH handling, the script-relative default from a foreign cwd, missing-selected-manifest failure isolation, unrelated cwd/dist preservation, metadata/export regeneration, private-state rewrites, and unchanged source maps/sources. The production change is narrowly scoped and preserves the existing export construction and error behavior. I also ran the six focused subprocess cases successfully on local Node 22.15.1, 24.16.0, and 26.2.0, plus Oxfmt, Oxlint, and node --check. No blocking findings.

PR #2621, opened 2026-09-05
fix(build): narrow Git-install staging detection

HAYDEN-OAI:
## Summary

Narrow the Git-install detector so an ordinary checkout is not classified as package-manager staging merely because its parent is named `tmp` or `.tmp`.

The detector gates the existing prepare command's build-and-swap workflow. A false positive replaces the checkout with the built package, deleting Git metadata, tests, and untracked notes. This was reproduced with ordinary offline installs in disposable `tmp/openai-node` and `.tmp/openai-node` checkouts.

Match the immediate parent and basename against the existing installed-package layout and observed staging names:

- `node_modules/*` — existing behavior, unchanged.
- `tmp/git-clone*` — npm Git staging.
- `tmp/_tmp_*` — pnpm Git staging.
- `.tmp/*.prepare` — Yarn Classic Git preparation.

Use native Bash parameter expansion to preserve literal path components. Command substitution around `basename` strips trailing newlines and can accidentally turn a nonmatching directory name into a matching one.

Only the handwritten detector and a regression test change. The prepare command, build script, swap implementation, dependencies, SDK source, and generation metadata remain unchanged.

## Verification

- The initial lifecycle regression run failed for the two ordinary temporary checkouts and passed all six controls. Ten final tests pass on exact Node 22.0.0, 24.19.0, and 26.7.0, including ancestor-name and literal-newline controls.
- Tests execute the repository's actual prepare command with byte-copied detector and swap scripts in disposable fixtures. Only successful compilation is stubbed. They check checkout-file preservation, skipped builds, expected staging promotion, retained dependencies, and loading the resulting fixture package.
- Genuine dependency-free local-Git installs pass with npm 10.5.1, pnpm 11.5.1, and Yarn Classic 1.22.22 using the final detector bytes and fixture-local caches/stores. Ordinary installs retain Git metadata, tests, and notes; Git-dependency installs run the original swap and produce loadable flattened artifacts. Compilation is stubbed in these fixtures as well; these are lifecycle checks, not a claim that the complete SDK was installed from Git.
- An independent Bash 3.2.57 matrix passes 40 isolated directory-name cases, including spaces, misleading ancestors, trailing newlines, and native cwd normalization.
- Normal full-project `tsc --noEmit`, pinned Oxfmt 0.62.0 and Oxlint 1.79.0 for the TypeScript test, `bash -n`, and whitespace checks pass. Local Vitest is 4.1.10; CI will run the pinned 4.1.11 environment.

## Adversarial review and scope

Independent review found the newline-normalization issue in the first implementation. The detector was tightened to use literal Bash expansion, two regression cases were added, and the real package-manager and isolated-name checks were repeated against the final bytes. Final reviews found no remaining issues within this scope.

Detection remains a layout heuristic, not proof that a directory is disposable: a manually created directory using one of the staging-like names can still match. This PR does not redefine the existing `node_modules` case or introduce a general package-manager detection framework. Compatibility was exercised for the versions above; older or custom staging conventions are not claimed as verified.

### DIFF
--- a/scripts/utils/check-is-in-git-install.sh
+++ b/scripts/utils/check-is-in-git-install.sh
@@ -1,8 +1,12 @@
 #!/usr/bin/env bash
-# Check if you happen to call prepare for a repository that's already in node_modules.
-[ "$(basename "$(dirname "$PWD")")" = 'node_modules' ] ||
-# The name of the containing directory that 'npm` uses, which looks like
-# $HOME/.npm/_cacache/git-cloneXXXXXX
-[ "$(basename "$(dirname "$PWD")")" = 'tmp' ] ||
-# The name of the containing directory that some package managers use, e.g. .tmp/XXXXX
-[ "$(basename "$(dirname "$PWD")")" = '.tmp' ]
+# Keep installed packages and the staging names used by npm, pnpm, and Yarn Classic.
+# A tmp or .tmp parent alone does not identify a disposable Git-install checkout.
+parent="${PWD%/*}"
+case "${parent##*/}/${PWD##*/}" in
+  node_modules/* | tmp/git-clone* | tmp/_tmp_* | .tmp/*.prepare)
+    exit 0
+    ;;
+  *)
+    exit 1
+    ;;
+esac
--- a/tests/git-install-scripts.test.ts
+++ b/tests/git-install-scripts.test.ts
@@ -1,0 +1,127 @@
+import { spawnSync } from 'node:child_process';
+import {
+  chmodSync,
+  copyFileSync,
+  existsSync,
+  mkdirSync,
+  mkdtempSync,
+  readFileSync,
+  rmSync,
+  writeFileSync,
+} from 'node:fs';
+import { tmpdir } from 'node:os';
+import path from 'node:path';
+
+const repoRoot = process.cwd();
+const {
+  scripts: { prepare },
+} = JSON.parse(readFileSync(path.join(repoRoot, 'package.json'), 'utf-8'));
+const describeBash = process.platform === 'win32' ? describe.skip : describe;
+const checkoutFiles = {
+  '.git/HEAD': 'ref: refs/heads/local-work\n',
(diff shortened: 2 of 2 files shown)


markstuart-oai (approved):
Reviewed the complete diff at `7b8f4346332bb196f5b334920e0464c2e6a8bc40`. No substantive correctness or maintainability findings.

The single Bash `case` keeps the decision at the existing detector boundary, matches only immediate path components, and preserves literal newlines without subprocess normalization. The tests exercise the actual prepare command and swap script, checking both checkout preservation and the resulting loadable package.

Validation: all 10 lifecycle tests passed with Vitest 4.1.11 on Node 24.19.0. An independent Bash 3.2 matrix passed all 10 directory cases; the unchanged-main detector misclassified four negative controls. Shell syntax, formatting, lint, and diff checks passed. I also checked npm/Yarn staging construction in their implementation source. The lifecycle tests stub compilation; I did not repeat full package-manager Git installs. The remaining staging-like-directory heuristic is documented and is not broadened by this change.

sylvesterkaczmarek (commented):
I focused on the detector’s false-positive side because the consequence here is unusually destructive: a mistaken staging classification can replace an ordinary checkout and remove Git metadata, tests and local files. Matching the immediate `parent/basename` pair is materially safer than treating any `tmp` or `.tmp` parent as sufficient evidence.

Using Bash parameter expansion instead of `basename` is also a worthwhile detail. It keeps the path components literal, including trailing newlines, so the detector itself does not normalize an otherwise non-matching directory into a staging-shaped name. The regression matrix covers the cases I would want here: ordinary `tmp`/`.tmp` checkouts, misleading ancestor names, newline-containing components, and positive npm/pnpm/Yarn Classic plus `node_modules` layouts. I also checked that the existing `node_modules/*` behaviour remains immediate-parent based.

Given the destructive nature of the prepare-and-swap path, this narrower heuristic and the negative coverage look like the right trade-off.

PR #2633, opened 2026-09-05
fix(build): fail when a compiler worker is signaled

HAYDEN-OAI:
## Problem

A compiler worker terminated by a signal reports a `null` exit code. The vendored build tool passed that value through as if it were numeric, then `codes.find(code => code !== 0) || 0` converted it to success. A signaled first target could also hide a later target's real compiler error.

This reproduces through the actual CLI, not a mocked child process. In an isolated copy of current main, the canonical `./scripts/build` also exited 0 despite a real TS2322 diagnostic when the first compiler worker received an injected SIGTERM after emitting its shared helpers. All later postprocessing and import smoke checks passed, so those checks did not recover the lost compiler failure.

## Change

Normalize a missing numeric exit code to `1` in the worker's exit callback. This is a one-line orchestration fix; successful `0`, ordinary numeric failures, first-failed-target ordering, spawn-error handling, and parent-exit cleanup remain unchanged.

Add a small real-CLI subprocess suite covering numeric 0/23, SIGTERM/SIGKILL, and both orders of mixed signaled/numeric failures. Each fixture compiler terminates only its own worker. Workers run sequentially and their invocation count is verified. Signal-specific cases are POSIX-only; numeric controls also run on Windows.

## Verification

- Before: the three signal-first regressions failed because the CLI returned 0; all three controls passed.
- After: all six regressions pass on Node 22.22.3, 24.19.0, and 26.7.0.
- The same isolated canonical-build failure now exits 1 at compilation, before postprocessing.
- Full handwritten suite: **7,660 tests passed across 197 files** on Node 24.19.0.
- Canonical `./scripts/lint`, TypeScript 6.0.3 `tsc --noEmit`, and normal `./scripts/build` pass with the repository-pinned tools.
- Normal build output is byte-for-byte identical to a clean build of main `03acf94b7d76551bf4d612fe7582ddb6975c8ea3`.

## Scope and review

Only the SDK-maintained vendored worker orchestration and a new handwritten test file change. Both files are absent from the verified pure-generated snapshot `d223f5ab382c6a7d64f3863e75865624cbefad21`; generated helpers, upstream provenance, license, dependencies, SDK source, and release policy are untouched. No existing open PR covers this failure.

Adversarial review checked signal ownership, numeric exit precedence, subprocess cleanup, platform differences, and regression false positives. The test was tightened to tolerate ambient Node warnings without changing the environment or weakening its exit-status assertions, then reviewed and verified again.

### DIFF
--- a/scripts/_vendor/tsc-multi/src/build.ts
+++ b/scripts/_vendor/tsc-multi/src/build.ts
@@ -73,7 +73,7 @@ async function runWorker({
   try {
     return await new Promise<number>((resolve, reject) => {
       worker.on('error', reject);
-      worker.on('exit', resolve);
+      worker.on('exit', (code) => resolve(code ?? 1));
     });
   } finally {
     removeExitHandler();
--- a/tests/tsc-multi-worker-exit.test.ts
+++ b/tests/tsc-multi-worker-exit.test.ts
@@ -1,0 +1,74 @@
+import { spawnSync } from 'node:child_process';
+import { mkdtempSync, readFileSync, rmSync, writeFileSync } from 'node:fs';
+import { createRequire } from 'node:module';
+import { tmpdir } from 'node:os';
+import path from 'node:path';
+
+const repoRoot = process.cwd();
+const cli = path.join(repoRoot, 'scripts/_vendor/tsc-multi/src/cli.ts');
+const register = createRequire(path.join(repoRoot, 'package.json')).resolve(
+  'ts-node/register/transpile-only',
+);
+const testSignals = process.platform === 'win32' ? test.skip : test;
+
+describe('tsc-multi compiler worker exits', () => {
+  let fixture: string;
+
+  beforeEach(() => {
+    fixture = mkdtempSync(path.join(tmpdir(), 'openai-tsc-worker-exit-'));
+  });
+
+  afterEach(() => {
+    rmSync(fixture, { recursive: true, force: true });
+  });
+
+  function runWorkers(outcomes: (number | NodeJS.Signals)[]): number | null {
+    const compiler = path.join(fixture, 'compiler.cjs');
+    const invocations = path.join(fixture, 'invocations');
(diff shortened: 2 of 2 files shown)


jbeckwith-oai (approved):
Reviewed exact head 3c488f02bf3582c47bf8117e68bdcf6439c7bc0b against the signaled compiler-worker failure case. Node reports a null child-process exit code when termination is by signal, so normalizing only nullish codes to 1 preserves numeric 0 and nonzero statuses while preventing the existing first-nonzero aggregation from turning a signaled worker into success. The real-CLI subprocess coverage exercises numeric 0 and 23, SIGTERM and SIGKILL, both mixed-failure orders, sequential invocation count, parent signal isolation, and Windows signal skips. Spawn-error rejection and parent-exit cleanup are unchanged. The two-file diff is narrowly scoped, no unresolved review threads exist, and active CI has no actual failure conclusions.

markstuart-oai (approved):
Reviewed both changed files and the worker → build → CLI exit path at 3c488f02bf3582c47bf8117e68bdcf6439c7bc0b. Normalizing the null exit code at the worker boundary is the right minimal fix: numeric success/failure and first-failing-target order remain intact, while signal termination can no longer collapse to success. Promise settlement, spawn-error rejection, and each worker's parent-exit cleanup remain independent. No substantive correctness or maintainability findings.

Validated on Linux / Node 24.20.0: all six real-CLI regression cases pass; the same suite against base reproduces all three signal-first failures with the other three controls passing. Additional instrumented checks cover error/exit ordering, single settlement and cleanup, and independent concurrent-worker cleanup; a disposable real spawn confirms ENOENT behavior. Canonical build (including CJS/ESM smoke checks) and lint pass. Full tsc --noEmit was attempted but is limited by unavailable example dependencies (dotenv, @azure/identity, express); Windows and other Node versions were not rerun.

PR #2634, opened 2026-09-05
fix(responses): omit parsed values from replayed history

HAYDEN-OAI:
## Problem

`toResponseInputItems()` promises clean input items for replaying response history, but it currently retains the SDK's enumerable `parsed` and `parsed_arguments` values. The public `responses.parse()` and `responses.stream()` helpers add those values for the application; they are not part of the response-history wire format.

This breaks an ordinary public workflow: a supported Standard Schema validator can return a `BigInt`, parsing succeeds, but `responses.create({ input: toResponseInputItems(parsed.output), ... })` then throws during JSON serialization before making the follow-up request. Streamed refusal content also retains the helper-added `parsed: null` field.

## Change

At the existing history-normalization boundary:

- Omit `parsed` from output-text and refusal content.
- Omit `parsed_arguments` from function-call items.
- Preserve the original parsed response, raw text/arguments, order, phase, namespaces, unknown API fields, and meaningful falsy values.
- Keep unchanged item identity and preserve input strings/content arrays, including assistant inputs with an extra ID. Existing `created_by` normalization remains in place.

Only the handwritten normalizer and a new cohesive regression file change. There is no parser, resource-type, request-encoder, dependency, generated-file, or exported API change. Both changed paths are absent from the verified pure-generated snapshot `d223f5ab382c6a7d64f3863e75865624cbefad21`; no open PR covers this issue.

## Verification

- The exact final seven-case regression file on unchanged main `03acf94b7d76551bf4d612fe7582ddb6975c8ea3` produces **six expected failures and one passing control**: four parse/stream × message/function-call BigInt failures, streamed-refusal metadata, and metadata removal fail; ordinary nonstreaming refusal remains successful.
- All seven cases pass after the fix. The public regressions make the follow-up request through the real SDK with in-memory fetch, assert its complete serialized input, and verify the original parsed values remain available.
- **7,661 handwritten tests pass across 197 files** on Node 24.19.0; the focused parser/stream/normalizer suite passes all 66 tests.
- Final normalizer regressions and existing tests pass all 11 cases on Node 22.22.3 and 26.7.0.
- Canonical full lint, full TypeScript 6.0.3 checking, and the final canonical build pass with the pinned tools. A focused helper-dependency check also passes TypeScript 4.9.5 on both main and the fix.
- The actual built CJS/ESM package passes all eight parse/stream × message/function-call replay probes on the exact Node 22.0.0 floor. The same built-package probe fails with the original BigInt error on main.
- CJS and ESM declaration contracts are unchanged apart from the explanatory JSDoc.

## Adversarial review

Reviewed message discrimination, string and array inputs, mutation and identity, non-JSON parsed values, refusal handling, function-call namespaces, future fields, and compatibility. Tightened the message-shape guard and its identity controls, then repeated review and verification. No broad schema filtering, new parsing logic, or global serialization change is introduced. All fixtures are synthetic; no live API request was made.

BEGIN_COMMIT_OVERRIDE
fix(responses): omit parsed values from replayed history (#2634)
END_COMMIT_OVERRIDE

### DIFF
--- a/src/lib/responses/ResponseInputItems.ts
+++ b/src/lib/responses/ResponseInputItems.ts
@@ -15,7 +15,8 @@ type ResponseAdditionalToolsInputItem = Extract<ResponseInputItem, { type: 'addi
 /**
  * Normalizes a mixed array of stored response history items into clean
  * `ResponseInputItem`s that can be sent back to `responses.create()`. Known items
- * that cannot be replayed without changing their meaning are omitted.
+ * that cannot be replayed without changing their meaning are omitted. SDK-only
+ * parsed values are removed without changing the stored items.
  *
  * @throws {TypeError} If an item type is not supported by the installed SDK.
  */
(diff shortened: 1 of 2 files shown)


jbeckwith-oai (approved):
Reviewed exact head dca4e68a2ef7e1c1f8794928679d1aeed739880d against the response-history replay contract. The assistant-output guard targets actual response messages while preserving string and input-content messages; output_text and refusal parts lose only SDK-only parsed metadata, and function calls lose only parsed_arguments. Modified items and parts are shallow-copied, original parsed responses remain untouched, unchanged items retain identity, and created_by normalization plus unknown fields, order, phase, raw text, raw arguments, namespaces, and meaningful falsy values are preserved. The public parse and stream regressions drive a real follow-up request with BigInt parsed values and cover messages, function calls, refusals, mutation, identity, and future fields. The two-file diff is scoped to the normalizer and cohesive tests; CI has no failures and there are no unresolved review threads.

PR #2635, opened 2026-09-05
fix(build): honor configured compiler worker limits

HAYDEN-OAI:
## Problem

The vendored compiler tool accepts and validates `maxWorkers` in `tsc-multi.json`, and its build scheduler already supports that limit. However, the CLI overwrites the configured value with `undefined` whenever `--maxWorkers` is omitted. A configuration that requests one worker therefore still runs multiple compilation targets concurrently.

This makes the configured concurrency ceiling ineffective for normal CLI builds. It is a configuration-precedence bug, not a change to the default worker count.

## Change

Use the configured worker limit when the CLI flag is absent. Explicit flags still take precedence, and leaving both unset preserves the existing default. Explicit zero remains intact for the scheduler's existing rejection rather than silently falling back to the configuration.

The production change is one line in the handwritten CLI. A separate six-case test exercises the actual CLI, argument parser, and configuration loader in isolated subprocesses, capturing only the options passed to the build boundary. It covers configured limits, smaller and larger explicit overrides, the unset default, flag-only configuration, and explicit zero.

Both changed paths are absent from the verified pure-generated snapshot `d223f5ab382c6a7d64f3863e75865624cbefad21`. The vendored README assigns this build tool to SDK maintainers. Generated helpers, provenance, license, SDK sources, dependencies, and configuration defaults are unchanged. No open PR covers this issue.

## Verification

- On unchanged main `124b6525497a93315f8563d8e4e3d552cbb0518f`, the exact regression has one expected failure and five passing controls. All six cases pass after the change.
- A separate real-worker CLI probe confirms the behavior without scheduler mocks. With a configured limit of one and no flag, main starts two workers concurrently; the fix runs them serially. Explicit limits of one and two behave identically before and after. Explicit zero is rejected before starting any compiler.
- The new regression and the existing compiler-worker exit suite pass all 12 cases on Node 22.22.3, 24.19.0, and 26.7.0.
- All 7,673 handwritten tests pass across 199 files on Node 24.19.0 through the canonical test script with two test workers. An initial full-suite run encountered local disk exhaustion; the unchanged rerun passed.
- Full TypeScript 6.0.3 checking, the canonical build, and canonical lint pass with the pinned repository tools.
- Every file in the default `dist` output is byte-for-byte identical to a build of the unchanged base revision.

## Adversarial review

Reviewed configuration validation, CLI precedence, meaningful zero, default behavior, child-process isolation and cleanup, generated ownership, and the test seam. An independent final review found no remaining required changes. The committed regressions use deterministic boundary assertions rather than timing assertions; the separate real-worker probe verifies the actual concurrency effect. No live API calls were made.

### DIFF
--- a/scripts/_vendor/tsc-multi/src/cli.ts
+++ b/scripts/_vendor/tsc-multi/src/cli.ts
@@ -82,7 +82,7 @@ yargs(process.argv.slice(2))
         force: args.force,
         watch: args.watch,
         clean: args.clean,
-        maxWorkers: args.maxWorkers,
+        maxWorkers: args.maxWorkers ?? config.maxWorkers,
       });
     },
   )
--- a/tests/tsc-multi-config.test.ts
+++ b/tests/tsc-multi-config.test.ts
@@ -1,0 +1,69 @@
+import { spawnSync } from 'node:child_process';
+import { mkdtempSync, rmSync, writeFileSync } from 'node:fs';
+import { createRequire } from 'node:module';
+import { tmpdir } from 'node:os';
+import path from 'node:path';
+
+const repoRoot = process.cwd();
+const cli = path.join(repoRoot, 'scripts/_vendor/tsc-multi/src/cli.ts');
+const build = path.join(repoRoot, 'scripts/_vendor/tsc-multi/src/build.ts');
+const register = createRequire(path.join(repoRoot, 'package.json')).resolve(
+  'ts-node/register/transpile-only',
+);
+
+describe('tsc-multi worker-limit configuration', () => {
+  let fixture: string;
+
+  beforeEach(() => {
+    fixture = mkdtempSync(path.join(tmpdir(), 'openai-tsc-config-'));
+    // Capture the build boundary while exercising the real CLI, config loader, and argument parser.
+    // Worker timing is irrelevant to whether the selected configuration reaches the scheduler.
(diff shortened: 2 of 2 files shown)


markstuart-oai (approved):
Reviewed both changed files and traced CLI configuration through validation, p-all scheduling, and worker exit handling. The nullish fallback fixes precedence at the owning CLI boundary without changing explicit overrides (including zero), scheduler defaults, or result ordering; no substantive correctness or maintainability findings.

Validated on Node 24.20.0: all 12 configuration/worker-exit tests pass; the new regression fails only its configured-limit case on the base revision. Twenty disposable real-worker subprocess probes across base and head confirm configured/explicit/default bounds, rejection of zero/negative/fractional limits before workers start, and continued bounded scheduling after worker failures. Changed-file formatting/lint checks also pass. Full SDK build and runtime matrix were not rerun.

jbeckwith-oai (approved):
Thoroughly reviewed the exact head. The nullish fallback correctly preserves CLI precedence: an explicit flag wins, an absent flag uses validated config, explicit zero remains zero, and both absent leaves the scheduler default unchanged. The focused tests exercise the real CLI/config path and cover all precedence cases. I also traced the value through config validation and p-all scheduling, checked the complete diff and review threads, and verified the active test matrix is green. No blocking findings.

PR #2636, opened 2026-09-05
fix(test): cancel Steady commands waiting for cache locks

HAYDEN-OAI:
## Problem

Steady's cache wrapper installs SIGINT and SIGTERM handlers before acquiring the cache lifecycle lock, but those handlers only forward signals to an already-created child. If an install or mock command is waiting for another cache user, cancellation is swallowed. Once the lock is released, the cancelled command can still start and exit successfully.

This is reproducible through the actual wrapper CLI on unchanged main. Holding the lock, interrupting the wrapper, and then releasing the lock still prints the command's execution marker and exits zero for either signal.

## Change

Make the existing lock wait cancellable with a standard AbortSignal before child startup. Recheck cancellation after acquiring the lock and before running the command. A pending SIGINT or SIGTERM returns the existing signal-equivalent status of 130 or 143 without starting the command or reacquiring another owner's lock for cleanup.

Preserve the first cancellation signal, distinguish the actual cancellation error from unrelated filesystem failures, and keep active-child signal forwarding, exit codes, lease cleanup, cache pruning, and lock timeout behavior unchanged.

Only the handwritten cache wrapper and a new cohesive CLI regression file change. Both paths are absent from the verified pure-generated snapshot `d223f5ab382c6a7d64f3863e75865624cbefad21`. There are no dependency, manifest, downloaded-source, SDK, lease-format, or generated-file changes, and no open PR covers this issue.

## Verification

- The exact final six-case regression file on unchanged main `124b6525497a93315f8563d8e4e3d552cbb0518f` has two expected pending-cancellation failures and four passing controls. All six pass after the fix on Node 22.22.3, 24.19.0, and 26.7.0.
- The regressions run the actual wrapper, hold another owner's lock throughout cancellation, and verify no command starts, the ownership sentinel survives, and no lease is created. Controls cover normal zero/nonzero exits and both signals after the command starts.
- Existing cache-lifecycle tests pass on all three Node lines. Cache and concurrent-publication controls also pass on the exact Node 22.0.0 floor.
- The complete Steady integration passes both before and after the fix, including the pinned installation, offline launch, source/runtime integrity checks, relative paths, server health, and signal-driven shutdown.
- Full TypeScript checking, canonical lint, and the final diff check pass.
- All 7,673 handwritten tests pass across 199 files on Node 24.19.0 through the canonical test script with two workers. An initial full-suite attempt was interrupted by host disk exhaustion; the unchanged rerun passed after disk space recovered.

## Adversarial review

Reviewed cancellation during lock acquisition, lock and lease ownership, error filtering, first-signal behavior, and active-child compatibility. Independent review caught a test-cleanup risk involving historical process IDs; the final fixture uses a bounded child lifetime and only signals the live owned wrapper handle. Repeated review found no remaining required changes.

Signal-dependent regressions are POSIX-only; ordinary exit controls remain portable. All fixture data is synthetic, and no live OpenAI API call is made.

### DIFF
--- a/scripts/steady/cache.cjs
+++ b/scripts/steady/cache.cjs
@@ -10,10 +10,10 @@ const { once } = require('node:events');
 const MAX_IDLE_MS = 30 * 24 * 60 * 60 * 1000;
 
 // Serialize lease changes and pruning, never the commands using the cache.
-async function locked(cache, action) {
+async function locked(cache, action, signal) {
   const lock = path.join(cache, '.lifecycle-lock');
   const deadline = Date.now() + 10_000;
-  for await (const tick of interval(50)) {
+  for await (const tick of interval(50, undefined, { signal })) {
     void tick;
     try {
       fs.mkdirSync(lock);
@@ -30,6 +30,7 @@ async function locked(cache, action) {
     }
   }
   try {
+    signal?.throwIfAborted();
     return action();
   } finally {
     fs.rmdirSync(lock);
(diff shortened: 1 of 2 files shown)


jbeckwith-oai (approved):
Reviewed the exact head across the full two-file diff. The AbortSignal integration makes only the pre-spawn lock wait cancellable, preserves the first signal, rechecks cancellation after lock acquisition, removes only a lock acquired by this process, and leaves active-child forwarding plus ordinary exit and cleanup behavior intact. The regression tests exercise the real wrapper for both pending signals, both active-child signals, and ordinary zero/nonzero exits while protecting another owners lock and asserting no lease or command startup. Review threads are clear and the full Node matrix is green. No blocking findings.

markstuart-oai (approved):
Reviewed both changed files and the Steady install/launch and lease-cleanup paths. The abort signal stays with the pending acquisition; the post-acquisition check releases an acquired lock before cancellation returns, and the cleanup guard avoids reacquiring another owner's lock while preserving unrelated error behavior. Active-child forwarding remains intact.

Validated on Node 24.20.0: all six CLI regressions pass at this head; the same tests on base reproduce only the two pending-cancellation failures. Existing cache lifecycle tests and additional isolated checks for pre-abort, acquisition-boundary cancellation, first-signal preservation, exactly-once release, filesystem/action errors, and listener cleanup pass. Canonical lint passes. No substantive correctness or maintainability findings. Full TypeScript checking could not pass with the available dependency tree because example-only dotenv, @azure/identity, and express imports are missing; I did not rerun the full SDK suite or download the Steady integration dependencies.

PR #2637, opened 2026-09-05
fix(build): preserve config-relative compiler project paths

HAYDEN-OAI:
## Problem

The compiler CLI resolves configured project globs from the selected `tsc-multi.json` directory, but passes the resulting relative paths to workers running in the caller's working directory. With a nested or separately located config, a build can fail to find its project or silently compile a same-named project from the wrong directory.

The regression reproduces the silent case through the real CLI and TypeScript compiler. A nested config selects its own project, while a different project with the same relative name exists beside the caller. On main, the build exits successfully but emits the caller-directory decoy instead of the configured project.

## Change

Request absolute output paths from the existing `fast-glob` call. Each match then retains the directory it was resolved against when it reaches the compiler worker.

Configured projects remain relative to the selected config file. Positional CLI projects remain relative to the caller and continue to override configured projects. Worker working directories, compiler resolution, glob patterns, scheduling, and configuration defaults are unchanged.

The production change is one line in the handwritten config loader, with a separate four-case real-CLI regression file. Both paths are absent from the verified pure-generated snapshot `d223f5ab382c6a7d64f3863e75865624cbefad21`; SDK maintainers own this vendored build tool. No dependency, generated helper, provenance, license, SDK source, or published SDK API changes are included. No open PR covers this issue.

## Verification

- The exact final four-case regression on unchanged main `124b6525497a93315f8563d8e4e3d552cbb0518f` has two expected wrong-project failures and two passing controls. Both nested directory and explicit tsconfig-file selections emit the decoy before the fix.
- All four cases pass after the fix on Node 22.22.3, 24.19.0, and 26.7.0. Controls verify default same-directory configuration and positional-project precedence.
- The new regression and existing compiler-worker exit tests pass all 10 cases on Node 24.19.0.
- All 7,671 handwritten tests pass across 199 files on Node 24.19.0 through the canonical test script with two workers.
- Canonical lint, full repository TypeScript checking, targeted checking and formatting of the vendored config loader, and the canonical build pass with the pinned tools.
- The complete default `dist` output is byte-for-byte identical to the verified build of unchanged main, including CommonJS/ESM output and source maps.

## Adversarial review

Reviewed config-directory versus caller-directory ownership, directory and file project patterns, positional precedence, worker consumers, fixture cleanup, and generated ownership. Independent review found no remaining required changes. A test-only typing adjustment was rechecked against both main and the fix without weakening its assertions.

The fix uses the existing glob library's output option and preserves process working directories. All compiler fixtures are synthetic; no live API calls or dependency installations are needed by the regressions.

### DIFF
--- a/scripts/_vendor/tsc-multi/src/config.ts
+++ b/scripts/_vendor/tsc-multi/src/config.ts
@@ -32,7 +32,7 @@ export type Config = InferConfig & {
 };
 
 export async function resolveProjectPath(cwd: string, projects: string[]): Promise<string[]> {
-  return glob(projects, { cwd, onlyFiles: false });
+  return glob(projects, { cwd, absolute: true, onlyFiles: false });
 }
 
 export interface LoadConfigOptions {
--- a/tests/tsc-multi-project-paths.test.ts
+++ b/tests/tsc-multi-project-paths.test.ts
@@ -1,0 +1,108 @@
+import { spawnSync } from 'node:child_process';
+import { existsSync, mkdirSync, mkdtempSync, readFileSync, rmSync, writeFileSync } from 'node:fs';
+import { createRequire } from 'node:module';
+import { tmpdir } from 'node:os';
+import path from 'node:path';
+
+const repoRoot = process.cwd();
+const cli = path.join(repoRoot, 'scripts/_vendor/tsc-multi/src/cli.ts');
+const register = createRequire(path.join(repoRoot, 'package.json')).resolve(
+  'ts-node/register/transpile-only',
+);
+
+describe('tsc-multi project path resolution', () => {
+  let fixture: string;
+
+  beforeEach(() => {
+    fixture = mkdtempSync(path.join(tmpdir(), 'openai-tsc-project-paths-'));
+  });
+
+  afterEach(() => {
+    rmSync(fixture, { recursive: true, force: true });
+  });
+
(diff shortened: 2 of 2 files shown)


markstuart-oai (approved):
Reviewed the full change and the config loader, CLI precedence, worker launch, and TypeScript builder/transpile consumers. No substantive findings. Requesting absolute paths from the existing glob resolver puts the invariant at the right boundary and avoids adding compensating path logic to workers. The focused regression covers both directory and explicit tsconfig selections, default configuration, and positional-project precedence.

Independently verified on Node 24.19.0: all 10 path-resolution and worker-exit tests pass through `./scripts/test`. The exact four-case regression on freshly fetched main (`aa08b449`) reproduces both wrong-project failures, with both controls passing. `git diff --check` passes. I did not rerun the full build, complete suite, or cross-runtime matrix.

jbeckwith-oai (approved):
Reviewed the exact two-file head and traced project paths from config selection through fast-glob, CLI positional override, and the compiler worker boundary. Returning absolute matches preserves the config directory as the resolution owner without changing glob patterns, worker cwd, positional-project precedence, or default same-directory behavior. The real CLI regression specifically guards the silent same-named decoy case for both directory and tsconfig-file selections and includes the relevant controls. Review threads are clear and the full Node test matrix is green. No blocking findings.

PR #2638, opened 2026-09-05
fix(build): fail on reported transpile errors

HAYDEN-OAI:
## Problem

The vendored compiler worker always returns success in `transpileOnly` mode, even after TypeScript reports an error. A syntax-error fixture prints TS1109 and emits JavaScript that `node --check` rejects, but the real compiler CLI exits zero. An unreadable project tsconfig likewise reports TS5083 and exits successfully without emitting output.

This lets callers treat a failed transpilation as a successful build. Transpile-only mode should still skip semantic type checking; it should not discard errors already reported by the compiler.

## Change

Return the existing transpilation result through the worker and CLI. Error-category diagnostics and reported unreadable-config failures produce a nonzero status. Aggregate that status across every project so a later successful project cannot erase an earlier failure, while continuing to process and emit all projects.

Diagnostic reporting, JavaScript/source-map/helper emission, semantic-type-error skipping, and the normal non-transpile build path remain unchanged. The implementation uses local return values and accumulators; it adds no parser, reporter abstraction, retained state, or new type checking.

Only the handwritten worker and a new cohesive regression file change. Both paths are absent from the verified pure-generated snapshot `d223f5ab382c6a7d64f3863e75865624cbefad21`; SDK maintainers own the vendored build tool. Dependencies, configuration, generated helpers, provenance, license, and published SDK APIs are untouched. No open PR covers this issue.

## Verification

- The exact final five-case regression on unchanged main `aa08b4498b1923f9fa8c11d0055e83411d9457f3` has three expected false-success failures and two passing controls. The failures cover syntax errors, mixed failed/successful projects, and an unreadable tsconfig.
- All five real-CLI cases pass after the change on Node 22.22.3, 24.19.0, and 26.7.0. Valid code and semantic-only type mismatches still succeed; syntax errors preserve their emitted files and diagnostics but fail the build. Node's own syntax check independently confirms the invalid emitted JavaScript.
- The new regressions plus existing worker-exit and worker-limit tests pass all 17 cases on Node 24.19.0.
- All 7,678 handwritten tests pass across 200 files on Node 24.19.0 through the canonical test script with two workers.
- Full repository lint and TypeScript checks, targeted vendored-worker type/format checks, the canonical build, and the final diff check pass with the pinned tools.
- Every file in the normal SDK `dist` output is byte-for-byte identical to the verified build of unchanged main, including CommonJS/ESM output and source maps.

## Adversarial review

Reviewed Error-category filtering, every-file/every-project processing, failure retention, unreadable-config handling, exception behavior, and unchanged diagnostics/emission/semantic skipping. Independent review found no remaining required changes. A test-only helper placement adjustment was rerun against both main and the fix without weakening assertions.

All regression projects and compiler inputs are synthetic, with bounded subprocesses and owned temporary-directory cleanup. No live API calls are made.

### DIFF
--- a/scripts/_vendor/tsc-multi/src/worker/worker.ts
+++ b/scripts/_vendor/tsc-multi/src/worker/worker.ts
@@ -50,8 +50,7 @@ export class Worker {
 
   public run(): number {
     if (this.data.transpileOnly) {
-      this.transpile();
-      return 0;
+      return this.transpile();
     }
 
     const builder = this.createBuilder();
@@ -375,13 +374,17 @@ export class Worker {
     );
   }
 
-  private transpile() {
+  private transpile(): number {
+    let exitCode = 0;
     for (const project of this.data.projects) {
-      this.transpileProject(project);
+      if (this.transpileProject(project) !== 0) {
+        exitCode = 1;
+      }
     }
+    return exitCode;
   }
 
-  private transpileProject(projectPath: string) {
+  private transpileProject(projectPath: string): number {
     const tsConfigPath = this.system.fileExists(projectPath)
       ? projectPath
       : nodePath.join(projectPath, 'tsconfig.json');
@@ -394,7 +397,7 @@ export class Worker {
 
     const config = this.ts.getParsedCommandLineOfConfigFile(tsConfigPath, options, parseConfigFileHost);
     if (!config) {
-      return;
+      return 1;
     }
 
     let resolvedShareHelpers: string | undefined;
@@ -420,6 +423,7 @@ export class Worker {
       ],
     };
 
+    let exitCode = 0;
     for (const inputPath of config.fileNames) {
       // - Ignore if file does not exist
       // - or if file is a declaration file, which will generate an empty file and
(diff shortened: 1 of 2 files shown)


jbeckwith-oai (approved):
Reviewed the exact two-file head and traced status from TypeScript transpileModule diagnostics through per-project aggregation, the worker return, and CLI process exit. Error diagnostics now retain a nonzero result while all files and later projects continue to emit; successful and semantic-only transpilation remain zero, and an unrecoverable config failure propagates as failure. The real-CLI tests cover syntax errors, mixed failed/successful projects, unreadable config, emitted output/source maps, and the semantic-only control. Review threads are clear; Node 26 and the non-matrix checks are green, with Node 22/24 still pending. No blocking findings.

markstuart-oai (approved):
Reviewed both changed files and the compiler/config/reporter/worker/CLI propagation. The local status accumulators preserve Error-category failures across files and projects without short-circuiting emission or introducing a second reporting policy; unreadable-config failures and exceptions reach a nonzero CLI exit, while semantic-only checking remains skipped. No correctness or structural findings.

Independently ran the exact five-case regression against base aa08b449: three expected false-success failures and two passing controls. All 17 focused tests pass on this head through ./scripts/test with Node 24.20.0 / TypeScript 6.0.3. Six additional bounded CLI probes pass for multiple files, reversed project order, missing-config followed by success, compiler exceptions, warning/message-only success, and Error retention alongside non-error diagnostics. Syntax-invalid emitted output is independently rejected by node --check. Changed-file formatting and diff checks pass. Full SDK suites and the other Node versions were not rerun in this review.

PR #2639, opened 2026-09-05
fix(examples): reject early Realtime socket closure

HAYDEN-OAI:
## Summary

Make the four handwritten Realtime examples report failure when their WebSocket closes before a terminal `response.done` event.

On current main, a server can send text (including `response.output_text.done`) and close the connection normally before `response.done`; both OpenAI and Azure examples then print `Connection closed!` and exit successfully. The two `ws` examples also exit successfully on an abrupt disconnect. This makes an unfinished request look successful to anyone running or adapting the examples.

Each example now records receipt of `response.done` and checks that flag in its existing socket-close handler. An early close prints a fixed diagnostic and sets `process.exitCode = 1`, allowing buffered output to drain. Successful responses still exit 0. Failed, cancelled, and incomplete terminal responses retain their existing diagnostic and exit status without an additional early-close warning. Existing transport-error handling is unchanged.

The implementation is eight net new lines per standalone example. There are no SDK implementation, generated-file, public API, dependency, authentication, or endpoint changes, and no reconnection or timeout framework.

## Regression coverage

Extend the existing real-entrypoint loopback TLS/WebSocket suite from 16 to 24 cases, covering OpenAI and Azure with both `ws` and the native WebSocket transport:

- A clean close after text completion but before `response.done`.
- An abrupt disconnect before response output.
- Existing completed, failed, cancelled, and incomplete terminal-response controls.

The tests run the actual examples and SDK transports, verify their three outgoing client events and provider URLs, retain the Azure bearer-header assertion, and ensure synthetic credentials, response metadata, and close reasons are not printed. Only Azure credential acquisition is replaced with a synthetic token; the connection and socket events are real.

Before the fix, the four clean-close cases and two `ws` abrupt-close cases returned exit code 0. Native abrupt-close cases already returned 1 but lacked the close diagnostic. All 16 terminal-response controls passed. With the fix, all 24 cases pass.

## Validation

- All 24 Realtime example cases pass on Node 22.22.3, 24.19.0, and 26.7.0.
- The complete canonical handwritten unit suite passes on Node 24.19.0: 7,681 tests across 199 files.
- Canonical `./scripts/lint`, repository TypeScript checks, and `git diff --check` pass.
- Canonical `./scripts/build` passes; the complete default SDK output is byte-identical to the clean main baseline.

## Adversarial review

Self-review and an independent adversarial review checked terminal-event versus socket-close ordering, preservation of failed terminal results, native and `ws` error behavior, output draining, synthetic credential isolation, and absence of close-reason or response-metadata logging. No required changes remained. The guard belongs only to these single-response examples; SDK transport behavior is unchanged.

### DIFF
--- a/examples/azure/realtime/websocket.ts
+++ b/examples/azure/realtime/websocket.ts
@@ -14,6 +14,7 @@ async function main() {
     deployment: deploymentName,
   });
   const rt = await OpenAIRealtimeWebSocket.azure(client);
+  let responseDone = false;
 
   // access the underlying `ws.WebSocket` instance
   rt.socket.addEventListener('open', () => {
@@ -54,6 +55,7 @@ async function main() {
 
   // response.done also covers failed, cancelled, and incomplete responses.
   rt.on('response.done', (event) => {
+    responseDone = true;
     if (event.response.status !== 'completed') {
       console.error('Response did not complete successfully.');
       process.exitCode = 1;
@@ -61,7 +63,13 @@ async function main() {
     rt.close();
   });
 
-  rt.socket.addEventListener('close', () => console.log('\nConnection closed!'));
+  rt.socket.addEventListener('close', () => {
+    if (!responseDone) {
+      console.error('WebSocket closed before the response completed.');
+      process.exitCode = 1;
+    }
+    console.log('\nConnection closed!');
+  });
 }
 
 main();
--- a/examples/azure/realtime/ws.ts
+++ b/examples/azure/realtime/ws.ts
@@ -14,6 +14,7 @@ async function main() {
     deployment: deploymentName,
   });
   const rt = await OpenAIRealtimeWS.azure(client);
+  let responseDone = false;
 
   // access the underlying `ws.WebSocket` instance
   rt.socket.on('open', () => {
@@ -54,6 +55,7 @@ async function main() {
 
   // response.done also covers failed, cancelled, and incomplete responses.
   rt.on('response.done', (event) => {
+    responseDone = true;
     if (event.response.status !== 'completed') {
       console.error('Response did not complete successfully.');
       process.exitCode = 1;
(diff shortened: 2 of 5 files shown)


jbeckwith-oai (approved):
Reviewed the exact five-file head across all four standalone examples and the real-entrypoint loopback suite. The response.done flag is set before the existing close call, so completed and terminal failed/cancelled/incomplete responses preserve their current outcomes while any clean or abrupt pre-terminal closure sets a nonzero exit after buffered output can drain. Existing transport-error handling is untouched. Coverage spans OpenAI/Azure and ws/native transports, verifies the actual client events and URLs, preserves Azure auth assertions, and checks that credentials, metadata, and close reasons do not leak. Review threads are clear; build/lint are green and the remaining matrix is pending. No blocking findings.

markstuart-oai (approved):
Reviewed all five files and every hunk at bc03f7894243fd28d72fcab4800da759f5e81809 against aa08b4498b1923f9fa8c11d0055e83411d9457f3, including both transport implementations and the terminal-event contract. No substantive findings.

The single responseDone flag is the minimal state these standalone examples need: socket closure cannot establish receipt of response.done. Setting it before rt.close() preserves completed, failed, cancelled, and incomplete outcomes; process.exitCode allows pending output to drain. The new diagnostic is fixed text and includes no credentials, response metadata, or close reason. A shared lifecycle abstraction would add indirection without simplifying these copyable examples.

Independently ran the canonical real-entrypoint loopback suite: all 24 cases pass on Node 22.22.3 and 24.20.0. On the actual base with the new regression cases, all 16 terminal controls pass; four clean closures and two ws abrupt closures incorrectly exit 0, while the two native abrupt closures lack the diagnostic. Synthetic Azure acquisition only; all connections stayed on loopback. Canonical lint, TypeScript no-emit, and diff checks also pass.

PR #2640, opened 2026-09-05
fix(lint): preserve leading-dash generated file paths

HAYDEN-OAI:
## Summary

Keep selected generated-file paths from being interpreted as oxlint command-line options.

The handwritten generated-file lint wrapper passes selected relative filenames directly after its options. A generated file named `-clean.ts` is therefore interpreted as oxlint's `-c` option instead of a file, causing both `--check` and `--fix` to fail before checking or fixing the intended file. This affects both direct filename arguments and the wrapper's existing file-list input.

Add the standard `--` end-of-options separator before the selected files in the two existing subprocess calls. File discovery, file-list parsing, generated ownership, lint configuration, repeated-fix convergence, and exit-status handling remain unchanged. No generated SDK files, dependencies, exported APIs, or custom-code budget settings change.

## Regression coverage

Three compact cases run the actual generated-file wrapper, discovery helper, configuration, and pinned oxlint CLI:

- A leading-dash filename passed directly.
- The same filename selected through a file list.
- An ordinary filename as the compatibility control.

Each case checks a clean file in both modes, verifies an unused-import diagnostic without modifying the file, applies the real fix, and checks the fixed file again. Unrelated generated and handwritten files must remain byte-identical.

Fixtures contain only owned directories and files. A fixture-owned launcher imports the unchanged real oxlint entrypoint by absolute file URL; no shared dependency directory is symlinked into the mutable fixture. Ambient runtime controls are preserved, and the test checks the relevant lint diagnostic rather than requiring all runtime stderr to be empty.

On unchanged main, both leading-dash cases fail while the ordinary-name control passes. With the two separators, all three cases pass.

## Validation

- All three new cases pass on Node 22.22.3, 24.19.0, and 26.7.0 with ambient runtime controls intact.
- The complete canonical handwritten unit suite passes on Node 24.19.0: 7,681 tests across 201 files.
- Canonical `./scripts/lint`, full repository TypeScript checking, and `git diff --check` pass.
- Canonical `./scripts/build` passes; the entire default SDK output is byte-identical to the clean `833e3e8d` baseline.

## Adversarial review

Self-review and an independent adversarial review checked placement of the separator after all intended options, unchanged direct/file-list selection and convergence behavior, real CLI execution, fixture ownership, and preservation of unrelated files. No required changes remained. The fix establishes the positional-argument boundary without adding path parsing or a new abstraction.

### DIFF
--- a/scripts/lint-generated.cjs
+++ b/scripts/lint-generated.cjs
@@ -54,7 +54,7 @@ function fixGeneratedFiles(files) {
     const before = fingerprint(files);
     const result = spawnSync(
       process.execPath,
-      [oxlint, '--config', generatedConfig, '--fix', '--quiet', ...files],
+      [oxlint, '--config', generatedConfig, '--fix', '--quiet', '--', ...files],
       { cwd: repositoryRoot, stdio: 'inherit' },
     );
 
@@ -77,7 +77,7 @@ function fixGeneratedFiles(files) {
 function checkGeneratedFiles(files) {
   const result = spawnSync(
     process.execPath,
-    [oxlint, '--config', generatedConfig, '--format', 'json', ...files],
+    [oxlint, '--config', generatedConfig, '--format', 'json', '--', ...files],
     { cwd: repositoryRoot, encoding: 'utf-8', maxBuffer: 10 * 1024 * 1024 },
   );
 
--- a/tests/lint-generated-literal-paths.test.ts
+++ b/tests/lint-generated-literal-paths.test.ts
@@ -1,0 +1,98 @@
+import { spawnSync } from 'node:child_process';
+import {
+  copyFileSync,
+  mkdirSync,
+  mkdtempSync,
+  readFileSync,
+  realpathSync,
+  rmSync,
+  writeFileSync,
+} from 'node:fs';
+import { tmpdir } from 'node:os';
+import path from 'node:path';
+import { pathToFileURL } from 'node:url';
+
+const repoRoot = process.cwd();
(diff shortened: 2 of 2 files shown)


jbeckwith-oai (approved):
Reviewed the exact two-file head. The standard end-of-options separator is placed after every intended oxlint option and before all selected paths in both check and fix subprocesses, so leading-dash generated filenames remain literal while file discovery, file-list parsing, convergence, and status handling are unchanged. The regression runs the real wrapper and pinned oxlint for direct, file-list, and ordinary-path cases; it verifies clean checks, diagnostics without mutation, real fixes, rechecks, and isolation of unrelated generated and handwritten files. Review threads are clear; CI has no failures and the Node matrix is pending. No blocking findings.

markstuart-oai (approved):
Reviewed both complete changed files and all three hunks against the PR merge base. Adding `--` after the intended options in both oxlint calls is the right minimal fix, consistent with fast-format's existing command boundary; no custom path parsing or abstraction is needed. Discovery, direct/file-list selection, repeated-fix convergence, diagnostics, and exit handling remain intact.

Independently ran the three new tests through ./scripts/test with real pinned oxlint 1.79.0, plus isolated direct/file-list probes covering CRLF lists, ordinary filenames, non-mutating diagnostics, 20 adjacent unused imports, rechecks, idempotence, and unchanged unrelated generated/handwritten bytes. The current base reproduces both leading-dash failures; the reviewed head passes. ./scripts/lint and git diff --check also pass. Fixtures own their mutable files and preserve ambient runtime controls. No substantive correctness or maintainability findings.

PR #2641, opened 2026-09-05
chore(ecosystem): remove unused proxy stub

HAYDEN-OAI:
## Summary

Remove the unused `ecosystem-tests/proxy.ts` stub and its four stale lint-exception entries.

The module defines an unexported `startProxy` function but never calls it. No tracked source, package script, documentation, or workflow references it; the only path references are the four exceptions being removed. The actual ecosystem runner uses its own separate proxy implementation in `ecosystem-tests/cli.ts`, and that was already the active implementation when the stub was introduced.

This is a deletion-only cleanup of handwritten files: 25 unused source lines and four obsolete configuration entries. The active CLI, its network behavior, all remaining lint exceptions, generated SDK files, dependencies, and published SDK APIs are unchanged.

## Validation

- The real printed lint configuration differs only by the four removed selectors; all other rules, override groups, settings, and selectors match. The active CLI's Git blob is unchanged.
- Eleven targeted existing tests pass: three lint/configuration checks, seven CLI unknown-project rejection cases, and help without credentials. The other 38 cases in those two files were intentionally not run for this deletion-only change.
- Canonical `./scripts/lint`, full repository TypeScript checking, and `git diff --check` pass on Node 24.19.0.
- Canonical `./scripts/build` passes, including SDK and Bedrock CommonJS/ESM imports. The complete SDK output is byte-identical to the clean `88963957` baseline.

No real proxy, live API, deployment, or network-test workflow was run locally for this cleanup.

## Adversarial review

Self-review and an independent adversarial review checked tracked references, introduction history, module exports and top-level behavior, active CLI ownership, and generated-file provenance. No required changes remained. The working proxy implementation is unchanged.


jbeckwith-oai (approved):
Reviewed the exact deletion-only head and independently checked repository references and history. ecosystem-tests/proxy.ts defines an unexported, never-invoked stub; the only path references are the four oxlint selectors removed in the same diff, while the active ecosystem proxy remains separately implemented in cli.ts. Removing the dead file and only its stale lint exceptions does not affect active network behavior, scripts, dependencies, generated output, or public APIs. Review threads are clear; CI has no failures and the Node/ecosystem matrix is pending. No blocking findings.

markstuart-oai (approved):
Reviewed the complete deletion and reference/introduction history. The stub has no exports or callers; the working proxy already lived in ecosystem-tests/cli.ts when it was introduced. That CLI, including proxy startup, environment configuration, and cleanup, is byte-identical. Removing the unused duplicate and its four exact lint selectors is the right simplification; no replacement abstraction or broader refactor is needed. Independently verified complete evaluated config equivalence except those selectors, 11 focused lint/CLI tests, and the canonical lint script plus git diff --check. No substantive correctness or maintainability findings. No live proxy/API tests were run.

PR #2643, opened 2026-09-05
test(bedrock): isolate fixtures from endpoint overrides

HAYDEN-OAI:
## Summary

Keep the two Bedrock cancellation/streaming test fixtures independent of a developer's `AWS_BEDROCK_BASE_URL` setting. Their clients intentionally exercise fixed synthetic endpoints, but the supported environment override could replace those endpoints before the assertions ran.

- Clear only that variable with file-local Vitest setup and restore its original value after each test.
- Preserve all existing assertions, synthetic bearer-token environment cases, and mock cleanup.
- Keep provider precedence, endpoint/region validation, SDK behavior, dependencies, and the global test harness unchanged.

Both changed files are wholly handwritten and absent from the verified pure-generated snapshot. The explicit `undefined` argument uses a narrow, documented lint exception because Vitest's typed API requires the second argument when removing a variable.

## Verification

Using the synthetic override `https://bedrock-runtime.us-west-2.amazonaws.com/openai/v1` with the actual public-client fixtures:

- Before: 75 of the existing 93 tests failed on the untouched base because the override conflicted with their fixed regions/endpoints.
- After: all 93 tests pass with the override on Node 22.22.3, 24.19.0, and 26.7.0.
- The same 93 tests also pass without the override.
- All 7,688 handwritten tests across 202 files pass with the synthetic override still present.
- Canonical formatting/lint, the full repository TypeScript check, and `git diff --check` pass.

The tests use their existing in-memory fetch fixtures; no live API requests, installations, or credential values are needed. Adversarial self-review and independent review confirmed the setup/teardown lifetime, environment restoration, unchanged authentication cases, and test-only scope.

### DIFF
--- a/tests/lib/bedrock-provider-bearer-cancellation.test.ts
+++ b/tests/lib/bedrock-provider-bearer-cancellation.test.ts
@@ -131,7 +131,15 @@ function throwAfterRemovingOnce(signal: AbortSignal): void {
   });
 }
 
-afterEach(() => vi.restoreAllMocks());
+beforeEach(() => {
+  // oxlint-disable-next-line unicorn/no-useless-undefined -- Vitest requires its second argument when removing an environment variable.
+  vi.stubEnv('AWS_BEDROCK_BASE_URL', undefined);
+});
+
+afterEach(() => {
+  vi.unstubAllEnvs();
+  vi.restoreAllMocks();
+});
 
 describe.each(providerCases)('%s %s credentials', (_name, endpoint, create) => {
   test('cancels a documented public models request while its token provider remains pending', async () => {
--- a/tests/lib/bedrock-runtime-streaming.test.ts
+++ b/tests/lib/bedrock-runtime-streaming.test.ts
@@ -1,3 +1,4 @@
+import { vi } from 'vitest';
 import OpenAI from 'openai';
 import { bedrock as bearerBedrock } from 'openai/providers/bedrock';
 import { bedrock } from 'openai/providers/bedrock/aws';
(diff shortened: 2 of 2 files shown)


jbeckwith-oai (approved):
Reviewed the exact two-file test-only head. Each affected fixture now removes only AWS_BEDROCK_BASE_URL with Vitests environment-stub API before the test and restores the original value afterward; existing bearer-token cases and mock cleanup remain intact. This isolates fixed synthetic endpoint/region assertions from developer overrides without changing provider precedence, SDK behavior, global harness state, or production code. The explicit undefined argument is required by the typed removal API, so the narrow documented lint exception is justified. Review threads are clear; CI has no failures and the Node/ecosystem matrix is pending. No blocking findings.

markstuart-oai (approved):
Reviewed both complete test files and every hunk. The file-local Vitest hooks give the synthetic endpoint fixtures clear ownership of AWS_BEDROCK_BASE_URL and restore it after each test, including failures. Existing bearer-token environment cases retain their finally cleanup, and mock restoration remains intact. Provider/authentication precedence, endpoint validation, runtime code, and ambient security/network controls are unchanged. The required undefined argument and narrow documented lint exception are justified; no shared helper is needed for these small local hooks.

Validation on Node 24.20.0 / Vitest 4.1.11: the canonical focused command passes all 93 tests with the synthetic endpoint override and with the ambient environment; the untouched base reproduces 75 failures with that override. Separate temporary instrumentation verified whole-environment restoration between tests and after intentional errors, with the override both present and absent. Targeted canonical formatter/linter checks and git diff --check pass. File sizes are 1,012→1,020 and 146→153 lines: no new 1k crossing. No substantive findings.

PR #2645, opened 2026-09-05
fix(examples): reject unsuccessful Responses streams

HAYDEN-OAI:
## Summary

Make `examples/responses/stream.ts` and `streaming-tools.ts` report unsuccessful final responses instead of exiting successfully. Draining a stream does not imply that its response completed: `finalResponse()` deliberately resolves the latest snapshot for failed or incomplete responses and after a clean EOF with an in-progress snapshot.

After draining, check the final response status, emit a status-only diagnostic, and set `process.exitCode = 1` unless it is `completed`. Keep existing event/final-response stdout, request parameters, tool schema, and normal rejected-request behavior unchanged. The SDK's `finalResponse()` contract is not changed, and no retry/resume behavior is added.

Reuse the existing Responses example integration harness for eight additional real-entrypoint cases: completed, failed, incomplete, and clean EOF after `response.created` for each example. Fixtures retain runtime/proxy/trust settings, use a synthetic API key and loopback endpoint, and verify request count, model, and the existing strict query tool.

## Validation

- Exact final tests against unchanged examples: six expected exit-status failures; both completed controls and all 26 existing background/guide cases pass. Request and authentication assertions pass before each failing exit-status assertion.
- All 34 focused integration tests pass on Node 22.22.3, 24.19.0, and 26.7.0. For the local Node 22.22.3 run, Node warnings were redirected to a temporary log because the environment's proxy preload emits an experimental warning; runtime/proxy/trust settings and test assertions were unchanged. The log contains only that known warning.
- All eight actual example-entrypoint checks pass on the exact Node 22.0.0 floor, with natural exits and one synthetic-authenticated loopback request each.
- Full handwritten suite: 7,707 tests across 203 files pass on Node 24.19.0.
- Canonical formatting, lint, full repository TypeScript check, and `git diff --check` pass.
- Adversarial self-review and independent review pass after narrowing the test changes; existing child cleanup is unchanged.

Only these two handwritten examples and one existing handwritten test file change. The example fixes total ten added lines; generated files, SDK source, dependencies, and exports are untouched.

### DIFF
--- a/examples/responses/stream.ts
+++ b/examples/responses/stream.ts
@@ -19,6 +19,10 @@ async function main() {
 
   const result = await runner.finalResponse();
   console.log(result);
+  if (result.status !== 'completed') {
+    console.error(`Response ended with status ${result.status ?? 'unknown'}.`);
+    process.exitCode = 1;
+  }
 }
 
 main();
--- a/examples/responses/streaming-tools.ts
+++ b/examples/responses/streaming-tools.ts
@@ -47,6 +47,12 @@ async function main() {
   for await (const event of stream) {
     console.dir(event, { depth: 10 });
   }
+
+  const result = await stream.finalResponse();
+  if (result.status !== 'completed') {
+    console.error(`Response ended with status ${result.status ?? 'unknown'}.`);
+    process.exitCode = 1;
+  }
 }
 
 main();
--- a/tests/lib/background-streaming-example.test.ts
+++ b/tests/lib/background-streaming-example.test.ts
@@ -10,7 +10,11 @@ import { expect, test } from 'vitest';
 type ResponseStatus = 'completed' | 'failed' | 'incomplete' | 'queued' | 'in_progress';
 const privateErrorDetail = 'Synthetic private response error detail';
 
-function responseEvents(chunks: readonly string[], status: ResponseStatus): ResponseStreamEvent[] {
+function responseEvents(
+  chunks: readonly string[],
+  status: ResponseStatus,
+  background: boolean,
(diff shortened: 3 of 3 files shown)


jbeckwith-oai (approved):
Reviewed the exact three-file head and traced finalResponse semantics for both real example entrypoints. Draining a stream is not treated as success: completed stays zero, while failed/incomplete and clean EOF with the latest in-progress snapshot emit a status-only diagnostic and set a nonzero exit. Existing event/final-response stdout, request parameters, strict query tool schema, normal request rejection, and SDK finalResponse behavior remain unchanged. The loopback integration adds both examples across completed/failed/incomplete/in-progress cases and verifies request count, auth, model/tool payloads, and private-detail non-leakage. Review threads are clear and CI is green. No blocking findings.

markstuart-oai (approved):
Reviewed all three changed files against ResponseStream, the accumulator/parser and iterator error contracts, and the streaming guide. The final status checks correctly distinguish completed responses from failed, incomplete and premature clean EOF snapshots, while preserving request/transport rejection. Keeping this small policy at each standalone example is clearer than introducing another helper; the existing loopback harness is reused without changing the SDK contract. No substantive findings.

Local validation: 55 ResponseStream/ResponsesParser tests passed, 32 synthetic base/head entrypoint cases passed, and canonical lint/format checks passed. The 34 loopback integration cases could not receive requests in this environment (a separate child-process probe reproduced request cancellation before server receipt). Full typechecking was limited by missing optional Azure/Express/dotenv example dependencies. I did not independently rerun the Node-version matrix.

PR #2648, opened 2026-09-05
docs(helpers): remove undefined main call from parsing example

HAYDEN-OAI:
## Summary

Remove the trailing `main();` call from the self-contained function-tool parsing example in `docs/helpers.md`.

The snippet already runs at the top level and does not define `main`. As written, copying it into a TypeScript module produces `TS2304: Cannot find name 'main'`. Transpile-only execution also throws `ReferenceError: main is not defined` after an otherwise successful API response. Removing the obsolete call preserves the existing request, parsing, output, and error propagation.

This changes only two lines in a handwritten guide (the call and its preceding blank line). No SDK code, generated files, dependencies, or other examples change.

## Verification

- Extracted the original and actual revised fenced examples and compiled them against the real public SDK declarations. The final example passes strict TypeScript 4.9.5 and 6.0.3; the original has the undefined-`main` diagnostic. The neighboring complete Zod example remains a clean control.
- Executed the original and actual final snippet using the real built SDK and a synthetic-authenticated loopback HTTP server on Node 22.0.0, 24.19.0, and 26.7.0: 18 executions covering a function call, an ordinary assistant message, and an HTTP 400, before and after.
- Successful responses now exit naturally with status 0 instead of 1, with identical stdout. HTTP 400 responses still exit with status 1 through the existing SDK error path. Each execution makes exactly one request with the existing strict query tool.
- Canonical `./scripts/format`, `./scripts/lint`, and `git diff --check` pass.
- Adversarial self-review and independent review confirm that no missing wrapper or second invocation is required.

Validation is scoped to the exact documentation artifact; this PR does not add a separate test framework or claim a full SDK test run. No live API calls or real credentials were used.

### DIFF
--- a/docs/helpers.md
+++ b/docs/helpers.md
@@ -147,8 +147,6 @@ if (toolCall?.type === 'function') {
   console.log(args);
   console.log(args.table_name);
 }
-
-main();
 ```
 
 ### Differences from `.create()`


jbeckwith-oai (approved):
Reviewed the exact handwritten example against its surrounding control flow and requested outcome. The snippet already executes at module top level and defines no main function, so removing the trailing main call eliminates the compile-time and runtime failure without changing the request, parsing, output, or SDK error propagation. Scope is limited to the obsolete call and blank line; completed and error behavior remain as intended. No in-scope blocking issues.

markstuart-oai (approved):
The example already awaits the parsing request at module scope and defines no main function. Deleting the stale call is the simplest fix: it removes the post-success ReferenceError while preserving the request, parsed output, and existing error propagation. No correctness, security, or maintainability issues found in the two-line deletion.

Independently checked the extracted before/after snippets against public SDK declarations built from matching source with strict TypeScript 4.9.5 and 6.0.3: the original fails only with TS2304 for main; the revised snippet and neighboring Zod example pass. git diff --check also passes. I did not rerun the author's runtime matrix or the full SDK suite.

PR #2649, opened 2026-09-05
docs(tools): support TypeScript 4.9 in the function loop

HAYDEN-OAI:
## Summary

Add an explicit `OpenAI.Responses.ResponseFunctionToolCall` type predicate to the function-call filter in the first complete example in `docs/tools.md`.

The repository supports TypeScript 4.9, but that compiler does not infer a type predicate from `item.type === 'function_call'`. Consequently the original `calls` variable remains `ResponseOutputItem[]` and the example fails with four `TS2339` diagnostics when reading `name`, `arguments`, and `call_id`. The current compiler infers the narrowing, which hid the compatibility problem.

The explicit public subtype fixes TypeScript 4.9 without adding an import, a cast, a new SDK type, or any runtime code. The existing discriminator check and function loop are unchanged.

## Verification

- Compiled the exact original and actual final fence against the real public SDK declarations with strict TypeScript 4.9.5 and 6.0.3. The original reproduces four errors on 4.9; the final fence has zero diagnostics on both.
- Confirmed that `calls` is `ResponseFunctionToolCall[]` on both compilers and that emitted JavaScript is byte-identical before and after.
- Executed the original and revised complete snippets through the real SDK with an in-memory synthetic transport. Mixed reasoning, message, and function-call output still produces exactly one correctly associated function result, followed by the expected final text; both requests assert the synthetic Authorization header.
- A complete explicit-schema example from `docs/structured-outputs.md` compiles unchanged on both compilers as a control.
- Canonical `./scripts/format`, `./scripts/lint`, and `git diff --check` pass.
- Adversarial self-review and independent review pass.

Only the handwritten guide changes (three added lines, one removed). Validation targets the documentation's compiler boundary; no SDK source, generated files, dependencies, runtime behavior, or test infrastructure changes. No live API calls or real credentials were used.

### DIFF
--- a/docs/tools.md
+++ b/docs/tools.md
@@ -38,7 +38,9 @@ let response = await client.responses.create({
 });
 
 while (true) {
-  const calls = response.output.filter((item) => item.type === 'function_call');
+  const calls = response.output.filter(
+    (item): item is OpenAI.Responses.ResponseFunctionToolCall => item.type === 'function_call',
+  );
 
   if (response.status !== 'completed') {
     const details = response.error ?? response.incomplete_details;


jbeckwith-oai (approved):
Reviewed the exact guide change against the supported TypeScript 4.9 boundary and public Responses types. The explicit ResponseFunctionToolCall predicate is a compile-time-only narrowing that matches the existing discriminator, requires no cast or new import, preserves emitted JavaScript and runtime behavior, and makes name, arguments, and call_id safely available in the loop on both old and current compilers. Scope stays within the handwritten guide. No in-scope blocking issues.

markstuart-oai (approved):
The explicit predicate matches the public ResponseFunctionToolCall discriminator and is the smallest fix for the documented TypeScript 4.9 floor. Independently checked the complete first example through the public SDK source entrypoint: 4.9.5 reproduces the four TS2339 diagnostics before the change and has no snippet diagnostics afterward; 6.0.3 accepts both. Both compilers infer ResponseFunctionToolCall[] after the change and emit byte-identical JavaScript before/after. The existing loop, validation, and call_id association are preserved, with no cast, helper, or additional runtime branch. No correctness, security, or structural issues found. Canonical ./scripts/lint and git diff --check pass.

PR #2650, opened 2026-09-06
fix(assistants): retain terminal message snapshots

HAYDEN-OAI:
## Summary

Make `AssistantStream.finalMessages()` retain the authoritative `thread.message.completed` or `thread.message.incomplete` message instead of the earlier creation/delta snapshot.

Previously, an ordinary completed stream could emit the correct `messageDone` event while `finalMessages()` and the terminal `textDone` snapshot still reported `status: 'in_progress'`, null completion/incompletion timestamps, and no incomplete reason. The earlier accumulated content could also retain delta-only fields instead of the final message content.

Keep the existing snapshot invariant and in-progress accumulation, but select the terminal event's message when it arrives. Terminal callbacks and final storage now share that authoritative object; previously retained creation/delta snapshots are not retroactively mutated. Document that transition on `finalMessages()`.

Before replacing the snapshot, preserve its existing message-ID aliases through the existing reservation method, matching the run-step boundary. This retains the established validation and duplicate-ID behavior without adding state, a merge helper, or a parser.

## Regression coverage

Extend the existing AssistantStream suite with complete public Message/Run fixtures, four new completed/incomplete cases, and strengthened existing controls:

- Completed and incomplete messages retain terminal status, timestamps, incomplete details, and content.
- Final storage and terminal callback contexts use the terminal object, while earlier snapshot identity/content and raw-event ordering remain unchanged.
- Aliases introduced through the previously exposed snapshot remain reserved when terminal replacement occurs; a subsequent conflicting creation is rejected before exposure.
- Existing no-terminal-message cases explicitly retain their original snapshot identity.

The two metadata regressions fail on the original source. The alias regressions pass on the original source and caught a regression in the initial patch; the implementation was tightened and independently reviewed again.

## Verification

- Full handwritten suite on Node 24.19.0: **7,749 tests across 203 files pass**, using `CI=1 OPENAI_TEST_SUITE=unit ./scripts/test`.
- Focused lifecycle, identity, and security coverage: **254 tests across six files pass**.
- Actual built public SDK, CJS and ESM: completed/incomplete SSE reproductions verify the failure before the fix and terminal snapshots after it.
- **24 independent public-SDK SSE cases** pass on Node 22.0.0, 22.22.3, and 26.7.0, covering differing accumulated/final content, retained-object identity, event order, and alias preservation. Each case makes one synthetic-authenticated request through an in-memory transport.
- Public exports, `.js` alias/class identity, methods, and declaration signatures are unchanged apart from documentation comments.
- Canonical formatting, lint, full repository TypeScript check, JS/MJS SDK build, and `git diff --check` pass.
- Adversarial self-review, independent review, tightening, and repeat review complete with no remaining findings.

Only the handwritten helper and its existing test file change. Generated code, dependencies, transport behavior, and public signatures are untouched. No live API calls or real credentials were used.

Generated API tests were not run locally: the fresh worktree lacks the pinned Steady mock source. The successful full handwritten run above is explicitly separate from that local generated-test setup limitation.

### DIFF
--- a/src/lib/AssistantStream.ts
+++ b/src/lib/AssistantStream.ts
@@ -322,7 +322,10 @@ export class AssistantStream
     return Object.values(this.#runStepSnapshots);
   }
 
-  /** Waits for successful completion and returns the final snapshot of every observed message. */
+  /**
+   * Waits for successful completion and returns the final snapshot of every observed message.
+   * Terminal message events replace accumulated snapshots without mutating earlier snapshots.
+   */
   async finalMessages(): Promise<Message[]> {
     await this.done();
 
@@ -479,6 +482,9 @@ export class AssistantStream
       case 'thread.message.delta':
       case 'thread.message.completed':
       case 'thread.message.incomplete': {
+        if (messageID !== undefined && this.#messageSnapshot) {
+          this.#reserveMessageAlias(this.#messageSnapshot, messageID);
+        }
         this.#handleMessage(stableEvent);
         if (messageID !== undefined) {
           this.#reserveMessageAlias(stableEvent.data, messageID);
@@ -918,9 +924,9 @@ export class AssistantStream
       case 'thread.message.in_progress':
       case 'thread.message.completed':
       case 'thread.message.incomplete': {
-        //No changes on other thread events
+        //Terminal events provide the authoritative message snapshot.
         if (snapshot) {
-          return [snapshot, newContent];
+          return [event.event === 'thread.message.in_progress' ? snapshot : event.data, newContent];
         }
         throw new Error('Received thread message event with no existing snapshot');
       }
--- a/tests/lib/AssistantStream.test.ts
+++ b/tests/lib/AssistantStream.test.ts
@@ -2,6 +2,8 @@ import { vi } from 'vitest';
 import { APIError, APIUserAbortError, OpenAIError } from 'openai/core/error';
 import { AssistantStream } from 'openai/lib/AssistantStream';
 import type { AssistantStreamEvent } from 'openai/resources/beta/assistants';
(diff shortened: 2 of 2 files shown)


jbeckwith-oai (approved):
Reviewed the exact head against the requested terminal-snapshot lifecycle behavior. Terminal completed and incomplete message objects now become authoritative for finalMessages and terminal callbacks, while earlier accumulated snapshots remain unmutated and prior aliases remain reserved. The focused completed, incomplete, identity, ordering, alias, and no-terminal controls cover the affected paths. I found no blocking issues.

markstuart-oai (approved):
Reviewed the full message lifecycle, callback identity, canonical-ID/alias reservations, and terminal/no-terminal controls at ad8895cf9e8817e19071777212a19b161d7d6d1f. No substantive correctness or maintainability findings. Selecting the authoritative terminal object in the existing accumulator and reserving the outgoing snapshot before replacement is a direct fix; it needs no new merge policy, state, or abstraction. Both changed files already exceeded 1,000 lines; this focused change does not introduce a new responsibility that warrants decomposition.

Independently passed 515 tests across eight focused stream/lifecycle/security suites, the canonical SDK build and lint, and git diff --check. A separate public-SDK SSE probe passed 24 cases per revision in CJS/ESM on both Node 24.20.0 and 22.22.3, reproducing stale base snapshots and verifying terminal content/metadata, callback identity, retained snapshots, and alias rejection on head. Full repository type checking is limited by missing Azure/dotenv/Express example dependencies; a broader handwritten-suite run reports failures outside this change, so I have not reproduced the claimed all-green full-suite result. Generated tests and the exact Node 22.0.0 floor were not run.

PR #2652, opened 2026-09-06
test: quote Cloudflare lifecycle preload paths

HAYDEN-OAI:
## Summary

Fix the Cloudflare credential-lifecycle test fixture when its temporary directory contains spaces.

Nine fault-injection preloads were passed as `NODE_OPTIONS="--require <raw path>"`. Node split a spaced filename into separate arguments and exited before the fixture could exercise the credential lifecycle.

- Add a shared helper that quotes the preload filename and escapes backslashes and double quotes.
- Preserve literal control characters, the existing `NODE_OPTIONS` override behavior, and preload inheritance into child processes.
- Put a deliberate space in the fixture directory prefix so the existing real-process tests permanently cover this boundary.

## Reproduction and regression

On the unchanged test, changing only `TMPDIR` from an ordinary directory to a directory containing spaces makes “never exposes a staged key through a pre-opened readable original inode” fail: the child exits 1 instead of 0. A native Node control reports `Cannot find module` for the path segment before the first space.

Applying only the new spaced fixture prefix reproduces the same failure. With the helper applied, the test passes, and all 79 credential-lifecycle cases pass as part of the full suite.

## Validation

- `./scripts/format` and `./scripts/lint`: passed with the pinned tools.
- `tsc --noEmit`: passed.
- `CI=true OPENAI_TEST_SUITE=unit ./scripts/test` on Node 24.19.0: **7,745 tests / 203 files passed**.
- Independent native Node checks loaded the exact final helper's preload in both parent and inherited child processes: **66 checks passed** across Node 22.0.0, 24.19.0, and 26.2.0. The 11 real filename cases cover spaces, quotes, backslashes, combinations, tabs, newlines, carriage returns, a control character, and Unicode.
- Deliberately adversarial reviews checked argument encoding, compatibility, inheritance, cleanup, and unchanged security assertions. JSON encoding was rejected because [Node's option parser](https://github.com/nodejs/node/blob/v24.19.0/src/node_options.cc#L2055-L2091) does not decode JSON control escapes. The final helper also passes the project's ES2020 TypeScript library target without suppressions.

The initial full-suite invocation omitted pnpm from my restricted PATH and failed three unrelated tooling fixtures; restoring the normal pinned-tool PATH passed the complete suite without further source changes. Native execution was on macOS; this does not claim to reproduce or resolve the separate Windows failure report.

## Scope

One handwritten test file only. No production, generated, dependency, compiler-configuration, assertion, timeout, credential-policy, or notifier-reset changes. No open PR currently covers this preload-path issue.

### DIFF
--- a/tests/ecosystem-cloudflare-credential-lifecycle-security.test.ts
+++ b/tests/ecosystem-cloudflare-credential-lifecycle-security.test.ts
@@ -67,7 +67,8 @@ const credentialStates: CredentialState[] = [
 ];
 
 function withFixture(run: (fixture: Fixture) => void) {
-  const directory = mkdtempSync(path.join(tmpdir(), 'openai-node-cloudflare-lifecycle-'));
+  // Exercise preload paths containing spaces regardless of the system temporary directory.
+  const directory = mkdtempSync(path.join(tmpdir(), 'openai-node-cloudflare lifecycle-'));
   const worker = path.join(directory, 'ecosystem-tests', 'cloudflare-worker');
   const bin = path.join(directory, 'bin');
   const fixture: Fixture = {
@@ -207,6 +208,10 @@ function holdOriginalCredentialDescriptor(fixture: Fixture): string {
 
 function expectNoCloudflareCredentialArtifacts(fixture: Fixture) {
   expect(readdirSync(fixture.worker).filter((name) => name.startsWith('.dev.vars.openai-'))).toEqual([]);
+}
+
+function preloadNodeOptions(preload: string) {
+  return `--require "${preload.split('\\').join('\\\\').split('"').join('\\"')}"`;
 }
 
 function runCloudflare(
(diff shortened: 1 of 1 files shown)


jbeckwith-oai (approved):
Reviewed exact head 9d25308cd382ea008024b61ede3c4d84d31146af. The helper correctly double-quotes NODE_OPTIONS preload paths and escapes the two characters interpreted by the Node option parser, while retaining literal control characters, caller override behavior, and child-process inheritance. The deliberate spaced fixture directory makes the regression part of every lifecycle run, and all nine preload sites now use the helper without changing production code or security assertions. Local format/lint/typecheck, 7,745 unit tests, 79 lifecycle cases, and the cross-version native parser matrix passed; hosted code/security reviews are clean. Remaining CI jobs are still pending and remain separate gates.

markstuart-oai (approved):
Reviewed exact head 9d25308cd382ea008024b61ede3c4d84d31146af against base 86acb953e27459ba73d6f930c6edb56369dfd002. No actionable findings. The shared helper matches Node’s NODE_OPTIONS parser, preserves literal control characters and inherited preloads, and fixes all nine call sites without changing lifecycle assertions. Keeping this encoding local to the fixture is the simplest appropriate boundary.

Independent validation on Linux / Node 24.20.0: reproduced the spaced-TMPDIR failure at the exact base; all 79 lifecycle tests pass at head; 14 real filename cases pass in parent and inherited child processes, with an unquoted failure control (29 checks); canonical lint/format checks pass. Full tsc is blocked by missing example dependencies (dotenv, @azure/identity, express) in the reused dependency tree. I did not rerun the full SDK suite or execute Windows/macOS or other Node versions.

PR #2653, opened 2026-09-06
fix(client): make retry backoff abortable via AbortSignal

### LINKED ISSUE #2580: Abort during retry backoff waits for the full delay
Aborting a request during `Retry-After` backoff does not settle the request
until the full delay expires. The retry path calls an uninterruptible
`sleep(timeoutMillis)`, even though the original request has an `AbortSignal`.

Reproduced on
https://github.com/openai/openai-node/commit/e6a74686 with Node.js 22 and a
local HTTP server that returns `429` with `retry-after-ms: 3000`:

```js
const controller = new AbortController();
const request = client.get('/retry-abort', {signal: controller.signal});

setTimeout(
  () => controller.abort(new Error('stop retrying')),
  100,
);
await request;
```

```text
pending 500 ms after abort: true
abort to rejection: 2906 ms
error: APIUserAbortError
cause preserved: true
HTTP requests: 1
```

Passing the signal through the backoff timer rejects one millisecond after the
abort in the same local-server repro. It also clears the timer, preserves the
abort reason as `cause`, and sends no retry request.

GISWLH:
## Summary
- Make `sleep` accept optional `AbortSignal`s and clear its timer on abort.
- Pass the request signal(s) through `retryRequest` backoff so abort during `Retry-After` / default backoff settles promptly instead of waiting the full delay.
- Preserve abort `cause` via `APIUserAbortError`.

Fixes #2580.

## Test plan
- [x] `vitest run --config vitest.config.mts tests/internal/utils.test.ts tests/client-behavior.test.ts` (74 passed)

### DIFF
--- a/src/client.ts
+++ b/src/client.ts
@@ -1681,7 +1681,7 @@ export class OpenAI {
       await x509Authentication.waitForRetry(timeoutMillis, x509Authentication.effectiveSignal());
     } else {
       const retrySignals =
-        requestSignal === options.signal ? [requestSignal] : [requestSignal, options.signal];
+        requestSignal === options.signal ? [requestSignal] : [options.signal, requestSignal];
       try {
         await sleep(timeoutMillis, ...retrySignals);
       } catch (error) {
--- a/src/client.ts
+++ b/src/client.ts
@@ -1090,6 +1090,8 @@ export class OpenAI {
           props.options,
           retriesRemaining,
           props.retryOfRequestLogID ?? props.requestLogID,
+          undefined,
+          props.requestSignal,
         );
         Object.assign(props, next);
       } finally {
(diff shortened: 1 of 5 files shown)


GISWLH:
Addressed the 3 P2 Codex review notes in 42067d76:

1. Prefer caller `options.signal` when selecting abort reason during retry backoff.
2. Route post-registration `aborted` probe failures through `settle()` so timers/listeners are cleaned up.
3. Remove listeners installed after synchronous cancellation during `addEventListener`.

`npx vitest run tests/internal/utils.test.ts tests/client-behavior.test.ts` — 77 passed.

HAYDEN-OAI:
Fixed in https://github.com/openai/openai-node/pull/2581 but appreciate the PR! 🙏

PR #2657, opened 2026-09-06
fix(files): use a monotonic clock for processing timeout

mukeshr06:

## Changes being requested

`files.waitForProcessing()` measures `maxWait` with `Date.now()`. A system clock correction can therefore cause a premature timeout or allow a response after the elapsed-time budget.

Use `performance.now()` in the handwritten `src/lib/file-processing.ts` helper to measure elapsed time. Preserve the existing polling interval, terminal statuses, initially terminal response behavior, and strict `>` timeout check after retrieval.

Add public-client regression cases for forward and backward clock corrections, and update the existing elapsed-time mocks. Both new cases fail with the previous implementation and pass with this change. No generated SDK files or dependencies change.

## Additional context & links

Validation on Node.js 24.19.0:
- `./scripts/test tests/lib/file-processing.test.ts`: 18 passed.
- `./scripts/lint`: passed.
- `./node_modules/.bin/tsc`: passed.
- `./scripts/build`: passed.
- `./scripts/test`: all 7,751 handwritten tests passed. Generated tests could not start locally because the pinned Steady installation failed while fetching the `@std/cli` manifest from JSR.
- `git diff --check`: passed.

AI assistance: Codex was used to prepare and test this patch.


HAYDEN-OAI (approved):
Reviewed d9c8f966. No actionable findings. The clock change preserves polling order, terminal-status handling, and the strict > timeout boundary. Source review only; tests were not run.

PR #2658, opened 2026-09-06
fix(streaming): avoid duplicate terminal tool argument events

HAYDEN-OAI:
## Summary

Record a function tool call in the existing `done_tool_calls` set after its choice has finished, after its arguments are parsed, and immediately before emitting `tool_calls.function.arguments.done`.

The stream already checks that set, but never populated it. When a choice receives metadata after its finish chunk, the accumulated `finish_reason` is still set, so every following annotation emits the same completion event again. A single tool call followed by two annotations currently produces three identical done events.

This is a supported stream shape: [Microsoft's asynchronous-filter example](https://learn.microsoft.com/en-us/azure/foundry/openai/concepts/content-streaming?view=foundry-classic) includes a blank-ID, no-delta annotation after the finish chunk and before `[DONE]`. The repository already covers that lifecycle for text content.

## Scope

- Three production lines in the handwritten `src/lib/ChatCompletionStream.ts`; no generated code, dependencies, public types, or parser changes.
- Reuse the existing per-choice set and per-request reset. The first event's timing and payload, custom tools, raw chunks, final completions, and tool execution remain unchanged.
- The completion marker is written only after successful parsing of a finished choice, before user listeners run. Final snapshot validation still runs normally.
- Preserve the existing pre-finish notifications when non-strict tool-call indices interleave. This fixes repeated terminal notifications, not the separate timing semantics of those earlier notifications.

## Regression coverage

Seven tests exercise the actual streaming helpers with synthetic SSE responses:

- No annotations versus two post-finish annotations, with two choices using the same tool index.
- Custom and strict function tools in the same completion, preserving their final payloads.
- Two streaming tool-runner requests reusing tool index zero, including callback and request counts.
- Preserve the complete existing notification sequence and final arguments when two non-strict calls interleave.

The four annotation/repeated-request regressions fail before the production change; the two no-annotation controls pass. Adversarial review caught an overly broad initial marker that suppressed later complete arguments for interleaved calls. The additional compatibility regression fails on that initial implementation and passes with the terminal-only guard.

## Verification

- Independent before/after real-HTTP loopback checks against the built public CJS and ESM entrypoints, using both clients: 44 cases on each of Node 22.0.0, 24.19.0, and 26.7.0. The final build removes terminal duplicates while preserving ordinary streams, strict parsing, multiple choices, custom calls, multi-request tool execution, and interleaved notifications.
- Twelve additional before/after public-entrypoint checks on Node 24.19.0 preserve throwing-listener errors, malformed-JSON rejection, and rejection of malformed arguments introduced by a done-event listener.
- Canonical build and full TypeScript check pass. Public `ChatCompletionStream` declarations are byte-identical in both module formats.
- Canonical full handwritten suite: 7,756 tests across 203 files pass on Node 24.19.0.
- Focused stream and post-event snapshot-validation suites: 191 tests pass on Node 22.22.3, 24.19.0, and 26.7.0.
- Canonical lint/format check passes on the final files.

No live OpenAI or Azure requests or real credentials are used. The fix was adversarially reviewed for parser failures, listener exceptions, per-choice/per-request isolation, and post-event snapshot validation.

### DIFF
--- a/src/lib/ChatCompletionStream.ts
+++ b/src/lib/ChatCompletionStream.ts
@@ -1518,6 +1518,9 @@ export class ChatCompletionStream<ParsedT = null>
         );
       }
 
+      if (choiceSnapshot.finish_reason) {
+        state.done_tool_calls.add(toolCallIndex);
+      }
       this._emit('tool_calls.function.arguments.done', {
         name: toolCallSnapshot.function.name,
         index: toolCallIndex,
--- a/tests/lib/ChatCompletionStream.test.ts
+++ b/tests/lib/ChatCompletionStream.test.ts
@@ -1,9 +1,12 @@
 import { vi } from 'vitest';
-import type OpenAI from 'openai';
+import OpenAI, { AzureOpenAI } from 'openai';
 import { APIError, OpenAIError } from 'openai/error';
 import { zodResponseFormat } from 'openai/helpers/zod';
 import { ChatCompletionStream } from 'openai/lib/ChatCompletionStream';
-import type { ChatCompletionSnapshot } from 'openai/lib/ChatCompletionStream';
+import type {
+  ChatCompletionSnapshot,
+  FunctionToolCallArgumentsDoneEvent,
+} from 'openai/lib/ChatCompletionStream';
 import { ChatCompletionStreamingRunner } from 'openai/lib/ChatCompletionStreamingRunner';
 import { makeParseableResponseFormat } from 'openai/lib/parser';
 import type { ChatCompletionTokenLogprob } from 'openai/resources';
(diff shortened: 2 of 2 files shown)


jbeckwith-oai (approved):
Reviewed exact head e342ad4d against the stated streaming-event contract. The fix narrowly populates the existing per-choice terminal set only after successful argument parsing and only once a finish reason is present, suppressing repeated post-finish annotation emissions without changing pre-finish interleaving notifications. Coverage exercises OpenAI and Azure clients, multiple choices, strict/custom tools, repeated tool-runner requests, and interleaved non-strict calls. I also ran the focused ChatCompletionStream suite (47 tests) and the full repository lint/format check locally; both pass. No blocking findings.

markstuart-oai (approved):
Reviewed both changed files and the surrounding accumulation, parsing, event dispatch, replay, and tool-runner paths at e342ad4d against d8f6be93. No substantive correctness or maintainability findings.

The existing per-choice set is the right owner: record terminal completion after successful parsing and before listeners. The finish guard preserves pre-finish index-switch notifications; marking every notification would suppress later complete interleaved arguments. Final snapshot validation remains independent, and request finalization resets the set. No additional state abstraction or parser-result cache is warranted here.

Independently reproduced three terminal events on base versus one on head through built public CJS/ESM clients. On Node 24.20.0: 598 focused tests passed; 108 base/head comparisons passed using fragmented synthetic SSE, including repeated terminal chunks, multiple/interleaved tools, parsing, cancellation, listener failures, and post-event snapshot mutation. The seven added tests give four expected failures and three passes on base. Both canonical builds and changed-file lint/format passed; stream declarations are byte-identical. Full tsc is limited by missing example-only dotenv/@azure/identity/express dependencies in the reused dependency tree; I did not rerun the author's full suite or runtime matrix.

File growth: ChatCompletionStream.ts 2550→2553 lines; its test file 1751→1963. Both were already over 1k; the production change stays in the existing finalization owner and the tests extend the existing stream fixtures.

PR #2660, opened 2026-09-06
chore: clarify Response types and remove unused handwritten code

HAYDEN-OAI:
## Summary

Clean up handwritten code behind 55 CodeQL quality findings:

- Distinguish API/HTTP response type imports from the global `Response` constructor in four test suites, addressing the ambiguity behind 48 false-positive errors. Their emitted JavaScript is unchanged.
- Remove an unreachable NDJSON condition, a repeated test-helper error branch, and unused test code while preserving existing cancellation behavior, assertions, and compile-time checks.
- Import `unittest.mock` explicitly in the handwritten budget tests.
- Remove the unused regex-validation binding with `void new RegExp(pattern);`, preserving validation and the existing catch without a suppression. The explicit discard satisfies both `no-new` and `unicorn/new-for-builtins`; a direct `RegExp(pattern);` call fails the latter pinned rule.

The 143 generated smoke-test bindings and generated reporter-test import are handled by the upstream Castiron fix, https://github.com/openai/openai/pull/1433965, and will arrive through normal regeneration. This PR does not carry those generated-file edits or change generation metadata or budget policy.

The 34 intentional/false-positive findings identified during triage remain prepared for UI dismissal; none have been dismissed through GitHub.

## Validation

- Pinned formatter/linter checks passed for all tracked files; TypeScript checking and CJS/ESM SDK build passed. Full `pnpm lint` encounters a formatting failure only in the worker-created, untracked `.symphony/prompts/attempt.md`.
- Nine affected handwritten suites: 345 tests passed; isolated global-fetch assignability test: 1 passed.
- The four Response suites emit byte-identical JavaScript; all 48 flagged expressions retain the global constructor type.
- Python custom-code suite: 63 tests run, 1 optional compiler-contract skip.
- Verified custom-code budget for this tree: 3,002 / 4,000 lines, leaving 998 lines of headroom.
- Verified that all 68 generated test files match the base revision exactly.
- Regex review follow-up: 52 Zod helper tests passed. Converted-pattern success and invalid-pattern fallback passed through CJS/ESM exports on Node 22.0.0, 24.19.0, and 26.7.0.

### DIFF
--- a/src/_vendor/zod-to-json-schema/parsers/string.ts
+++ b/src/_vendor/zod-to-json-schema/parsers/string.ts
@@ -410,8 +410,7 @@ const processRegExp = (regexOrFunction: RegExp | (() => RegExp), refs: Refs): st
   }
 
   try {
-    // oxlint-disable-next-line no-new -- Construction validates the converted pattern; only failures matter.
-    new RegExp(pattern);
+    const regexTest = new RegExp(pattern);
   } catch {
     console.warn(
       `Could not convert regex pattern at ${refs.currentPath.join(
--- a/src/_vendor/zod-to-json-schema/parsers/string.ts
+++ b/src/_vendor/zod-to-json-schema/parsers/string.ts
@@ -410,7 +410,7 @@ const processRegExp = (regexOrFunction: RegExp | (() => RegExp), refs: Refs): st
   }
 
   try {
-    const regexTest = new RegExp(pattern);
+    void new RegExp(pattern);
   } catch {
     console.warn(
       `Could not convert regex pattern at ${refs.currentPath.join(
--- a/src/_vendor/zod-to-json-schema/parsers/string.ts
+++ b/src/_vendor/zod-to-json-schema/parsers/string.ts
@@ -410,7 +410,8 @@ const processRegExp = (regexOrFunction: RegExp | (() => RegExp), refs: Refs): st
   }
 
   try {
-    const regexTest = new RegExp(pattern);
+    // oxlint-disable-next-line no-new -- Construction validates the converted pattern; only failures matter.
+    new RegExp(pattern);
   } catch {
     console.warn(
       `Could not convert regex pattern at ${refs.currentPath.join(
--- a/ecosystem-tests/bun/openai.test.ts
+++ b/ecosystem-tests/bun/openai.test.ts
@@ -23,14 +23,6 @@ async function typeTests() {
 }
 
 function expectSimilar(received: any, comparedTo: string, expectedDistance: number) {
-  const message = () =>
-    [
-      `Received: ${JSON.stringify(received)}`,
-      `Expected: ${JSON.stringify(comparedTo)}`,
-      `Expected distance: ${expectedDistance}`,
-      `Received distance: ${actualDistance}`,
-    ].join('\n');
-
   const actualDistance = distance(received, comparedTo);
   expect(actualDistance).toBeLessThan(expectedDistance);
 }
--- a/scripts/castiron/test_custom_code_budget.py
+++ b/scripts/castiron/test_custom_code_budget.py
@@ -7,9 +7,9 @@ import subprocess
 import sys
 import tempfile
 import unittest
+import unittest.mock as mock
 from pathlib import Path
 from typing import Any, cast
-from unittest import mock
 
 import custom_code_budget as budget
 import test_custom_code_report as fixtures
--- a/scripts/castiron/test_custom_code_report.py
+++ b/scripts/castiron/test_custom_code_report.py
@@ -11,9 +11,9 @@ import subprocess
 import tempfile
 import textwrap
 import unittest
-import unittest.mock as mock
 from pathlib import Path
 from typing import Any
+from unittest import mock
 
 import custom_code_report as report
 
--- a/scripts/castiron/test_custom_code_report.py
+++ b/scripts/castiron/test_custom_code_report.py
@@ -11,9 +11,9 @@ import subprocess
 import tempfile
 import textwrap
 import unittest
+import unittest.mock as mock
 from pathlib import Path
 from typing import Any
-from unittest import mock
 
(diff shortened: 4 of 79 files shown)


src/_vendor/zod-to-json-schema/parsers/string.ts
```diff
@@ -410,7 +410,8 @@ const processRegExp = (regexOrFunction: RegExp | (() => RegExp), refs: Refs): st
   }
 
   try {
-    const regexTest = new RegExp(pattern);
+    // oxlint-disable-next-line no-new -- Construction validates the converted pattern; only failures matter.
```
jbeckwith-oai: Please avoid adding this suppression. `RegExp(pattern);` performs the same validation for this string input, still throws into the existing catch on invalid patterns, and is a normal call expression that does not trigger `no-new`. That is the smaller compliant cleanup and follows the repository rule to avoid suppressions when a straightforward lint-clean form exists.

jbeckwith-oai (changes requested):
Reviewed the final 11-file handwritten cleanup against the revised 55-finding scope. The mechanical Response aliases, unreachable branches, unused test code, and generated-file revert are coherent, but one introduced lint suppression conflicts with the repository review contract because a straightforward compliant form exists. Please replace the suppression plus unused `new` expression with a direct `RegExp(pattern)` validation call. No other blocking finding found; I can re-review the updated head.

jbeckwith-oai (changes requested):
Re-reviewed exact head a111fe1d3f57ec937f7fb938bc0170f3130d8ae6. The suppression is gone, but the replacement introduces another unused binding: `const regexTest = new RegExp(pattern);`. This PR is specifically removing unused handwritten code, so please use the previously suggested `RegExp(pattern);` call expression, which preserves the validation and catch behavior without a suppression or unused local. No other in-scope blocker found in the 11-file diff. Current CI is green where complete; test matrix and ecosystem jobs are still running.

markstuart-oai (approved):
Reviewed all 10 files in a111fe1d against b0f6f20b, including the surrounding stream lifecycle and test-helper contracts. The removed branches are unreachable, the remaining assertions and constructor type check are preserved, and the four Response suites emit identical CJS/ESM JavaScript. All 48 relevant Response expressions still resolve to the global constructor. No correctness, security, or structural maintainability findings.

Validation: 322 focused tests passed, the isolated global-fetch test passed, Python ran 63 tests with one optional skip, and lint, TypeScript checking, and CJS/ESM builds passed. Local checks used cached dependencies matching the manifests/lockfile after registry installation failed; live/Bun and generated suites were not run.

On the regex discussion: the final regex file exactly matches the PR base; this head does not introduce that binding. I also reproduced the pinned unicorn/new-for-builtins rejection of RegExp(pattern). Leaving that pre-existing vendored code unchanged is reasonable here. All 68 generated-file edits are reverted, so the upstream generator cleanup can arrive through normal regeneration without blocking this handwritten PR. Thanks for narrowing the scope.

src/_vendor/zod-to-json-schema/parsers/string.ts
HAYDEN-OAI: This is not a suppression on head `bec1e9a76ee50c200641a065daa25267f79988d6`: there is no eslint directive or rule disable. The pinned `unicorn/new-for-builtins` rule rejects `RegExp(pattern);`, while `no-new` rejects an undiscarded constructor expression. `void new RegExp(pattern);` preserves constructor validation and the existing catch while explicitly discarding the value, satisfying both rules. Keeping the current lint-clean form.

jbeckwith-oai (approved):
Re-reviewed exact head bec1e9a76ee50c200641a065daa25267f79988d6. The remaining blocker is resolved: regex validation now has no lint suppression and no unused binding while preserving the existing throw/catch fallback. I rechecked the complete 11-file handwritten cleanup against the revised 55-finding scope; no other in-scope blocking findings remain, `git diff --check` is clean, and all exact-head CI is green except the final aggregate test-matrix job still queued.

markstuart-oai:
Review checkpoint: re-reviewing all 11 changed files at bec1e9a76ee50c200641a065daa25267f79988d6 against base/merge-base b0f6f20b2a90df25c1a764f13bf14ede97b2b760. I've inspected the regex conversion/fallback, NDJSON cancellation/finalization, Response type/runtime boundaries, and test-helper assertions. Local validation passed: 374 focused tests, the isolated global-fetch test, Python's 63-test suite (one optional skip), lint, TypeScript checking, and CJS/ESM builds. Independent regex probes preserve successful conversion and invalid-pattern fallback through CJS/ESM exports on Node 22.23.2 and 24.20.0. No blocker found so far; checking the live head and complete discussion before submitting the review.

markstuart-oai (approved):
Full re-review of all 11 changed files at bec1e9a7 against b0f6f20b, following my earlier approval on a111fe1d. No correctness, security, or structural maintainability findings.

The explicit `void new RegExp(pattern)` keeps construction and the existing throw/catch fallback without an unused binding or lint suppression. I independently checked the pinned rules and compared successful conversion, invalid-pattern fallback, warnings, and flag-bypass paths with the base through CJS/ESM exports on Node 22.23.2 and 24.20.0. The author’s explanation in the regex thread holds up.

The remaining cleanup preserves reachable assertions, awaited calls, cancellation/finalization, and global-fetch type checking. The four Response suites still emit identical CJS/ESM JavaScript, with all 48 relevant expressions resolving to the global constructor. All 68 generated-file reverts remain exact; Castiron can arrive separately through normal regeneration.

Validation on this head: 374 focused tests passed (including 52 Zod helper tests), isolated global-fetch test passed, Python ran 63 tests with one optional skip, and full lint, TypeScript checking, and CJS/ESM builds passed. Reused matching cached dependencies; live/Bun and generated suites and the full runtime matrix were not run. Thanks for resolving the regex cleanup without weakening the lint rules.

PR #2661, opened 2026-09-06
fix(test): use a monotonic Steady lock timeout

HAYDEN-OAI:
## Summary

- Use Node's monotonic clock for the Steady cache lock's existing 10-second acquisition budget.
- Add two regression tests through the actual cache CLI for backward and forward wall-clock corrections. The test preloads and child program are static; paths and clock direction are passed as data.
- Keep the 50-ms polling interval, cancellation, lease ownership, separate cleanup acquisition, and wall-clock file-aging timestamps unchanged. No generated files, SDK runtime code, dependency pins, or cache policy changes.

## Reproduction

On current main, hold the lifecycle lock and release it after 12 real seconds. With no clock adjustment, the wrapper rejects the command with its existing lock-timeout error. With a single 60-second backward `Date.now()` correction after the acquisition deadline is established, the same wrapper incorrectly launches the command after roughly 12.1 seconds and exits successfully.

This unaccelerated command-boundary reproduction fails before the fix on Node 22.0.0, 24.19.0, and 26.7.0. After the fix, both the normal-clock and corrected-clock cases reject late execution on all three runtimes, preserve the existing owner's lock until it is released, and leave no leases. The separate cleanup lock explains why the final process exit occurs after the fixture releases the lock; that lifecycle behavior is not changed here.

## Validation

- Both new regressions failed before the production change and passed afterward.
- New clock and existing cancellation tests: 8 passed on Node 22.0.0, 22.22.3, 24.19.0, and 26.7.0. An independent reviewer also repeated the clock tests five times successfully.
- Canonical cache expiry, active-lease, and cleanup checks passed on Node 22.0.0, 24.19.0, and 26.7.0.
- Full canonical handwritten suite on the final files: 7,800 tests passed across 205 files.
- Canonical lint/format, TypeScript checking, CJS/ESM build, and `git diff --check` passed.
- Independent adversarial review found no security, compatibility, lock-ownership, cancellation, or regression-test issues. Windows was reviewed for path/spawn compatibility but not executed locally; the generated suite will run in CI.

This is separate from #2657's file-processing timeout and retains #2636's Steady cancellation behavior.

### DIFF
--- a/scripts/steady/cache.cjs
+++ b/scripts/steady/cache.cjs
@@ -6,13 +6,14 @@ const { spawn } = require('node:child_process');
 const { randomUUID } = require('node:crypto');
 const { setInterval: interval } = require('node:timers/promises');
 const { once } = require('node:events');
+const { performance } = require('node:perf_hooks');
 
 const MAX_IDLE_MS = 30 * 24 * 60 * 60 * 1000;
 
 // Serialize lease changes and pruning, never the commands using the cache.
 async function locked(cache, action, signal) {
   const lock = path.join(cache, '.lifecycle-lock');
-  const deadline = Date.now() + 10_000;
+  const deadline = performance.now() + 10_000;
   for await (const tick of interval(50, undefined, { signal })) {
     void tick;
     try {
@@ -22,7 +23,7 @@ async function locked(cache, action, signal) {
       if (error.code !== 'EEXIST') {
         throw error;
       }
-      if (Date.now() >= deadline) {
+      if (performance.now() >= deadline) {
         throw new Error(`Steady cache is locked. If no install or mock is running, remove ${lock}.`, {
           cause: error,
         });
--- a/tests/steady-cache-clock.test.ts
+++ b/tests/steady-cache-clock.test.ts
@@ -1,0 +1,136 @@
+import { spawn } from 'node:child_process';
+import { once } from 'node:events';
+import {
+  existsSync,
+  mkdirSync,
+  mkdtempSync,
+  readFileSync,
+  readdirSync,
+  realpathSync,
+  rmSync,
+  writeFileSync,
+} from 'node:fs';
+import { tmpdir } from 'node:os';
+import path from 'node:path';
+import { vi } from 'vitest';
+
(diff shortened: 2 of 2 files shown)


jbeckwith-oai (approved):
Reviewed exact head 5f3f6964b08dd37330a87e9f68e9ce1e936bb1ab against the stated lock-timeout outcome and narrow two-file scope. The acquisition budget now uses a monotonic clock while wall-clock aging, polling, cancellation, ownership, and cleanup semantics remain unchanged. The backward/forward clock regressions exercise the real wrapper boundary with static child/preload programs and data-only paths/direction. No blocking findings. Focused tests passed 2/2 locally, changed files pass oxlint/oxfmt and `git diff --check`, no unresolved threads exist, and exact-head CI is fully green.

markstuart-oai (approved):
Reviewed both changed files and the surrounding cache acquisition, cancellation, lease, cleanup, aging, and command-spawn contracts at 5f3f6964b08dd37330a87e9f68e9ce1e936bb1ab. No substantive correctness, security, or maintainability findings. Using the existing lock helper with Node's monotonic clock is the direct fix; no extra production branching or abstraction is needed. The static CLI fixtures keep paths and clock direction as data.

Independently verified on Linux with Node 22.23.2 and 24.20.0: all eight clock/cancellation tests and canonical cache-lifetime checks pass. Both new regressions fail against the exact base. Unaccelerated 12-second contention controls reproduce late execution after a backward clock correction on base and reject it on head, preserving the other owner's lock until release and leaving no leases. Cleanup separately acquires the lock, so final exit can occur after the acquisition budget. Three extra clock-test runs with spaces/metacharacters in the temporary path also pass.

Canonical lint/format, CJS/ESM build, and diff whitespace checks pass. Validation used cached dependencies matching the lockfile after internal tooling and the pinned pnpm fresh install failed locally. Repository-wide TypeScript checking reports missing example dependencies (dotenv, @azure/identity, express). I did not run the full/generated suite, CI, Windows/macOS, exact Node 22.0.0 minimum, or Node 26; the author's broader validation claims are not independently certified here. #2636 cancellation remains intact; #2657's SDK file-processing timeout is a separate boundary.

PR #2663, opened 2026-09-06
fix(examples): respect raw streaming proxy backpressure

HAYDEN-OAI:
## Summary

- Respect `ServerResponse.write(false)` in the raw streaming proxy example by waiting for Node's `drain` event before reading another upstream chunk.
- Use the existing client-disconnect signal to cancel that wait, and suppress only the matching Node abort rejection. Preserve genuine write, drain, and upstream errors.
- Extend the existing example regression harness with backpressure, cleanup, same-tick error/disconnect races, and real HTTP transport coverage.

Only the raw example and its handwritten tests change. The Express streaming-helper example, authentication, TLS, origin checks, request handling, SDK runtime, and payload limits are unchanged.

## Reproduction

Send a synthetic 16 MiB streamed response through the actual raw-example route and built SDK to a paused local HTTP client. On main, the route consumes all 65 chunks while the client remains paused and queues roughly 14.9 MB in the response.

With this change, the same check pauses upstream consumption, peaks at 262,282 buffered bytes for 256 KiB chunks, and either delivers the complete 16 MiB after the client resumes or cancels the upstream request when the client disconnects. It leaves no drain, close, or error listeners behind. These are observed fixture results, not a new payload cap.

## Validation

- Original focused regressions: 7 failed / 38 passed before the production change.
- Independent adversarial review caught a genuine response-error/disconnect race in the first candidate. Both added error-first regression cases failed before the correction and pass afterward.
- Final four-suite example checks: 81 passed on exact Node 22.0.0, Node 24.19.0, and Node 26.2.0.
- Independent actual-route race/cancellation review: 18 checks passed across Node 22.0.0 and 24.19.0, including unrelated AbortError-shaped errors, synchronous write failures, and listener cleanup.
- Final actual-source, built-SDK HTTP proof passes on Node 22.0.0, 24.19.0, and 26.7.0: exact 16 MiB resume delivery, disconnect abort, bounded response queue, and cleanup.
- Full canonical handwritten suite: 7,811 tests passed across 205 files. Canonical formatting/lint, pinned TypeScript checking, CJS/ESM build, and `git diff --check` pass.

The generated resource suite and remaining platform integrations are left to CI.

### DIFF
--- a/tests/lib/streaming-example-disconnect-security.test.ts
+++ b/tests/lib/streaming-example-disconnect-security.test.ts
@@ -1,405 +1,15 @@
-import { timingSafeEqual } from 'node:crypto';
-import { EventEmitter, once } from 'node:events';
-import { readFileSync } from 'node:fs';
-import { createServer } from 'node:http';
-import type { IncomingMessage, ServerResponse } from 'node:http';
-import { setTimeout } from 'node:timers/promises';
-import { runInNewContext } from 'node:vm';
-
-import ts from 'typescript';
 import { describe, expect, test, vi } from 'vitest';
 
-import OpenAI, { APIUserAbortError } from '../../src/index';
-
-const examples = ['stream-to-client-express.ts', 'stream-to-client-raw.ts'] as const;
-const largeContentChunk = 'x'.repeat(256 * 1024);
-const largeContentChunkCount = 64;
-
-type Example = (typeof examples)[number];
-interface Completion {
-  choices: [{ delta: { content: string } }];
-}
-type RouteHandler = (request: unknown, response: unknown) => Promise<void>;
-
-interface ExampleRuntime {
-  apiCalls: number;
-  aborts: number;
-  generated: number;
-  cancellations: number;
-  consoleError: ReturnType<typeof vi.fn>;
-  handler?: RouteHandler;
-  signal?: AbortSignal;
-  onProvider?: () => void;
-  pendingCreate: boolean;
-  pendingNext: boolean;
-  abortError?: Error;
-  rejectCreate?: (error: Error) => void;
-  rejectNext?: (error: Error) => void;
-}
-
-function createNodeHTTPEmitter(): EventEmitter {
-  // Node HTTP requests and responses expose EventEmitter lifecycle events.
-  // oxlint-disable-next-line unicorn/prefer-event-target
-  return new EventEmitter();
-}
-
-function createRequest() {
-  return Object.assign(createNodeHTTPEmitter(), {
-    body: 'Tell me why dogs are better than cats',
-    get: vi.fn(),
-  });
-}
-
-function createResponse() {
-  const response = Object.assign(createNodeHTTPEmitter(), {
-    body: '',
-    destroyed: false,
-    headersSent: false,
-    statusCode: 200,
-    writableEnded: false,
-    writeResult: true,
-    onWrite: null as (() => void) | null,
-  });
-
-  return Object.assign(response, {
-    header: vi.fn((_name: string, _value: string) => response),
-    status: vi.fn((code: number) => {
-      response.statusCode = code;
-      return response;
-    }),
-
-    write: vi.fn((chunk: unknown) => {
-      if (response.destroyed) {
-        throw new Error('Attempted to write to a destroyed socket');
-      }
-
-      response.headersSent = true;
-      response.body += String(chunk);
-      response.onWrite?.();
-      return response.writeResult;
-    }),
-
-    end: vi.fn((chunk?: string) => {
-      if (response.destroyed) {
-        throw new Error('Attempted to end a destroyed socket');
-      }
-
-      response.body += chunk ?? '';
-      response.headersSent = true;
-      response.writableEnded = true;
-      response.emit('finish');
-      return response;
-    }),
-
-    destroy: vi.fn(() => {
-      response.destroyed = true;
-      response.emit('close');
-      return response;
-    }),
-  });
-}
-
-function completionChunks(runtime: ExampleRuntime, encoded: boolean): AsyncIterable<string | Completion> {
-  return {
-    [Symbol.asyncIterator]() {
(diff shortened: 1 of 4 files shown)


jbeckwith-oai (approved):
Reviewed exact head 88c9ff8c280c282f443eca531a3393d4006172b0.\n\nThe raw streaming example now correctly stops pulling upstream chunks after res.write() reports backpressure, resumes on drain, and cancels that wait through the existing client-disconnect signal. The abort filtering is appropriately strict: it suppresses only the exact Node AbortError/ABORT_ERR tied to this disconnect signal and preserves genuine write, drain, and upstream failures, including same-tick error/disconnect races. Listener cleanup remains covered on every tested path.\n\nVerified locally at the exact head:\n- focused disconnect/backpressure/security regression suite: 47/47 passed\n- oxlint passed for both changed files\n- oxfmt check passed\n- git diff --check passed\n\nNo unresolved review threads and no in-scope blocking findings.

tests/lib/streaming-example-disconnect-security.test.ts
```diff
@@ -519,6 +534,219 @@ test.each(examples)('%s does not rewrite an already-ended response after an erro
...
+  'the raw example handles a paused real downstream client followed by %s',
+  async (mode) => {
+    const client = await loadPublicExample(
+      'stream-to-client-raw.ts',
+      mode === 'upstream failure' ? 'large-failure' : 'large',
+    );
+    let incomingRequest: IncomingMessage | undefined;
+    let outgoing: ServerResponse | undefined;
+    let routeFinished = false;
+    const downstream = createServer(async (request, response) => {
```
markstuart-oai: [P2] Decompose the growing suite and share the HTTP lifecycle fixture

This pushes the file past 1k lines (835 → 1,063). Can we decompose this first? This new block repeats the downstream server setup already used by the upstream-failure HTTP test below: request-body collection, Express response adaptation, route-completion tracking, listen/address handling, and teardown. Meanwhile `loadPublicExample` acquires more coupled `large`/`large-failure` mode branches. Maintaining two copies makes lifecycle and cleanup fixes easy to apply inconsistently.

Please put the raw backpressure cases in a focused suite and give the shared example/HTTP harness one owner, including server setup and cleanup. Let the upstream fixture accept the content count and terminal event needed by each case instead of expanding combined mode names. Keep the pause, error-order, and listener assertions in the tests. That removes duplicated orchestration and keeps both suites readable; simply moving the copied block intact would retain the problem.

markstuart-oai (commented):
Reviewed exact head 88c9ff8c280c282f443eca531a3393d4006172b0 against freshly fetched main/base 7b5d1209090f82f56b4971976147b5c2a859452e. One maintainability issue below should be addressed before approval: the test file grows from 835 to 1,063 lines while duplicating its HTTP lifecycle fixture. The raw example grows from 215 to 242 lines.

The production boundary uses the appropriate Node primitive, events.once. Its narrow abort/cause check preserves error-first failures even when disconnect follows in the same tick; reusing the SDK's broad isAbortError or introducing competing wait promises would weaken this. I found no additional in-scope correctness/security defect. Express behavior, auth/TLS/origin/request handling, SDK runtime and payload limits are unchanged.

Independently executed:
- Canonical four focused suites: 81/81 on Node 22.23.2 and 24.20.0.
- Eight actual-route event/write probes plus two built-SDK localhost HTTP probes on each version. Upstream consumption stayed paused at 21 chunks; maximum response queue was 262,310 bytes; resume delivered all 16 MiB; disconnect canceled upstream with listener cleanup. Base failed the same queue assertion at about 11 MB buffered. These are fixture observations, not limits.
- CJS/ESM build, changed-file pinned lint/format checks, and diff whitespace check passed.

Validation limits: initial localhost checks hit the environment proxy; a process-local localhost exemption fixed them. Type-checking remains blocked by missing example dependencies (express, dotenv, @azure/identity). I did not rerun the author's 7,811-test suite, 18-check harness, exact Node 22.0.0/26 matrix, or platform integrations. The HTTP probes invoke the actual captured route with real ServerResponse and built SDK, not a full Express/TLS deployment.

tests/lib/streaming-example-disconnect-security.test.ts
HAYDEN-OAI: Addressed on exact head `912c68a18b5a`. Please re-review.

markstuart-oai (approved):
Re-reviewed all four changed files at 912c68a18b5a05a33a073f21e9bfe3f00ceb2f82 following author response 5127024766 to my review 5126983318.

The P2 is addressed: both HTTP suites now share one setup/teardown owner, and explicit content/count/terminal-event inputs replace the combined fixture modes. The disconnect suite is 434 lines (835 on base; 1,063 in the first candidate), the focused backpressure suite 198, and the shared harness 474. The raw example remains 242 lines. The split preserves the test descriptions and assertion counts while removing duplicated orchestration. No remaining actionable maintainability, correctness, or security findings.

The production drain/cancel boundary is unchanged from the reviewed implementation. Fresh checks confirm backpressure pauses upstream consumption, matching Node aborts are suppressed, genuine write/error-first failures remain visible, and listeners are cleaned up. Express, auth/TLS/origin/request handling, SDK runtime, and payload limits remain unchanged.

Fresh validation: canonical five focused suites passed 81/81 on Node 22.23.2 and 24.20.0; eight independent route/event probes and two real localhost/built-SDK probes passed on each. During the pause, consumption stayed at 21 chunks and the response queue peaked at 262,310 bytes; resume delivered 16 MiB and disconnect canceled upstream. CJS/ESM build, four-file pinned lint/format checks, and diff whitespace check passed.

Limits: tsc remains blocked by missing example dependencies (express, dotenv, @azure/identity). I did not rerun the author's full 7,811-test suite, 18-check harness, exact Node 22.0.0/26 matrix, or platform integrations. The synthetic HTTP probes exercise the actual captured route and built SDK, not a full Express/TLS deployment.

PR #2664, opened 2026-09-06
fix(test): reap Steady commands when lease creation fails

HAYDEN-OAI:
## Summary

- If persisting a newly spawned Steady command's cache lease fails, terminate the owned child and wait for it to close before releasing the lifecycle lock or reporting the original filesystem error.
- Observe child exit/error and closure before writing the lease, and do not signal a failed spawn without a PID.
- Keep successful commands outside the lifecycle lock. Lease format, cache aging/pruning policy, ordinary exit codes, signal forwarding, and existing cancellation behavior are unchanged.

Only the handwritten cache wrapper and a focused regression-test module change. No generated files, SDK runtime, dependencies, or cache-policy changes.

## Reproduction

Make the existing `.leases` directory genuinely unwritable, then launch a command through the actual cache CLI. On main, the lease write reports `EACCES`, but the child remains running with `STEADY_CACHE_LEASE=1`, no lease, and the lifecycle lock already released. If spawning also fails, the abandoned child additionally emits an uncaught `ENOENT`.

The regression uses real filesystem permissions and real child processes, with a small observational preload to record child ownership and lock state. Before the source change, both failure cases fail and the three writable controls pass.

## Validation

- Five new CLI cases pass, covering lease-write failure with a live child or failed spawn, writable normal/nonzero exits, and writable spawn errors.
- New and existing cancellation/clock regressions: 13 passed on Node 22.0.0, 24.19.0, and 26.7.0. Offline cache-lifetime checks also pass on all three.
- Independent real-process checks verify the child is closed while the lock is held, the exact original filesystem error is preserved, failed spawns produce no unhandled errors, and normal concurrent commands still overlap and retain their exit codes.
- Full canonical handwritten suite: 7,805 tests passed across 206 files.
- Canonical formatting/lint, pinned TypeScript checking, CJS/ESM build, and `git diff --check` pass.
- Independent adversarial review covers process ownership, missing PIDs, spawn errors, lock lifetime, cleanup, cancellation, and runtime-floor compatibility.

The two permission-dependent cases skip Windows and root execution; the writable controls remain enabled. Local process/permission validation was on macOS. Generated and remaining integration tests are left to CI.

### DIFF
--- a/scripts/steady/cache.cjs
+++ b/scripts/steady/cache.cjs
@@ -32,7 +32,7 @@ async function locked(cache, action, signal) {
   }
   try {
     signal?.throwIfAborted();
-    return action();
+    return await action();
   } finally {
     fs.rmdirSync(lock);
   }
@@ -105,15 +105,29 @@ async function run(cache, selected, command, args) {
   try {
     const { exited } = await locked(
       cache,
-      () => {
+      async () => {
         prune(cache, selected);
         child = spawn(command, args, { stdio: 'inherit', env: { ...process.env, STEADY_CACHE_LEASE: '1' } });
-        fs.writeFileSync(
-          lease,
-          JSON.stringify({ pids: [process.pid, child.pid].filter(Boolean), entries: selected }),
-          { flag: 'wx' },
-        );
-        return { exited: once(child, 'exit') };
+        const exit = once(child, 'exit');
+        void exit.catch(() => {
+          // Rollback reports the lease error after the child has closed.
+        });
+        const { promise: closed, resolve } = Promise.withResolvers();
+        child.once('close', resolve);
+        try {
+          fs.writeFileSync(
+            lease,
+            JSON.stringify({ pids: [process.pid, child.pid].filter(Boolean), entries: selected }),
+            { flag: 'wx' },
+          );
+        } catch (error) {
+          if (child.pid !== undefined) {
+            child.kill('SIGKILL');
+          }
+          await closed;
+          throw error;
+        }
+        return { exited: exit };
       },
       pending.signal,
     );
--- a/tests/steady-cache-lease-failure.test.ts
+++ b/tests/steady-cache-lease-failure.test.ts
@@ -1,0 +1,180 @@
+import { spawn } from 'node:child_process';
+import { once } from 'node:events';
+import {
+  accessSync,
+  chmodSync,
+  constants,
+  existsSync,
+  mkdirSync,
+  mkdtempSync,
+  readdirSync,
+  realpathSync,
+  rmSync,
+  writeFileSync,
+} from 'node:fs';
+import { tmpdir } from 'node:os';
+import path from 'node:path';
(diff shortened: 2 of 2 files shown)


jbeckwith-oai (approved):
Reviewed exact head bfde7ae9259dacefab962a619669e52fede58aeb.\n\nThe lifecycle lock now correctly remains held through asynchronous rollback, and lease-write failure terminates the directly owned child and waits for close before surfacing the original filesystem error. The separate exit rejection handler prevents failed-spawn errors from becoming unhandled while preserving the normal writable-path spawn error. The missing-PID guard, ordinary exit-code behavior, signal forwarding, lease cleanup, and successful-command lock release remain intact.\n\nI also checked the process/signal security boundary: the rollback can only signal the just-spawned ChildProcess, never a lease-provided or caller-provided PID, and it does not signal failed spawns without a PID.\n\nVerified locally at the exact head:\n- focused lease-failure, cancellation, and lock-clock suites: 13/13 passed\n- Steady cache lifetime/lease cleanup regression passed\n- oxlint passed for both changed files\n- oxfmt check passed\n- git diff --check passed\n\nNo unresolved review threads and no in-scope blocking findings. Node 22 and 24 hosted matrix jobs were still running at submission; Node 26 and the other completed checks were green.

markstuart-oai (approved):
Reviewed exact head bfde7ae9259dacefab962a619669e52fede58aeb. No actionable correctness or structural findings.

The existing lifecycle lock now awaits rollback; child exit/error and close are observed before lease persistence. On write failure only the owned ChildProcess is killed (with a PID guard), close completes under the lock, and the original filesystem error survives. The separate close observer is justified because an error-rejecting `once(child, 'close')` would not establish closure after failed spawn. Boxing the normal exit promise keeps successful commands outside the lock. Cancellation, signal forwarding, lease format, exclusive creation, pruning and aging remain unchanged. The wrapper grows 172→186 lines; the focused 180-line CLI regression module is a reasonable boundary, with no need for another lifecycle abstraction.

Fresh Linux/non-root validation:
- All 13 focused lease/cancellation/clock tests passed on Node 22.0.0, 22.23.2, 24.20.0 and 26.7.0; permission cases executed, not skipped. Explicit 22/26 runs suppressed environment-injected proxy warnings.
- Base reproduction: the two failure cases fail, three writable controls pass.
- Independent real-process probes verified original EACCES object identity, no kill without PID, close-before-unlock (including delayed close), overlapping normal commands with exit codes 0/23, and listener/lease cleanup. Offline cache-lifetime checks passed.
- Canonical lint, CJS/ESM build, and diff check passed.

Validation limits: the full handwritten run produced 7,654 passes and 151 failures across 206 files; those failures outside the focused modules were not all diagnosed or attributed to this change. Pinned TypeScript checking encountered missing example dependencies (express, @azure/identity, dotenv). I did not reproduce the reported all-green full suite or run macOS/Windows/generated integrations.

PR #2668, opened 2026-09-07
fix(examples): reject unsuccessful structured-tool responses

HAYDEN-OAI:
## Summary

- Reject explicitly unsuccessful Responses statuses before the structured-output tool example prints or consumes a function call.
- Preserve completed responses, existing missing-tool errors, and the SDK's compatibility behavior for responses that omit `status`.
- Keep the fix to a three-line guard and the existing real-HTTP regression fixture. No SDK parsing, generated code, models, dependencies, or public types change.

## Reproduction

An HTTP 200 response can have `status: "failed"`, `"incomplete"`, or `"cancelled"` and still contain a partial function call. The SDK intentionally leaves that call's `parsed_arguments` as `null`, but this example previously printed `null` and exited successfully.

The example now exits with status 1 and an explicit response-status diagnostic before printing partial output. The guard mirrors the parser's existing completion policy rather than making the optional `status` field mandatory.

## Validation

- The existing loopback HTTP fixture now has eight cases: all four original controls, omitted-status success, and three unsuccessful responses with partial JSON arguments. Before the fix, the three new failure cases exited 0; all five controls passed. Afterward, all eight pass on exact Node 22.0.0, Node 24.19.0, and Node 26.7.0.
- Independent checks execute the actual example against the freshly built public CommonJS and ESM SDK entrypoints. All 48 cases pass across those three runtimes. Before the fix, both module formats reproduce the three unsuccessful-status failures while retaining their success/error controls.
- Full canonical handwritten suite: 7,888 tests pass across 208 files.
- Canonical build, formatting/lint, TypeScript 6, and `git diff --check` pass. The SDK source and dependency files are unchanged.
- Independent adversarial review covered omitted-status compatibility, error precedence, partial-output handling, synthetic credential isolation, process cleanup, and scope.

The generated resource suite and remaining ecosystem/packaging checks run in CI.

### DIFF
--- a/examples/responses/structured-outputs-tools.ts
+++ b/examples/responses/structured-outputs-tools.ts
@@ -44,6 +44,10 @@ async function main() {
     tools: [tool],
   });
 
+  if (rsp.status && rsp.status !== 'completed') {
+    throw new Error(`Response ended with status ${rsp.status}.`);
+  }
+
   console.log(rsp);
 
   const functionCall = rsp.output.find((item) => item.type === 'function_call');
--- a/tests/lib/structured-output-tool-example.test.ts
+++ b/tests/lib/structured-output-tool-example.test.ts
@@ -72,14 +72,31 @@ async function runExample(baseURL: string) {
   }
 }
 
-const cases: { name: string; output: ResponseOutputItem[]; succeeds: boolean }[] = [
-  { name: 'function call first', output: [toolCall], succeeds: true },
-  { name: 'message before function call', output: [message, toolCall], succeeds: true },
-  { name: 'message without a function call', output: [message], succeeds: false },
-  { name: 'empty output', output: [], succeeds: false },
+const cases: {
+  name: string;
+  status: Response['status'];
+  output: ResponseOutputItem[];
+  succeeds: boolean;
+}[] = [
+  { name: 'function call first', status: 'completed', output: [toolCall], succeeds: true },
+  {
+    name: 'message before function call',
+    status: 'completed',
+    output: [message, toolCall],
+    succeeds: true,
+  },
+  { name: 'message without a function call', status: 'completed', output: [message], succeeds: false },
+  { name: 'empty output', status: 'completed', output: [], succeeds: false },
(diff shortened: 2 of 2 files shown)


jbeckwith-oai (approved):
Reviewed exact head 2a22c34963da78a63d029c262c22d277f6649db4 against the scoped goal of rejecting explicitly unsuccessful structured-tool responses while preserving completed and omitted-status compatibility. Checked error precedence, partial-output suppression, parser-policy alignment, fixture isolation/cleanup, and the absence of SDK/generated/public API changes. The focused executable example suite passed 8/8; both changed files passed oxfmt and oxlint; git diff --check passed. No unresolved review threads or in-scope blockers. Hosted Node 26, build, lint, CodeQL, and Castiron checks are green; Node 22/24 and ecosystem jobs are still running.

markstuart-oai (approved):
Reviewed exact head 2a22c34963da78a63d029c262c22d277f6649db4 against main/merge-base fdb038e21509264bcdf81741a40988dd694188b1. No actionable findings.

The direct guard belongs in this handwritten example: it stops before output is printed, prioritizes unsuccessful response status over missing-tool errors, and preserves the parser's omitted-status compatibility. The SDK continues returning inspectable unsuccessful responses. No new helper, generated change, or public contract change is warranted. The existing fixture remains small and isolates credentials, loopback ports, and child/server lifetimes.

Independently verified 256 real-HTTP checks using exact head/base example sources through freshly built public CJS/ESM entrypoints on Node 22.0.0, 22.23.2, 24.20.0, and 26.7.0. These reproduce the base's partial-tool null/exit-0 behavior and verify the head's status error/exit-1/no-stdout behavior, success and missing-tool controls, omitted/unknown statuses, and parser-error precedence. The repository fixture passed all eight cases on 22.0.0, 24.20.0, and 26.7.0; 18 parser tests, canonical build/lint, TypeScript 6.0.3, and diff checks also passed.

Dependency download was blocked; validation used a private copy of dependencies with an identical lockfile. Full handwritten/generated suites and ecosystem/packed-package checks were not rerun. No live API or shared ports were used.

PR #2669, opened 2026-09-07
fix(examples): avoid unhandled WebSocket rejections

HAYDEN-OAI:
## Summary

- Use the SDK's existing connecting-state send queue instead of a separate raw-socket opening waiter, so the example's response/error/close handlers are installed before connection events can arrive.
- Let the existing SDK `error` callback own protocol errors, rather than rejecting and removing that callback during generic event dispatch.
- Delete 48 example lines without changing the SDK, generated code, parsers, tools, models, or dependencies.

## Reproduction

For an API error frame, the SDK emits the generic `event` callback before its dedicated `error` callback. The example previously handled the generic event and removed its error listener, so the SDK then produced an unhandled rejection for the same failure. Opening handshake failures and an error frame delivered together with HTTP 101 exposed related listener gaps.

These cases already exited with status 1; the bug was the additional uncaught rejection and crash diagnostics. The fix retains normal failure exit codes, API-error precedence over a subsequent close, the six-request tool loop, and response-ID chaining. The first demo heading can now print while the connection is opening.

## Validation

- Extend the existing real-WebSocket fixture to ten cases. Seven regressions fail before the source change, while the three original successful/early-close controls pass. All ten pass afterward on exact Node 22.0.0, Node 24.19.0, and Node 26.7.0.
- The fixture observes `uncaughtExceptionMonitor` without handling exceptions or changing Node's default exit behavior. It covers flat/nested API errors, immediate closes, rejected upgrades, and HTTP 101 coalesced with an initial error frame.
- Independent checks execute the actual example against the freshly built public SDK: all 13 scenarios pass on each of those three runtimes, with zero uncaught failures. Before the fix, eight scenarios exhibited unhandled rejections; normal request counts and exit-code controls still passed.
- Full canonical handwritten suite: 7,890 tests pass across 208 files.
- Canonical build, formatting/lint, TypeScript 6, and whitespace checks pass. Independent adversarial review found no remaining issues.

### DIFF
--- a/examples/responses/websocket.ts
+++ b/examples/responses/websocket.ts
@@ -74,14 +74,6 @@ type RunTurnResult = {
   responseID: string;
 };
 
-type OpenableSocket = {
-  readyState: number;
-  on(event: 'open' | 'close', listener: () => void): void;
-  on(event: 'error', listener: (err: Error) => void): void;
-  off(event: 'open' | 'close', listener: () => void): void;
-  off(event: 'error', listener: (err: Error) => void): void;
-};
-
 type CLIArgs = {
   model: string;
   useBetaHeader: boolean;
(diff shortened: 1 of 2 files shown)


markstuart-oai (approved):
Reviewed both changed files and the SDK send queue, generic/dedicated error dispatch, adapter/emitter lifecycle, and tool-loop callers. No actionable correctness, security, or maintainability findings. This is a useful structural simplification: the canonical queue removes the raw opening waiter, and the dedicated SDK error callback retains ownership until dispatch. Printing the first heading while connecting is reasonable.

Independent validation: 124 focused tests passed; the ten-case example fixture also passed on Node 22.0.0 and 26.7.0 (primary run: 24.20.0). Running the new fixture against base reproduced seven failures with all three controls passing. A separately written real-loopback fixture exercised the actual transpiled example through freshly built public CJS and ESM entrypoints: 78 head runs passed across those three runtimes, with zero uncaught errors, one failure report per failing scenario, preserved error-over-close precedence, six-request response-ID/tool-output chaining, and owned child/server cleanup. Another 78 base runs reproduced the seven targeted unhandled-rejection scenarios. This includes rejected upgrades and a single transport write containing HTTP 101 plus the initial error frame.

Canonical build/lint, TypeScript 6, and whitespace checks passed. Registry installation was unavailable, so validation used a private copy of cached dependencies with the matching lockfile. I did not rerun the full handwritten/generated or ecosystem suites.

jbeckwith-oai (approved):
Reviewed exact head 09444de4. Removing the example-local opening waiter and manual error branch correctly leaves socket-open queuing and API/transport error ownership with the SDK connection. The regression covers rejected upgrades, nested/flat API errors, same-write handshake/error delivery, close precedence, multi-turn chaining, and uncaught rejection monitoring. CI is green and there are no unresolved threads. No in-scope findings.

PR #2670, opened 2026-09-07
fix(format): ignore missing literal file-list paths

HAYDEN-OAI:
## Summary

- Filter the line-based `fast-format` input to existing literal files before passing the shared list to Oxlint and Oxfmt.
- Keep the existing handling of spaces, tabs, quotes, leading dashes, CRLF, and a final line without a newline. Existing filenames containing glob characters still work as literal filenames.
- Change only the handwritten wrapper and its existing regression fixture. No generated code, SDK behavior, dependencies, or lockfiles change.

## Reproduction

With `unrelated.ts` present and `[u]nrelated.ts` absent, a file list containing only `[u]nrelated.ts` causes both native tools to interpret the missing filename as a glob and modify `unrelated.ts`. Missing `*.ts` and brace-looking filenames similarly modify files that were never listed. The commands still exit successfully.

The wrapper now checks each literal path once at the shared input boundary. Missing paths are skipped before either tool sees them; no glob parser or per-tool workaround is added. Regular-file symlinks retain their behavior. Directories and directory symlinks are ignored, consistent with the documented one-file-path-per-line input.

## Validation

- Extend the existing native-CLI fixture from 11 to 19 cases. Against the unchanged base wrapper, four regressions fail while 15 controls pass. All 19 pass afterward on exact Node 22.0.0, Node 24.19.0, and Node 26.7.0.
- Independently execute the actual wrapper with the pinned Oxlint/Oxfmt binaries: three missing glob-looking paths modify unlisted files before the change; all eight literal/deleted-path cases pass afterward.
- Independent adversarial review compares 17 baseline/current scenarios, including broken symlinks, existing literal metacharacters, backslashes, leading dashes, spaces, tabs, regular-file symlinks, CRLF, and missing final newlines, using Bash 3.2 and the pinned native tools. No remaining findings.
- Full canonical handwritten suite: **7,892 tests pass across 208 files**. An initial run hit local `ENOSPC`; the complete rerun passed after disk space was restored.
- Canonical formatting/lint, pinned TypeScript 6, Bash syntax, and whitespace checks pass.

### DIFF
--- a/scripts/fast-format
+++ b/scripts/fast-format
@@ -29,7 +29,13 @@ node ./scripts/lint-generated.cjs --fix --file-list "$FILE_LIST"
 
 echo "==> Running oxlint --fix"
 # Match the generated linter's line-based input, without letting xargs split paths.
-FILES="$(sed -e $'s/\r$//' -e '/^$/d' "$FILE_LIST")"
+# Only pass existing literal files; missing paths can be interpreted as globs.
+FILES="$(while IFS= read -r file || [ -n "$file" ]; do
+    file="${file%$'\r'}"
+    if [ -f "$file" ]; then
+        printf '%s\n' "$file"
+    fi
+done < "$FILE_LIST")"
 LINT_FILES="$(printf '%s\n' "$FILES" | grep -E '\.(ts|tsx|mts|cts|js|jsx|mjs|cjs)$' || true)"
 if [ -n "$LINT_FILES" ]; then
    printf '%s\n' "$LINT_FILES" | tr '\n' '\0' | xargs -0 ./node_modules/.bin/oxlint --fix --no-error-on-unmatched-pattern --
--- a/tests/fast-format-file-list.test.ts
+++ b/tests/fast-format-file-list.test.ts
@@ -54,18 +54,25 @@ describeBash('fast-format file lists', () => {
     withFormatFixture((root) => fastFormat(root, list));
   });
 
-  test.each(['selected.ts', 'with spaces.ts', "with'quote.ts", 'with"quote.ts', '-leading-dash.ts'])(
-    'formats the exact listed path %s',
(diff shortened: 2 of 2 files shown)


markstuart-oai (approved):
Reviewed 36a245dc227660dd21ff5d681e91780df74686d6 against fdb038e21509264bcdf81741a40988dd694188b1. No actionable correctness or maintainability findings. Establishing the existing-file invariant once in the shared input list is the right boundary; it avoids a glob parser and per-tool special cases while preserving literal path arguments.

Independent validation: the 19-case native regression suite passes on Node 22.0.0, 24.20.0 and 26.7.0; the same suite with the base wrapper produces four failures and 15 passing controls. A separate 30-scenario base/head matrix passes all head cases, covering metacharacters, whitespace/quotes/backslashes/dashes, CRLF/EOF, regular and broken symlinks, directories, mixed/empty input and formatter-only JSON. Canonical lint/format, Bash 5.2 syntax and diff whitespace checks pass.

Validation qualifications: isolated pinned Oxfmt 0.62.0 reproduces unlisted-file edits for missing glob-looking paths; pinned Oxlint 1.79.0 did not reproduce those edits here, although its literal-file fix controls pass. Ten related generated-lint/config tests pass; one package-command fixture is blocked by pnpm's offline-store SQLite initialization. I did not run the full SDK suite or an actual Bash 3.2 binary.

jbeckwith-oai (approved):
Reviewed exact head 36a245dc. Filtering the shared line-based input to existing regular files before invoking Oxfmt/Oxlint prevents missing literal paths from being reinterpreted as globs while preserving spaces, tabs, quotes, leading dashes, metacharacters, CRLF, and mixed lists. The focused regressions and full CI matrix are green, with no unresolved threads. No in-scope findings.

PR #2672, opened 2026-09-07
fix(examples): avoid extra requests for empty tool calls

HAYDEN-OAI:
## Summary

- Stop the raw tool-calling example when the collected tool-call array is empty, using the same length check as the SDK's tool runner.
- Keep the check after the stream is fully reduced. The reducer, model, streaming delay, output, SDK, generated code, and dependencies are unchanged.
- Extend the existing redirected-output fixture to cover omitted and empty tool-call arrays, both for direct answers and after a valid tool round trip.

## Reproduction

A completed assistant response with `finish_reason: 'stop'` and `tool_calls: []` leaves an empty array in the example's collected message. Because an empty array is truthy, the old guard falls through, executes zero tools, and starts another completion request.

The actual example sends an unwanted second POST for a direct answer, or a third POST after one valid tool round trip. Omitted tool calls correctly stop after one or two requests. The reproduction server rejects the first unwanted request with a non-retryable HTTP 400, so the failure is bounded and does not rely on a process timeout. Repeated empty-array answers could otherwise keep the request loop running.

## Validation

- Expand the existing executable fixture from three to six cases. Against the unchanged example, the two empty-array regressions fail their exact request-count assertions while four controls pass. All six pass afterward on exact Node 22.0.0, Node 24.19.0, and Node 26.7.0.
- Execute the actual example against the built public SDK in four scenarios on each runtime: all 12 checks pass with exact request counts `1, 1, 2, 2`, natural successful exits, and final text printed once.
- Fragmented function arguments, tool-call IDs, and tool results remain unchanged; both legacy function examples retain their redirected-output controls.
- Full canonical handwritten suite: **7,887 tests pass across 208 files**. CJS/ESM build and imports, formatting/lint, pinned TypeScript 6, and whitespace checks pass.
- Independent adversarial review found no remaining issues. The guard remains after complete stream reduction, preserving non-empty tool calls and avoiding changes to the custom reducer.

### DIFF
--- a/examples/chat-completions/tool-calls-stream.ts
+++ b/examples/chat-completions/tool-calls-stream.ts
@@ -147,7 +147,7 @@ async function main() {
     messages.push(message);
 
     // If there are no tool calls, we're done and can exit this loop
-    if (!message.tool_calls) {
+    if (!message.tool_calls?.length) {
       return;
     }
 
--- a/tests/lib/streaming-example-nontty.test.ts
+++ b/tests/lib/streaming-example-nontty.test.ts
@@ -10,9 +10,17 @@ import type {
 import { expect, test } from 'vitest';
 
 const examples = ['function-call-stream.ts', 'function-call-stream-raw.ts', 'tool-calls-stream.ts'] as const;
+const cases = [
+  ...examples.map((filename) => ({ filename, hasToolCall: true, emptyToolCalls: false })),
+  { filename: 'tool-calls-stream.ts', hasToolCall: true, emptyToolCalls: true },
+  { filename: 'tool-calls-stream.ts', hasToolCall: false, emptyToolCalls: false },
+  { filename: 'tool-calls-stream.ts', hasToolCall: false, emptyToolCalls: true },
+];
 
-test.each(examples)('%s completes a tool round trip with redirected stdout', async (filename) => {
+test.each(cases)('$filename (tool=$hasToolCall, empty=$emptyToolCalls)', async (scenario) => {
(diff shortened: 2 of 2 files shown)


markstuart-oai (approved):
Independently reproduced the base example's unwanted second/third requests for completed empty tool-call arrays. This guard matches the SDK runner and stays after full stream reduction; it is the simplest correct fix without changing the reducer or introducing another abstraction. No correctness, security, or structural findings.

Verified the actual example through the built public SDK on Node 22.0.0, 24.20.0, and 26.7.0: omitted/empty calls terminate directly and after a tool round trip with exact counts 1/1/2/2, natural exits, complete output, and preserved fragmented arguments/IDs/results. Five additional error-path checks preserve failures without extra requests. The six-case regression fixture passes on all three runtimes, including both legacy controls; its HTTP 400 bounds unwanted requests without relying on timeout. All 41 runner tests, CJS/ESM build/import checks, TypeScript 6, lint, and diff checks pass.

Validation limit: used copied local dependencies after registry installation failed; stopped the broader handwritten run after failures outside this patch. No full-suite or generated/ecosystem-suite pass is claimed.

jbeckwith-oai (approved):
Reviewed exact head 0946565a. Treating an empty tool_calls array as terminal matches the SDK runner and prevents unintended follow-up completions while preserving real tool-call rounds. The fixture bounds extra requests with an explicit 400 and covers omitted/empty calls both initially and after a tool round. CI is green and there are no unresolved threads. No in-scope findings.

PR #2673, opened 2026-09-07
fix(ecosystem): preserve retry delays in parallel workers

HAYDEN-OAI:
## Summary

Forward `--retryDelay` to parallel ecosystem-test workers alongside `--retry`. The parent already parses this option, but workers previously reverted to the 1,000 ms default for ordinary project retries.

This is one production-line addition. It preserves explicit `0`, custom delays, and the existing default without changing retry counts, option parsing, credentials, cleanup, or SDK code.

## Regression coverage

Strengthen the existing offline local-package fixture so each project fails its first typecheck and then succeeds. It runs the actual CLI, actual pnpm worker processes, and real local npm installations, and still verifies the selected package version.

- Before the fix: `--jobs=2 --retryDelay=25` and `--parallel --retryDelay=0` both report 1,000 ms project retries. The sequential custom-delay and parallel default-delay controls pass: two failures, two passes.
- After the fix: all four cases use the expected delay and complete successfully.

## Validation

- Four focused cases pass on Node 22.22.3, 24.19.0, and 26.7.0.
- Full handwritten suite: 7,885 tests across 208 files pass on Node 24.19.0.
- Canonical formatting and lint checks, TypeScript 6, and the typechecked CLI `--help` entrypoint pass.
- An independent before/after CLI probe confirms the actual worker arguments and exactly two project attempts in five scenarios, including custom, zero, default, sequential, and explicit-worker-count controls.
- Independent adversarial review checked zero/default preservation, both worker-selection modes, argument-array safety, retry-count behavior, and fixture scope; no further changes were needed.

The first full run encountered an unrelated local SSH-signing-agent failure in a temporary Git fixture. The focused fixture and full suite passed with commit signing disabled only for synthetic test commits in that command's environment; repository signing configuration is unchanged.

Exact Node 22.0.0 cannot load this existing CLI's pinned ESM yargs dependency on either main or this branch. Repository tooling targets Node 24; this change does not alter the consumer SDK's Node 22.0.0 floor.

### DIFF
--- a/ecosystem-tests/cli.ts
+++ b/ecosystem-tests/cli.ts
@@ -870,6 +870,7 @@ async function main() {
                       '--skip-pack',
                       '--noCleanup',
                       `--retry=${args.retry}`,
+                      `--retryDelay=${args.retryDelay}`,
                       ...(args.live ? ['--live'] : []),
                       ...(args.verbose ? ['--verbose'] : []),
                       ...(args.deploy ? ['--deploy'] : []),
--- a/tests/ecosystem-cli.test.ts
+++ b/tests/ecosystem-cli.test.ts
@@ -269,10 +269,16 @@ describe('ecosystem test CLI', () => {
   });
 
   test.each([
-    { name: 'sequential', options: [], packageOption: '--fromNpm' },
-    { name: 'explicit worker count', options: ['--jobs=2'], packageOption: '--fromNpm' },
-    { name: 'parallel', options: ['--parallel'], packageOption: '--from-npm' },
-  ])('installs the selected local package in $name mode', ({ options, packageOption }) => {
+    { name: 'sequential', options: [], packageOption: '--fromNpm', retryDelay: 7 },
+    { name: 'explicit worker count', options: ['--jobs=2'], packageOption: '--fromNpm', retryDelay: 25 },
+    { name: 'parallel', options: ['--parallel'], packageOption: '--from-npm', retryDelay: 0 },
+    {
+      name: 'default-delay parallel',
+      options: ['--parallel'],
+      packageOption: '--from-npm',
+      retryDelay: undefined,
+    },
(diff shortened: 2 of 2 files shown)


markstuart-oai (approved):
Reviewed the full diff against fdb038e21509264bcdf81741a40988dd694188b1 and traced parsed retryDelay through both worker-selection modes into the project retry loop. The direct argument-array addition preserves explicit zero, custom/default delays, retry budgets, and existing credential/cleanup behavior without adding an abstraction. The existing fixture remains the right home: CLI 1238→1239 lines, tests 600→618; no new 1k-line crossing or substantive structural regression.

Independently reproduced two failures/two passing controls on the base using the strengthened fixture, then all four passing on head. All 71 ecosystem CLI tests pass on Node 24.20.0; the four focused cases also pass with Node 22.22.3 and 26.7.0. A separate offline real-worker/local-install probe passed seven scenarios, checking package versions, install counts, and exactly 1/2/3 project attempts for the relevant retry budgets. Canonical lint/format checks, TypeScript 6 no-emit checking, and typechecked CLI help pass.

Validation limits: these are synthetic project commands (including the typecheck script), not full ecosystem compiler suites or wall-clock timing assertions; the complete SDK suite was not rerun. Exact Node 22.0.0 fails CLI bootstrap with the same existing yargs ERR_REQUIRE_ESM on base and head, independently of the consumer SDK support floor. Reused matching installed dependencies; no live API calls, source edits, or signing/auth changes. No substantive findings.

jbeckwith-oai (approved):
Reviewed exact head 2a4b8011. Forwarding retryDelay into parallel worker arguments is the missing link in the existing option flow and preserves explicit zero, custom values, the 1000 ms default, retry counts, and sequential behavior. Focused real-worker coverage and the full CI matrix are green, with no unresolved threads. No in-scope findings.

PR #2674, opened 2026-09-07
test: isolate line-ending fixture from commit signing

HAYDEN-OAI:
## Summary

Keep the disposable Git commits in the CRLF checkout test independent of a contributor's commit-signing setup.

The fixture already supplies a synthetic Git identity, but inherits `commit.gpgsign=true`. An unavailable SSH/GPG signer can therefore fail the test before it exercises formatting. Set `commit.gpgsign=false` in the temporary fixture repository's local configuration.

This is two added lines in one handwritten test file, including the explanatory comment. All line-ending assertions, commits, cleanup, production code, and real-repository signing settings remain unchanged.

## Validation

- Deterministic before/after reproduction through the canonical test command with a private synthetic global configuration: signing enabled, `/usr/bin/false` as the signer, and a synthetic key identifier. Main fails at the first fixture commit; the fix passes under the identical configuration without using a real signer.
- The focused regression passes on Node 22.22.3, 24.19.0, and 26.7.0.
- Full handwritten suite: 7,884 tests across 208 files pass on Node 24.19.0 with ordinary globally enabled signing and no command-level signing override.
- Canonical formatting, lint, TypeScript 6, and diff checks pass.
- Independent native-Git checks verify the fixture-local override and successful disposable commits while the synthetic global configuration and actual source repository's local signing configuration remain unchanged. Adversarial review found no further issue.

The exact Node 22.0.0 probe gets past the commits but encounters the existing Oxfmt TypeScript-config loader requirement. Repository tooling continues to target Node 24; no SDK runtime requirement changes.

### DIFF
--- a/tests/oxlint-config.test.ts
+++ b/tests/oxlint-config.test.ts
@@ -49,6 +49,8 @@ test('formats an existing CRLF checkout after the LF policy is pulled', () => {
     runGit(['init', '--quiet']);
     runGit(['config', 'user.email', 'line-endings@example.com']);
     runGit(['config', 'user.name', 'Line Endings Test']);
+    // Keep synthetic fixture commits independent of the developer's signing agent.
+    runGit(['config', 'commit.gpgsign', 'false']);
     runGit(['config', 'core.autocrlf', 'true']);
     runGit(['config', 'core.safecrlf', 'false']);
 


markstuart-oai (approved):
Reviewed the complete test and runner/config propagation. This is the direct fixture-level fix: runGit uses the newly initialized temporary repository, so commit.gpgsign is written locally; source/global settings and hook configuration are unchanged. Both commits, CRLF-to-LF assertions, clean-index check, and finally cleanup remain intact. File growth is 472 → 474 lines, with no structural regression or warranted extra abstraction.

Independent validation: the canonical focused test fails at the first commit on main with a harmless failing synthetic signer and passes at this head on Node 22.23.2 and 24.20.0. Native Git checks confirmed local config origin, two disposable commits, unchanged global signing/hook settings, and cleanup; source/user/review-store config hashes stayed unchanged. Changed-file formatting, lint, and diff checks pass.

Limits: the full changed-file run passed 10 tests; its separate public-package-command test was blocked by pnpm's store database access error, including after a private-store retry. Exact Node 22.0.0 gets past the commits but hits the existing Oxfmt TypeScript-config loader requirement. Repository tooling targets Node 24; this does not change the SDK consumer floor. I did not run the full SDK suite or live API tests.

jbeckwith-oai (approved):
Reviewed exact head 5aa0435e51567ffabccd9351b95c517cff0f3742 against the scoped goal of isolating the disposable CRLF fixture repository from contributor commit-signing settings. The local `commit.gpgsign=false` is set after fixture initialization and before both commits, affects only the temporary repository, and leaves the line-ending assertions and source-repository configuration unchanged. Reproduced under a synthetic global SSH signing policy using `/usr/bin/false`; the focused regression passed 1/1. The changed file passed oxfmt and oxlint, and git diff --check passed. No unresolved threads or in-scope blockers; hosted build, lint, CodeQL, Castiron, and breaking-change checks are green while Node/ecosystem jobs continue.

PR #2687, opened 2026-09-07
fix(responses): validate indexed response collections

HAYDEN-OAI:
Response accumulation now verifies that an indexed collection is an array before trusting its length. This prevents malformed shell command objects from being accepted as command arrays and driving output allocation. Existing index rules and support for large valid arrays remain unchanged.

Validation:
- Four public SSE/replay cases covering shell-output delta and done events fail before the fix and pass after it.
- Four controls preserve genuine 100,001-command arrays.
- All 246 Responses streaming and accumulator tests pass on Node 24.20.0 and the minimum supported Node 22.0.0.
- Repository lint, TypeScript checking, and production CJS/ESM build pass.
- Independent review completed.

Generated API-resource tests were not run: the pinned mock-server setup stalled fetching JSR dependency metadata.

### DIFF
--- a/src/internal/responses/response-accumulator.ts
+++ b/src/internal/responses/response-accumulator.ts
@@ -153,6 +153,7 @@ function validateArrayIndex(
   allowAppend = false,
 ): void {
   if (
+    !Array.isArray(collection) ||
     !Number.isSafeInteger(index) ||
     index < 0 ||
     index > collection.length ||
--- a/tests/lib/ResponseStream.test.ts
+++ b/tests/lib/ResponseStream.test.ts
@@ -383,6 +383,88 @@ describe('.stream()', () => {
     );
   });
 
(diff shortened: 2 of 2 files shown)


markstuart-oai (approved):
Reviewed both changed files and the shared validator's output/content/annotation/command callers against the actual main merge-base. The array check establishes the invariant in the existing owner before trusting indexed collections, preserves valid append/index behavior and large arrays, and needs no extra abstraction. No actionable correctness or structural findings.

Independently ran 365 streaming/accumulator tests on Node 24.20.0 and 22.0.0. Loading the exact pre-fix accumulator made all four new rejection cases fail while the four large-array controls passed. Repository lint and production CJS/ESM build passed. Validation used existing cached tooling because the configured registry was unreachable; full tsc reported missing example dependencies (Azure identity, dotenv, Express). Generated API-resource tests were not run.

PR #2688, opened 2026-09-07
fix: serialize tool-call IDs once per completion

HAYDEN-OAI:
Streaming tool runners now include the tool-call ID mapping only with the first result of each completion. Replay consumes that mapping when it finalizes the completion; repeating it for later results inflated serialized output and could exhaust the existing iterator buffer.

The first mapping remains intact so IDs generated for empty provider IDs survive replay, and each new completion refreshes the mapping. The relay format is unchanged.

Validation:
- Three public SSE/serialization/replay regressions fail before the change: repeated maps for 32 calls, a buffer failure for 64 calls, and repeated mappings across two completions with generated IDs.
- 93 focused tests pass on Node 24.20.0 and minimum Node 22.0.0.
- Canonical lint, TypeScript checking, build, and whitespace checks pass.
- Local generated-test setup could not finish fetching JSR metadata; CI runs the full generated suite.

### DIFF
--- a/src/lib/ChatCompletionStreamingRunner.ts
+++ b/src/lib/ChatCompletionStreamingRunner.ts
@@ -96,6 +96,7 @@ export class ChatCompletionStreamingRunner<ParsedT = null>
               throw new OpenAIError('cannot serialize a tool message before receiving any chunks');
             }
             push(makeChatCompletionReadableStreamMessageChunk(lastChunk, message, toolCallIds));
+            toolCallIds = undefined;
           }
         };
         this.on('chunk', onChunk);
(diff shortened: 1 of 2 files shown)


markstuart-oai (approved):
Reviewed 904a2e18db3a0ff93a549b1a8ebd7f86839de5f9 against the actual main-base PR diff. No actionable correctness, security, or maintainability findings.

The reset matches the existing replay lifecycle: the first tool-result record restores IDs and finalizes the preceding completion, while each subsequent assistant message refreshes the mapping. This removes redundant serialization without new branches, helpers, or relay-format changes. Reviewed both changed files, normalization/result emission, replay, iterator buffering, legacy-consumer coverage, and prior discussion.

Independently verified all three new regressions fail against base c581c17f1edb141a9ca7838a6ecb2e970e16e924, including the 64-call buffer failure. All 110 focused tests pass on Node 24.20.0 and minimum Node 22.0.0. Canonical lint, TypeScript checking, build (including CJS/ESM smoke checks), and diff whitespace checks pass. Validation reused copied local dependencies; a fresh dependency install and generated suite were not run.

PR #2690, opened 2026-09-07
fix: isolate realtime connection credentials

HAYDEN-OAI:
Concurrent Realtime factories now retain the API credential returned for each connection. Stable and beta transports use that captured value for WebSocket authentication and native browser eligibility, so a later provider response or URL callback cannot replace a connection’s credential.

`_callApiKey()` keeps its boolean return and `client.apiKey` update. An optional internal capture callback supplies the connection-local value without replacing the client or bypassing its URL/security hooks. Legacy overrides that ignore capture retain their existing shared-property behavior; overrides must forward the callback to obtain the same concurrency isolation.

Validation:
- 20 regressions fail before the change across stable/beta × native/`ws` × OpenAI/Azure, browser eligibility, and custom URL callbacks.
- Controls cover static and Azure auth modes, provider failures, legacy/forwarding overrides, and the no-argument credential-hook contract.
- 481 tests across eight transport/authentication suites pass on Node 24.20.0 and exact minimum Node 22.0.0, including 54 new cases.
- Canonical lint, TypeScript checking, build/import checks, and published-source TypeScript 4.9/6 checks pass.

### DIFF
--- a/src/beta/realtime/websocket.ts
+++ b/src/beta/realtime/websocket.ts
@@ -1,3 +1,4 @@
+import { getRealtimeAPIKey, resolveRealtimeAPIKey } from '../../internal/realtime-credentials';
 import type { AzureOpenAI } from '../../index';
 import { assertBedrockWebSocketOrigin } from '../../internal/bedrock';
 import { OpenAI } from '../../index';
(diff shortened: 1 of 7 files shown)


markstuart-oai (approved):
Reviewed all seven changed files at 8818e9df89ca0e633dfbb9af22bd17437b1f8b1c against the actual PR merge base, including stable/beta native and ws transports, Azure authentication, browser eligibility, URL callbacks, and credential-hook compatibility. No actionable correctness, security, or structural findings. The connection-local capture preserves client identity and the existing boolean hook contract without introducing shared synchronization; the documented legacy-override limitation is explicit. No file crosses 1,000 lines.

Independently verified: 481 tests across eight relevant suites pass on Node 24.20.0 and exact minimum 22.0.0; all 54 new cases also pass on Node 26.7.0. Running the new suite against merge-base source produces 28 failures and 26 passes. Canonical lint, CJS/ESM build, scoped source/test typechecking, published-source TypeScript 4.9/6 checks, and all four built Realtime entrypoint imports in CJS/ESM on Node 22.0.0 pass. Full-project typechecking is limited by missing example-only dotenv, @azure/identity, and express dependencies in the reused dependency tree. No live-service or real-browser validation performed.

PR #2692, opened 2026-09-07
fix: bound partial number recovery to its token

HAYDEN-OAI:
Partial number recovery now confines exponent lookup to the current numeric token. This removes repeated scans of the full input during malformed structured-stream recovery while preserving valid JSON and existing partial-exponent behavior.

Initial local validation at `afdd3709`:
- Two public SSE regressions fail before the change on deterministic search-work counts; both content and strict-tool paths retain their sanitized final error.
- Controls cover incomplete/terminated exponents, malformed partial arrays, and valid 1 MiB structured content and tool arguments.
- 210 focused tests pass on Node 24.20.0 and exact minimum Node 22.0.0.
- Canonical lint, TypeScript checking, build/import checks, and eight built CJS/ESM parser checks on Node 22.0.0 pass.

Follow-up test commits add two exponent-recovery controls; the production fix is unchanged. CI passes at `e795519b`.

### DIFF
--- a/src/_vendor/partial-json-parser/parser.ts
+++ b/src/_vendor/partial-json-parser/parser.ts
@@ -248,14 +248,15 @@ const _parseJSON = (jsonString: string, allow: number) => {
       markPartialJSON('Unterminated number literal');
     }
 
+    const number = jsonString.substring(start, index);
     try {
-      return JSON.parse(jsonString.substring(start, index));
+      return JSON.parse(number);
     } catch {
-      if (jsonString.substring(start, index) === '-' && Allow.NUM & allow) {
+      if (number === '-' && Allow.NUM & allow) {
         markPartialJSON("Not sure what '-' is");
       }
       try {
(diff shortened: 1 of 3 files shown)


markstuart-oai (approved):
Reviewed all three changed files at e795519b4f7dfbf4b1b55c478c9e32ae7dbe63af and the surrounding partial-parser and content/strict-tool streaming contracts. The token-local slice is the direct fix: it removes repeated extraction and keeps recovery within the existing token boundary without new branches or abstractions. No actionable correctness, security, or maintainability findings.

Independent validation: all 163 tests in the two affected suites pass on Node 24.20.0 and exact minimum Node 22.0.0, including exponent controls and valid 1 MiB content/tool arguments. Substituting the merge-base parser in memory makes both new public SSE regressions fail specifically at the search-work assertion (104,888,000 versus 1,048,880), while their sanitized-error assertions still pass. Canonical lint passes. Broader TypeScript checking is limited by missing example dependencies (@azure/identity, dotenv, express); full build/package checks were not rerun.

PR #2693, opened 2026-09-07
fix: validate partial image indices in streaming example

HAYDEN-OAI:
The streaming-image example validates each partial image index before constructing a filename or logging it. Indices must be integers within the requested partial-image count; the same constant now supplies the request, display, and validation range.

Validation:
- Ten malformed-index regressions fail before the change, using the executable example and a loopback SSE server.
- Controls preserve first/last partial images and all seven existing success, error, and early-final scenarios.
- All 19 executable-example tests pass on Node 24.20.0 and exact minimum Node 22.0.0; the final Node 24 rerun also passes.
- Canonical formatting/lint, TypeScript checking, and whitespace checks pass.


markstuart-oai (approved):
Reviewed both changed files and the image event/request contracts plus stream cleanup. The integer/range guard establishes the filename invariant before logging or writing, and sharing the requested partial count keeps the validation and display aligned. No actionable correctness, security, or structural simplification findings.

Independently ran all 19 executable-example tests on Node 24.20.0 and exact minimum 22.0.0; both passed. A separate base/head source-level probe also verified malformed and non-finite indices are rejected before side effects while boundary indices remain valid. Canonical lint and diff whitespace checks passed. Full TypeScript checking was attempted but is limited by missing unrelated Azure/dotenv/Express example dependencies in the reused dependency set; no live API tests were run.

PR #2695, opened 2026-09-07
fix: preserve own Zod schema definition names

HAYDEN-OAI:
The vendored Zod converter currently drops a registered definition named `__proto__` from serialized output, leaving references to that definition unresolved. Define entries as ordinary own data properties so the name survives serialization while the definitions object's prototype stays unchanged.

Validation:
- Four new regressions fail before the fix, covering named and unnamed roots with both `definitions` and `$defs`.
- Checks include serialized references, ordinary definitions, frozen input identity, and unchanged object prototypes.
- All 172 focused Zod/root-schema/parser tests pass on Node 24.20.0 and the exact minimum Node 22.0.0, including existing Zod v3/v4 high-level validation.
- Built CommonJS and ES module converter exports and high-level Zod controls pass on both runtimes (12 checks each).
- Formatting, lint, TypeScript, and build pass.

### DIFF
--- a/src/_vendor/zod-to-json-schema/zodToJsonSchema.ts
+++ b/src/_vendor/zod-to-json-schema/zodToJsonSchema.ts
@@ -137,12 +137,17 @@ const zodToJsonSchema = <Target extends Targets = 'jsonSchema7'>(
       }
 
       for (const [key, schema] of newDefinitions) {
-        definitions[key] =
-          parseDef(
-            zodDef(schema),
-            { ...refs, currentPath: [...refs.basePath, refs.definitionPath, key] },
-            true,
-          ) ?? {};
+        Object.defineProperty(definitions, key, {
+          value:
+            parseDef(
+              zodDef(schema),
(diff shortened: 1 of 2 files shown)


markstuart-oai (approved):
Reviewed both changed files and the surrounding definition collection, reference generation, public Zod helpers, and export contracts at a77b9f916e58. No actionable correctness or maintainability findings. Defining ordinary own data properties preserves serialization and the existing object prototype without introducing a special-case branch or helper; a null-prototype map would change that contract.

Independent validation: 48 source-extracted converter assembly cases passed across definition paths, naming strategies, and ordinary/prototype-sensitive names, with pre-fix loss reproduced. Parser/reference setup was stubbed in that probe. Full Zod tests, lint, typecheck, and build were not independently run: the frozen dependency install was blocked by configured registry connectivity. Reviewed the four new regression cases and existing v3/v4 controls; author/bot test reports were not treated as independently verified.

PR #2685, opened 2026-09-07
fix(assistants): reject identity fields in run-step deltas

HAYDEN-OAI:
Run-step deltas reject a root `id` before raw-event dispatch or snapshot accumulation. Accumulation uses a private representation while valid `runStepDelta` callbacks retain the original delta, including its prototype, non-enumerable metadata, and frozen state. Permitted delta properties are read on their original receiver; listener-added root identity fields are excluded without invoking their getters. Envelope aliases and nested tool-call IDs retain their existing behavior.

Validation:
- Public SSE, serialized-stream, and custom-transport regressions cover callback identity, listener mutations, and receiver-sensitive getters. Both new getter regressions fail on the preceding PR head and pass after the fix.
- All 295 AssistantStream tests pass on Node 24.20.0; all 68 run-step identity tests pass on exact minimum Node 22.0.0.
- Canonical lint, full TypeScript checking, and production CJS/ESM builds and import checks pass.

PR #2698, opened 2026-09-08
fix(assistants): reject root IDs in run-step deltas

HAYDEN-OAI:
Supersedes #2685.

Run-step deltas must not carry a root `id`. Reject that field before raw-event dispatch and recheck the captured delta immediately before accumulation, so an ordinary raw listener cannot introduce it between validation and merging. A listener-added root ID rejects the event instead of being silently stripped.

This keeps run-step identity out of delta accumulation while preserving valid callback objects, nested tool-call IDs, extension properties, and existing alias handling. The production change is 14 additions and 3 deletions in `AssistantStream.ts`.

## Scope

The security guarantee covers decoded SSE/NDJSON protocol data. The pre-merge check also covers ordinary listener additions and replacements. It follows the repository's [documented distinction between protocol data and executable application authority](https://github.com/openai/openai-node/blob/80ce89f355ddb2ea74e48da0284a85d536597446/docs/architecture/security-model.md#L90); arbitrary hostile getters, Proxies, and callback code running with application authority are outside this change's security guarantee.

## Validation

- Added 14 focused cases covering both byte-decoding entrypoints: rejection before raw dispatch, listener-added/replaced root IDs, and preservation of valid callback identity, nested IDs, and extension fields.
- All 278 tests across the seven AssistantStream suites pass on Node 24.20.0 using `./scripts/test tests/lib/AssistantStream.test.ts tests/lib/assistant-stream`.
- All 51 run-step identity tests pass on exact minimum Node 22.0.0.
- Canonical `pnpm lint`, full `pnpm exec tsc`, and `pnpm build` pass with the pinned dependencies.
- Built AssistantStream CJS/ESM imports pass on Node 22.0.0 and 24.20.0.
- One independent static review found no actionable issues within the stated scope.

Validation uses focused rejection tests; the original heap-exhaustion experiment was not rerun.

### DIFF
--- a/src/lib/AssistantStream.ts
+++ b/src/lib/AssistantStream.ts
@@ -525,6 +525,13 @@ export class AssistantStream
       throw new OpenAIError('Received assistant run-step event with an invalid run-step ID');
     }
 
+    if (event.event === 'thread.run.step.delta') {
+      const delta = event.data.delta;
+      if (delta && hasOwn(delta, 'id')) {
+        throw new OpenAIError('Run-step deltas must not contain an id field');
+      }
+    }
+
     if (event.event === 'thread.run.step.created') {
       if (this.#activeRunStepID !== undefined) {
         throw new OpenAIError(
@@ -842,10 +849,14 @@ export class AssistantStream
           throw new Error('Received a RunStepDelta before creation of a snapshot');
         }
 
-        const data = event.data;
+        const delta = event.data.delta;
 
-        if (data.delta) {
-          const accumulated = accumulateAssistantStreamDelta(snapshot, data.delta, true) as Runs.RunStep;
+        if (delta) {
+          // Raw-event listeners can replace or modify the delta after initial validation.
+          if (hasOwn(delta, 'id')) {
+            throw new OpenAIError('Run-step deltas must not contain an id field');
+          }
+          const accumulated = accumulateAssistantStreamDelta(snapshot, delta, true) as Runs.RunStep;
           this.#runStepSnapshots[runStepID] = accumulated;
         }
 
--- a/tests/lib/assistant-stream-run-step-identity-security.test.ts
+++ b/tests/lib/assistant-stream-run-step-identity-security.test.ts
@@ -83,6 +83,104 @@ function toolCallDelta(id: string) {
 }
 
 describe('AssistantStream run-step identity security', () => {
+  describe.each([
+    ['SSE', publicAssistantStream],
+    ['serialized stream', assistantStream],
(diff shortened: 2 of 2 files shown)


jbeckwith-oai (approved):
Reviewed the exact 2-file patch against the requested outcome: reject root run-step delta IDs before raw SSE/serialized dispatch and again after ordinary raw-listener mutation immediately before accumulation, while preserving callback object identity, nested tool-call IDs, extension fields, and existing alias behavior. The new checks are placed at both mutation boundaries and do not broaden into Proxy/getter hardening outside the documented application-authority scope. The full seven-suite AssistantStream target passes locally: 278 tests. Git diff check passes, no public API or generated/runtime surface beyond the 14-line accumulator validation changes, no unresolved threads, and every exact-head lint/build/Node 22/24/26/ecosystem/CodeQL/Castiron check is green. No blocking findings.

markstuart-oai (approved):
Reviewed 99110fc2713aa382f4162c79d9e8d42f96d51d4c against merge-base 80ce89f355ddb2ea74e48da0284a85d536597446. No actionable correctness or structural findings.

Both checks are justified: decoded root IDs must fail before raw dispatch, while ordinary listeners can modify or replace the delta before accumulation. Capturing and checking that delta immediately before merging preserves valid callback identity, nested tool-call IDs, extension fields, and the existing canonical-ID/alias handling. I read the repository security model and evaluated its protocol-data/application-authority distinction. A shared recursive prohibition would incorrectly reject nested IDs; a single check would lose one timing guarantee. No additional ownership framework is warranted. File sizes are 1,025→1,036 lines for AssistantStream and 647→745 for the identity tests; neither newly crosses 1,000 lines.

Fresh validation: 278 tests across seven AssistantStream suites passed on Node 24.20.0; all 51 identity tests passed on exact Node 22.0.0. Independent SSE/NDJSON probes reproduced root-ID accumulation using the merge-base AssistantStream and confirmed protocol/listener rejection, error/end ordering, and valid identity/nested-ID/extension preservation at this head. Build passed on Node 24.20.0; AssistantStream CJS/ESM imports passed on both runtimes. Diff whitespace checks passed.

Validation limits: internal js installation and pinned pnpm installation were blocked by tooling/registry access, so execution used cached dependencies (Vitest 4.1.11, TypeScript 6.0.3). Canonical lint passed with cached Oxlint 1.79.0 rather than pinned 1.80.0. Full type checking could not pass because example dependencies dotenv, @azure/identity, and express were unavailable. These are validation limits, not patch findings. No heap-exhaustion experiment was rerun.

PR #2699, opened 2026-09-08
fix: preserve suppressed workload authorization

HAYDEN-OAI:
Preserve explicit null and empty `Authorization` overrides from client defaults and request headers by resolving only the SDK workload-identity placeholder at dispatch.

Supersedes #2689 with a one-line production fix and 12 focused cases in the existing workload-identity suite. The cases cover client/request overrides, case-insensitive precedence, replacement credentials, undefined values, native Headers, and placeholder resolution. Each case makes two requests to check the result with a populated credential cache.

### Review scope

The acceptance criterion is that dispatch respects the final merged Authorization value, including explicit suppression. Credential-acquisition timing, 401 retry attribution, and transparent ownership through custom hooks/header copies should be evaluated as separate changes.

### Validation

- `./scripts/test tests/lib/workload-identity.test.ts tests/buildHeaders.test.ts`: 46 tests passed on Node 24.20.0 and exact minimum Node 22.0.0.
- `pnpm lint` and `pnpm exec tsc` passed with pnpm 11.5.1.
- `pnpm build` passed, including CJS/ESM core and provider import checks.
- Built CJS and ESM core imports also passed on Node 22.0.0.

Validation used synthetic credentials and in-memory fetch responses.

### DIFF
--- a/src/client.ts
+++ b/src/client.ts
@@ -1550,7 +1550,7 @@ export class OpenAI {
     if (this._workloadIdentityAuth && !this.#x509Fetch && schemes.bearerAuth) {
       const headers = init.headers as Headers;
       const authHeader = headers.get('Authorization');
-      if (!authHeader || authHeader === `Bearer ${WORKLOAD_IDENTITY_API_KEY_PLACEHOLDER}`) {
+      if (authHeader === `Bearer ${WORKLOAD_IDENTITY_API_KEY_PLACEHOLDER}`) {
         const token = await this._workloadIdentityAuth.getToken();
         headers.set('Authorization', `Bearer ${token}`);
       }
--- a/tests/lib/workload-identity.test.ts
+++ b/tests/lib/workload-identity.test.ts
@@ -92,6 +92,60 @@ describe('OpenAI with Workload Identity', () => {
     expect(apiRequestHeaders!.get('Authorization')).toBe('Bearer exchanged-access-token');
   });
 
+  test.each([
+    ['default null', { Authorization: null }, undefined, null],
+    ['default empty', { Authorization: '' }, undefined, ''],
+    ['request null', undefined, { Authorization: null }, null],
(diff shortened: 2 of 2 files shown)


jbeckwith-oai (approved):
Reviewed the exact 2-file patch against the narrow dispatch outcome: preserve the final merged Authorization value, including explicit null and empty client/request overrides, and resolve only the SDK workload-identity placeholder. The equality check is case-insensitive through native Headers lookup, leaves replacement credentials and inherited defaults intact, and preserves suppression on repeated cached-token requests. Credential acquisition timing, 401 provenance, and custom hook ownership are unchanged and remain explicit non-goals. The focused workload-identity and header suites pass locally: 46 tests. Git diff check passes, no API or architecture expansion, no unresolved threads, and every exact-head lint/build/Node 22/24/26/ecosystem/CodeQL/Castiron check is green. No blocking findings.

markstuart-oai (approved):
Reviewed the complete two-file diff and the unchanged header merge, workload credential cache, and dispatch path. No findings. Resolving only the existing workload placeholder preserves the final merged Authorization value, including null/empty suppression, case-insensitive overrides, replacement credentials, and undefined inheritance. This is the right structural simplification: remove the fallback, with no additional state or header machinery. Client size stays 2,362 lines; tests grow 457 to 511.

Fresh validation: 46 focused tests pass on Node 24.20.0 and exact minimum 22.0.0; the same tests against the actual base fail seven suppression cases. An independent built CJS/ESM matrix passes 192 cases per runtime, each on first and cached requests. Canonical build and core/provider import checks pass. These runs used copied local dependencies because the configured registry was unavailable; Vitest matches, but the full dependency tree does not match the lockfile. Exact pinned lint could not be completed, and whole-project tsc is blocked by missing example dependencies. Author-reported lint/type results remain unverified locally. Credential-acquisition timing, 401 attribution, and arbitrary hook/header-copy ownership remain separate follow-ups.

PR #2708, opened 2026-09-09
feat(api): Add API key expiration controls

apcha-oai:
## Summary

Add optional expiration controls and metadata for project and service-account API keys. Clarify the public handling contract for SIP headers in incoming-call webhooks.

## Changes

- Add `expires_in_seconds` to service-account creation parameters for setting the initial API key lifetime, including documented policy and service-account-only constraints.
- Expose optional `expires_at` timestamps on project API keys and newly created service-account API keys.
- Document that incoming-call webhook SIP headers exclude authorization headers, preserve retained header names, values, repetitions, and order, and must be treated as untrusted metadata.


tests/api-resources/admin/organization/projects/service-accounts/service-accounts.test.ts
```diff
@@ -26,6 +26,7 @@ describe('resource serviceAccounts', () => {
     const response = await client.admin.organization.projects.serviceAccounts.create('project_id', {
       name: 'name',
       create_service_account_only: true,
+      expires_in_seconds: 1,
```
markstuart-oai, line 29: [P3] Use a valid expiration fixture

This combines `expires_in_seconds: 1` with `create_service_account_only: true`, although the new parameter contract forbids a non-null lifetime when no initial key is created. The SDK forwards both fields unchanged, so this smoke fixture describes a request the API rejects. Please set `create_service_account_only` to `false` for this expiration case, and keep the keyless case separate with the lifetime omitted or null. Prefer fixing the generated fixture input so regeneration preserves the correction.

markstuart-oai (approved):
The expiration fields match the schema and use the existing request/response flow without adding runtime special cases. I checked the full generated diff, nullable/omitted serialization with a synthetic fetch, and focused TypeScript compilation. One low-priority fixture correction below: the optional-parameters smoke test combines mutually exclusive options. No substantive implementation or structural issues found.

jbeckwith-oai (dismissed):
Two in-scope blockers remain on this exact head. The new service-account test sends expires_in_seconds together with create_service_account_only: true even though the documented contract forbids that combination, so it does not exercise a successful expiring-key request (see the existing inline thread). Also, the shared moderation-model documentation now labels a link as “available models” but points it back to the moderation guide in api_reference/openapi.transformed.yml; that link should target the models documentation. Please correct the source and regenerate.

PR #2713, opened 2026-09-09
test: reduce CI subprocess and validation overhead

apcha-oai:
## Summary

The Node test jobs on current main take about 5m38s, compared with 1m32s in the August 12 reference run. The regular test step grew from 32s to 224–228s; packed-package validation grew from 38s to 77–79s.

Compile the example, ecosystem CLI, and SDK sources once per Vitest invocation with the existing `ts-node`/SWC compiler and repository configuration. Each subprocess test then runs that JavaScript in a fresh Node process. Previously, hundreds of child processes independently loaded a TypeScript compiler and transpiled the same source graph.

Also enable Node's compilation cache within each job, overlap independent packed-package downloads and no-emit type checks with bounded concurrency, and hoist an invariant JSON parse out of a property-test prefix loop. Keep the existing CI jobs, sequential test/package steps, four Vitest workers, and full runtime matrix. No SDK runtime or dependency changes.

The final measured run is green. This reduces the Node 22/24 jobs from about 5m38s to 3m47s/3m21s; the original 1–2 minute target is not reached. No additional CI jobs or suite/runtime exclusions are introduced.

## CI wall time

| Measurement | Baseline: Aug 12 | Current main: Sep 9 | After this fix |
| --- | ---: | ---: | ---: |
| Node 22 job | 1m32s | 5m37s | 3m47s |
| Node 22 — run tests | 32s | 3m44s | 1m58s |
| Node 22 — packed package, including build | 38s | 1m19s | 1m14s |
| Node 24 job | 1m05s | 5m38s | 3m21s |
| Node 24 — run tests | 22s | 3m48s | 1m50s |
| Node 24 — packed package, including build | 23s | 1m17s | 1m05s |
| Node 26 experimental job | 1m06s | 3m43s | 2m32s |
| Node 26 — run tests | 31s | 3m10s | 1m56s |

[Baseline run](https://github.com/openai/openai-node/actions/runs/31556390296) · [Current-main run](https://github.com/openai/openai-node/actions/runs/34372510439) · [After-fix run](https://github.com/openai/openai-node/actions/runs/34392311247) (`a5abbb71`). Durations are job/step execution time, excluding queue time. The historical baseline has substantially less coverage; the fix preserves all current tests. The table reports one successful run per revision, not a benchmark median; runner and download variance still affect these numbers.

## Coverage and behavior preserved

| Optimization | What changes | What remains covered |
| --- | --- | --- |
| Shared subprocess compilation | Emit a fresh source snapshot once per Vitest invocation using the existing SWC compiler and TypeScript configuration. Resolve SDK aliases to that snapshot and preserve private package imports through the existing build metadata projection. Delete the snapshot after the suite. | Every existing example/CLI case and assertion still executes in a fresh child process against the current source, with real exits, signals, filesystem operations, HTTP/SSE and WebSocket exchanges. Documentation snippets still execute from the docs. No results or module state are shared. Normal source type checking and real CJS/ESM packed builds remain separate checks. |
| Native Node compilation cache | Reuse compiled code only within the job's temporary directory. Clean-environment subprocess fixtures forward only the cache path. | Fresh processes and module state per fixture; no cached results or skipped execution. Nothing is restored from another run or uploaded. |
| Packed-package concurrency | Download the same six exact Undici releases with at most three concurrent downloads; run the same seven no-emit type checks with at most two compiler children. | Real npm packing/installation, isolated consumers, absent/present optional peers, every Undici compatibility case, CJS/ESM/browser checks, source maps, both TypeScript 4.9 and 6, and the exact Node 22.0.0 smoke check. `stopOnError: false` waits for all children before temporary-directory cleanup; any failure still exits nonzero. |
| Property-test invariant | Parse each generated complete JSON value once, outside the prefix loop. | The same generator, number of cases, prefix traversal, partial parser calls, and assertions. |

No tests, assertions, suites, runtime matrix entries, package variants, or dependency versions are removed. No SDK runtime or API behavior changes.

## Validation

- Pinned Node 24.19.0 / pnpm 11.5.1 frozen install passed.
- Formatting, lint, TypeScript checking, and SDK build passed locally.
- All 212 handwritten suites / 8,167 tests passed locally. The first attempt encountered a macOS temporary-path alias mismatch in unchanged Node-review tests; using a canonical `/private/tmp` path resolved all 41 tests in that suite.
- Full packed-package validation passed locally, including the concurrent downloads and compiler checks.
- All 82 generated suites / 556 tests and 12 snapshots passed locally (the existing 1 skipped suite / 2 skipped tests are unchanged).
- Public CI passed at `a5abbb71`: Node 22/24/26 tests, packed validation on both required Node lines, exact Node 22.0.0 smoke, build, lint, ecosystem checks, and performance benchmarks.
- Independent review completed cleanly: no actionable P0–P2 findings. One separate Codex task used `gpt-5.6-sol` with high reasoning, read-only in the same checkout with no worktree or full e2e sweep.
- Marked ready for review to run [full PR CI](https://github.com/openai/openai-node/actions/runs/34392991940). Both [CodeQL merge-protection analyses](https://github.com/openai/openai-node/actions/runs/34392991898) passed with no findings.


markstuart-oai (approved):
Reviewed a5abbb7143ec2c0fe4dd7ea79f825ac6fc9218bc with two independent passes for correctness and structural simplification. No actionable findings. The shared compilation snapshot removes repeated compiler startup while keeping fresh child processes, SDK alias/private-import routing, and the existing assertions. The bounded packed-package work drains all children before cleanup, including failures, without dropping version or compiler cases.

Validation here: diff checks and the Node version-policy check passed on Node 24.19.0. Dependency installation was blocked by the execution environment, so I did not independently run the fixture suites, build or packed-package tests.

PR #2717, opened 2026-09-10
Close converted stream sources after read or serialization errors

sylvesterkaczmarek:
## Summary

`Stream.toReadableStream()` errors the readable when `next()` or JSON serialization fails, but an errored readable does not invoke its cancellation hook. A suspended generator remains open, and a failed tee branch keeps accepting sibling-driven results.

Reuse the existing iterator cancellation path on failure. Preserve the original error even when cleanup throws, rejects, or stalls; keep healthy tee siblings running. Share one cleanup promise with explicit cancellation so an overlapping pending pull cannot close the source twice.

## Regression coverage

Nine public stream tests cover generator finalization, custom iterator failures, failing/stalled cleanup, sibling isolation, closing the remaining tee branch, cancellation overlap, and normal exhaustion. The seven failure cases were reproduced against unmodified `fe2d6a3` before applying the fix.

## Validation

- Focused streaming suites: 84 tests passed on Node 24.19.0.
- `pnpm exec tsc --noEmit`, `pnpm lint`, and `pnpm build` passed, including CJS/ESM build smoke checks.
- Full handwritten suite: 8,172 passed; four `node-version-review.test.ts` failures also reproduce on a clean, unchanged base checkout.
- Custom-code budget passes at 3,017/4,000 lines. No policy, dependency, snapshot, or public-signature changes.

### DIFF
--- a/src/core/streaming.ts
+++ b/src/core/streaming.ts
@@ -352,10 +352,13 @@ export class Stream<Item> implements AsyncIterable<Item> {
    * which can be turned back into a Stream with `Stream.fromReadableStream()`.
    * Canceling a response-backed readable aborts its request. Canceling a tee
    * branch discards its buffered events and leaves sibling consumers running.
+   * Read or serialization failures also release the iterator without replacing the original error.
    */
   toReadableStream(): ReadableStream {
     const { controller } = this;
     let iter: AsyncIterator<Item>;
+    let cancellation: Promise<void> | undefined;
+    const cancel = () => (cancellation ??= this.#cancelIterator(iter, controller));
 
     return makeReadableStream({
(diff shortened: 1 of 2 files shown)


HAYDEN-OAI (approved):
Reviewed with Codex at `0e143f79f4eedf10f2fc081b0ab8cc8784042906`. Traced iterator cleanup after read and serialization failures, overlapping cancellation, original-error preservation, and tee sibling isolation. No actionable correctness or lifecycle findings. LGTM.

sylvesterkaczmarek:
Thanks for the review and merge. I appreciate the careful check of stream cleanup and cancellation/error preservation.

PR #2718, opened 2026-09-10
feat(api) Add Live API

apcha-oai:
- Add Live API endpoints.
- Add WebSocket and browser WebRTC helpers.
- Add `TranscriptGrouper` for streaming transcripts.


api.md
mccrearydamion69-crypto, line 1172: Bypassing some statically

mccrearydamion69-crypto: Maybe 🤔 you 🏆 could

src/client.ts
mccrearydamion69-crypto, line 2: Rich 🤑 games

mccrearydamion69-crypto, line 2: Richest people

PR #2726, opened 2026-09-12
fix(streaming): emit terminal SSE events missing a trailing blank line

### LINKED ISSUE #2725: SSE stream drops the final event (finish_reason / response.completed lost) when the server omits the trailing blank line
## Summary
When an SSE stream ends right after the final `data:` line (without the trailing blank line `\n\n`), the final event is **silently dropped**. For the Responses API this event is usually `response.completed`, so downstream strict clients see the stream end with no terminal event and raise "OpenAI Responses stream ended before a terminal response event", then retry.

Reproduced on openai-node 6.40.0 and still present on 7.15.0 (Node v22).

## Repro
A local HTTP server returns a `text/event-stream` body for a `chat.completions` streaming call, in three variants:

- `A` final event ends with a trailing blank line `\n\n` (conformant)
- `B` final event ends with a single `\n`
- `C` final event ends with no newline at all

```js
for await (const p of stream) { if (p.choices[0].finish_reason) ... }
```

Result: `A` reports `finish_reason`; `B` and `C` never do (1 chunk instead of 2), with **no error raised**.

## Root cause
`core/streaming.mjs` -> `streamSSE()`: `SSEDecoder.decode(line)` only emits a parsed event when it sees an empty line; `iterSSEChunks()` yields the leftover bytes at EOF, but the decoder is never flushed, so a trailing unterminated event is discarded.

## Proposed fix (verified on 6.40.0 and 7.15.0)
After the `lineDecoder.flush()` loop in `streamSSE()`, flush any pending event:

```js
const sse = sseDecoder.decode("");
if (sse) yield sse;
```

Verified against conformant, multi-event, and comment-trailing streams: no regression; the trailing event is now delivered.

## Impact
Any gateway/proxy that closes the connection immediately after the final `data:` line (a common server race, especially under load) triggers this. The symptom is intermittent and load-dependent, so it reads as network flakiness.

GISWLH:
## Summary
- `SSEDecoder` previously emitted only on a blank line. `iterSSEChunks` already yields leftover bytes at EOF and `LineDecoder` is flushed, but the SSE decoder was never flushed, so a final `data:` / `event:` record without `\n\n` was discarded.
- Flush the pending SSE event at EOF so `finish_reason` / `response.completed` are delivered when a gateway closes the stream immediately after the last field.
- A stream that already ends with a blank line is unchanged: `flush()` returns `null` when no event is in progress, so there is no double-emit. Comment-only trailers still produce no event. Abort still returns before flush.

Fixes #2725.

## Test plan
- [x] `./scripts/test tests/lib/streaming-core.test.ts tests/streaming.test.ts` (86 passed)
- [x] `./scripts/test tests/lib/streaming-named-error-events.test.ts tests/lib/streaming-sentinel-integrity.test.ts tests/lib/streaming-malformed-sse-privacy.test.ts tests/core/streaming-backpressure.test.ts tests/core/streaming-large-events.test.ts tests/core/streaming-security.test.ts` (98 passed)
- [x] Coverage for trailing `\n\n`, single `\n` / `\r` / `\r\n`, no final newline, multi-event, comment-only trailer, and abort mid-stream

### DIFF
--- a/src/core/streaming.ts
+++ b/src/core/streaming.ts
@@ -569,6 +569,15 @@ export async function* _iterSSEMessages(
         yield sse;
       }
     }
+    // Servers sometimes omit the trailing blank line that normally
+    // terminates the last event. Flush any in-progress event exactly once.
+    if (signal.aborted) {
+      return;
+    }
+    const pending = sseDecoder.flush();
+    if (pending) {
+      yield pending;
+    }
   } catch (error) {
     failed = true;
     if (!signal.aborted || (!isAbortError(error) && error !== signal.reason)) {
@@ -698,6 +707,15 @@ class SSEDecoder {
     }
 
     return null;
+  }
+
+  /**
+   * Emits a pending event at EOF when the stream omitted the trailing blank
+   * line. Returns `null` when no event is in progress so a record that already
+   * ended with a blank line is not delivered twice.
+   */
+  flush(): ServerSentEvent | null {
+    return this.decode('');
   }
 }
 
--- a/tests/lib/streaming-core.test.ts
+++ b/tests/lib/streaming-core.test.ts
@@ -235,6 +235,35 @@ describe('Stream.fromSSEResponse', () => {
 
     expect(controller.signal.aborted).toBe(true);
   });
+
+  test.each([
+    ['a trailing blank line', '\n\n'],
+    ['a single trailing newline', '\n'],
+    ['no trailing newline', ''],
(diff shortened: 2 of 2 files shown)


HAYDEN-OAI (approved):
Reviewed with Codex at `1d4018d9619b2b9f615b42d55b64572dec96ed6f`. EOF flushing matches the behavior requested in #2725. Traced line decoding, abort handling, source disposal, completed-event deduplication, comment trailers, and `[DONE]` handling. No actionable correctness or compatibility findings. LGTM.

PR #2740, opened 2026-09-14
fix(live): trim transcript acknowledgment suffixes in linear time

HAYDEN-OAI:

## Changes being requested

Long punctuation or whitespace runs followed by a letter make acknowledgment normalization spend quadratic time searching for a trailing match. Replace that suffix regex with a backward scan, preserving the existing normalization and acknowledgment behavior.

This patch covers the cost of normalizing one string. Repeated normalization of accumulated transcript fragments remains a separate follow-up.

## Additional context & links

Related: [SDK-898](https://linear.app/openai/issue/SDK-898).

Validation:
- The regression exceeded a bounded five-second child-process deadline before the fix; the existing 43 controls passed.
- All 47 transcript grouper tests pass on Node 24 and the supported floor, Node 22.0.0, including large valid inputs and Unicode/custom acknowledgment controls.
- Lint, TypeScript, package build, built CJS/ESM import checks, and the custom-code budget check pass.


jbeckwith-oai (approved):
Exact-head review complete. The backward suffix scan preserves acknowledgment normalization while making per-string trimming linear, and the focused regression covers both adversarial and Unicode/custom-acknowledgment cases.

markstuart-oai (approved):
Reviewed `b9df30736da9a8a03fc69b4b7b6feb1aee9cf3f6` with independent review. No actionable findings. The reverse suffix scan removes the demonstrated quadratic search while preserving whole-string lowercasing, the exact trim character set, and acknowledgment behavior. This is a direct fix at the existing normalization boundary; no additional abstraction or payload limit is needed.

The public transcript tests passed 47/47 on Node 24.19.0. Both long-input reproductions time out with the base-equivalent normalizer and complete with the expected output on this head; 347,680 additional base/head normalization comparisons matched. Lint passed. On Node 22.23.2, the suite passed 47/47 with NODE_NO_WARNINGS=1; the default environment's injected proxy warning trips two stderr assertions. Exact-head hosted checks are green or skipped. Exact Node 22.0.0 and the full generated suite were not run locally.

PR #2742, opened 2026-09-14
fix(websocket): discard unread events when an iterator returns

HAYDEN-OAI:

## Changes being requested

Explicitly returning a WebSocket stream iterator currently leaves unread events retained and available to subsequent reads. Clear that iterator's queue and reset its cursor during return, alongside the existing listener cleanup.

Apply the same two-line change to stable/beta Responses and the three Live WebSocket variants. Normal socket closure still drains queued events, and other iterators continue independently. Active slow-consumer buffering remains a separate follow-up.

## Additional context & links

Related: [SDK-896](https://linear.app/openai/issue/SDK-896).

Validation:
- The 15 new cancellation cases fail before the fix; the 15 existing buffering controls pass.
- All 30 buffering tests pass on Node 24 and Node 22.0.0; 96 related Responses/Live tests also pass.
- Lint, TypeScript, package build, built CJS/ESM import checks, and the custom-code budget check pass.


jbeckwith-oai (approved):
Exact-head review complete. Iterator return clears only that iterator’s retained queue/cursor across all five variants, while preserving other iterators and normal socket-close draining; the focused cases cover open and closed sockets.

markstuart-oai (approved):
Reviewed `c25337c0d42656f939ec7e5428cca0289fb339b2` with independent review. No actionable findings. Clearing the iterator-owned queue after marking it done makes explicit return terminal without changing normal close drainage, another iterator's state, or the underlying connection. The same two-statement lifecycle correction belongs in each existing owner; a new shared abstraction is not warranted.

The canonical buffering suite passed 30/30 on Node 24.19.0 and 22.23.2, and 66 related WebSocket tests passed on Node 24. The new tests run against exact base versions of the five production modules reproduce 15 failures while all 15 existing controls pass. Diff checks passed; exact-head hosted checks are green or skipped. Local validation did not include exact Node 22.0.0, full generated tests, or packed-install/build checks.

PR #2743, opened 2026-09-14
fix(tests): handle ecosystem proxy socket failures

HAYDEN-OAI:

## Changes being requested

Socket errors in the ecosystem runner's CONNECT proxy can terminate the runner as unhandled errors. Attach error handlers to both tunnel sockets, destroy both on failure, and release a pending upstream connection when the client closes.

The implementation change is confined to the existing CONNECT handler. Tests run the actual CLI against synthetic loopback connections, including a successful forwarding control.

## Additional context & links

Related: [SDK-895](https://linear.app/openai/issue/SDK-895).

Validation on the repository's Node 24 toolchain:
- Before the fix, all three failure/cleanup regressions fail and the normal forwarding control passes.
- After the fix, the four proxy cases and existing CLI tests pass (75 total).
- Lint and TypeScript pass.

This changes test infrastructure and does not require an SDK runtime release. An extra Node 22.0.0 test run stops in the existing CLI dependency loader because its CommonJS output requires ESM-only yargs; the repository tooling uses Node 24.


jbeckwith-oai (approved):
Exact-head review complete. The CONNECT proxy now contains both tunnel-side socket errors and tears down a pending upstream connection when the client closes, while the CLI-level regressions retain a successful relay control.

markstuart-oai (approved):
Reviewed `99008202e99731f3bb22bf7cef27a5548eb8ad12` with an independent cross-check. No actionable findings. The paired error handlers and downstream-close cleanup prevent a failed or abandoned CONNECT tunnel from leaving its peer alive. The connect-time guard closes the readiness race before acknowledgement and piping. Normal relay and half-close behavior remain intact, with cleanup kept in the existing owner.

Canonical proxy and CLI tests passed 75/75 on Node 24.19.0. The regression invokes the actual compiled CLI and real loopback CONNECT path, covering both socket errors, close-before-connect and successful relay. Frozen install, lint, TypeScript and diff checks passed; exact-head hosted checks are green or skipped. No live API traffic, exact Node 22.0.0 run, or local full generated-suite run was performed.

PR #2751, opened 2026-09-16
fix(zod): omit impossible optional branch in strict schemas

### LINKED ISSUE #2462: Zod v3: a `.nullable().optional()` schema shared by two fields is encoded two different ways, and one of them uses `not`
## Summary

When one Zod **v3** schema object ending in `.nullable().optional()` is reached
from more than one field, the strict Structured Outputs helpers encode it twice,
differently, in a single document:

- the first occurrence is inlined and is clean;
- the second becomes a `$ref` into `definitions`, and that `definitions` entry
  keeps an `anyOf: [{ "not": {} }, …]` wrapper.

`not` is not in the JSON Schema subset strict Structured Outputs accepts. The
SDK's own `toStrictJsonSchema()` rejects it — `src/lib/transform.ts` lists `not`
as unsupported and raises *"uses unsupported keyword `not`"* — but only the Zod
v4 and Standard Schema helpers call that function. `zodV3ToJsonSchema()`
(`src/helpers/zod.ts`) returns its schema without it, so on the v3 path the
wrapper reaches the request body unchecked.

Zod v4 is unaffected: `zodV4ToJsonSchema()` goes through `z4.toJSONSchema()` and
never enters the vendored converter.

## Reproduction

No API key, no network.

```bash
npm init -y
npm i openai@7.5.0 zod@3
node repro.mjs
```

```js
import { zodResponseFormat } from 'openai/helpers/zod';
import { z } from 'zod/v3';

// `.optional()` on its own is rejected by these helpers with
// "uses `.optional()` without `.nullable()`", so `.nullable().optional()` is the
// accepted shape.
const OptionalName = z.string().nullable().optional();

// One schema object, two fields.
console.log(
  JSON.stringify(
    zodResponseFormat(z.object({ a: OptionalName, b: OptionalName }), 'p').json_schema.schema,
    null,
    2,
  ),
);
```

Actual:

```json
{
  "type": "object",
  "properties": {
    "a": { "type": "string", "nullable": true },
    "b": { "$ref": "#/definitions/p_properties_a" }
  },
  "required": ["a", "b"],
  "additionalProperties": false,
  "definitions": {
    "p_properties_a": {
      "anyOf": [
        { "not": {} },
        { "type": "string", "nullable": true }
      ]
    },
    "p": { "…": "same as the root" }
  },
  "$schema": "http://json-schema.org/draft-07/schema#"
}
```

Expected — `properties.a` and `definitions.p_properties_a` are the same Zod
node, so they should be the same JSON Schema:

```json
"definitions": {
  "p_properties_a": { "type": "string", "nullable": true }
}
```

## Scope

Measured on `openai@7.5.0` (2026-08-17), which already contains the Zod v3
fixes from #2346, #2355, #2357 and #2358.

Affected, in all three of `zodResponseFormat`, `zodTextFormat` and
`zodFunction`:

| shape | affected |
|---|---|
| `.nullable().optional()` string reached from two fields | yes |
| the same, object / array / enum inner type | yes |
| the same, reached from three fields | yes |
| the same, nested one level down | yes |
| supplied through `schemaDefinitions` (no sharing needed) | yes |
| `.nullable()` without `.optional()`, shared | no |
| `.nullable().optional()` used once | no |
| any other shared schema (plain, object, enum, array, union) | no |

Also reaches `zodResponsesFunction` and the non-strict Realtime tool path.

## Cause

`zodToJsonSchema.ts` materializes a definition because some property `$ref`s it,
then parses it with `currentPath` pointing at the definition but no
`propertyPath`:

```ts
definitions[key] =
  parseDef(
    zodDef(schema),
    { ...refs, currentPath: [...refs.basePath, refs.definitionPath, key] },
    true,
  ) ?? {};
```

`parseOptionalDef` (`parsers/optional.ts`) branches on exactly that field:

```ts
if (
  refs.propertyPath &&
  refs.currentPath.slice(0, refs.propertyPath.length).toString() === refs.propertyPath.toString()
) {
  return parseDef(def.innerType._def, { ...refs, currentPath: refs.currentPath }, forceResolution);
}
const innerSchema = parseDef(def.innerType._def, {
  ...refs,
  currentPath: [...refs.currentPath, 'anyOf', '1'],
}, forceResolution);
return innerSchema ? { anyOf: [{ not: {} }, innerSchema] } : {};
```

Inline, `parseObjectDef` sets `propertyPath` and the first branch is taken. In
the definitions loop it is `undefined`, so the same Zod node takes the second
branch and gains the `anyOf`/`not` wrapper.

`parseOptionalDef` is the only reader of `propertyPath` in the vendored
converter — `grep -rn propertyPath src/_vendor/` returns `Refs.ts` (declaration),
`parsers/object.ts` (writer) and `parsers/optional.ts` (reader) — so this is the
only parser the omission can affect.

## Environment

- `openai@7.5.0`, `zod@3.25.76`, Node v23.1.0
- Also reproduced against `main` @ `39c06d1`

## Not verified

No request was sent to the API. The claim that `not` is unacceptable rests on
the SDK's own `toStrictJsonSchema()` treating it as unsupported, not on an
observed 400. If the API in fact tolerates `anyOf: [{not: {}}, X]`, the
inconsistency still stands but the severity is lower than stated here.

HAYDEN-OAI:

## Changes being requested

Shared Zod v3 `.nullable().optional()` fields can generate extracted definitions containing `anyOf: [{ not: {} }, innerSchema]`. The `not` branch accepts no JSON values and is unsupported by strict Structured Outputs.

For strict conversion without an override callback, generate `anyOf: [innerSchema]` and parse the child at `anyOf/0`. Keeping the wrapper preserves existing optional-root diagnostics and metadata behavior, while generating references against the final layout avoids post-processing or reference repair. Non-strict and override-enabled conversion retain their existing behavior.

The production change is six added lines and seven removed lines in the optional parser. Focused tests cover all four strict helpers, both nullable/optional orders, supplied definitions and descriptions, escaped definition names, shared nested and recursive schemas, and non-strict/override controls.

## Additional context & links

Fixes #2462. Supersedes #2463 with a smaller fix at schema generation. The acceptance scope is removing the generated impossible optional branch while preserving valid references and existing diagnostics; general JSON Schema normalization is outside this change.

Validation:

- New regression suite: 17 passed; 16 fail against the unchanged parser and the non-strict/override control passes.
- Schema/helper suite: 1,154 passed.
- Full handwritten unit suite: 8,677 passed, four failures in `tests/node-version-review.test.ts`. The same four failures reproduce with the parser change removed.
- `./scripts/lint`, `pnpm exec tsc --noEmit`, and `pnpm build` passed using Node 24.19.0.
- Built CommonJS and ESM helper regressions passed on Node 22.0.0 and Node 24.19.0.
- Independent review found no correctness or security findings. No live API request was used for validation.


markstuart-oai (approved):
Reviewed `3a14cd427d6f16545c8d2eb48d3a7098a92f0b3b`. This is a good simplification at the owning boundary: omit the impossible optional branch during generation and register the child at its final `anyOf/0` path. Keeping the wrapper preserves annotations and optional-root diagnostics, while the explicit override guard retains existing callback output and path behavior.

No actionable findings after independent reference and compatibility reviews. Shared, nested, recursive, and supplied-definition cases preserve reference meaning; non-strict conversion, custom overrides, Zod v4, Realtime, and response parsing retain their existing contracts.

Validation: 1,127 schema/helper tests passed, including all 17 new regressions; lint, typecheck, and build passed. Built CJS and ESM helpers passed conversion, reference, parsing, and root-diagnostic probes on Node 22.0.0 and 24.20.0. CI for this commit passed, including the Node 22/24/26 matrix. All validation was local or CI; no live API calls were made.

jbeckwith-oai (approved):
Reviewed exact head `3a14cd427d6f16545c8d2eb48d3a7098a92f0b3b`; no blocking findings. The change removes only the impossible strict-mode optional branch and records the child at its final `anyOf/0` path, while preserving the wrapper, non-strict generation, and caller override behavior. Shared, nested, recursive, and supplied-definition references remain valid. I ran the 17 focused regressions and the full Zod helper suite locally (811 tests); all passed. Current CI, CodeQL, Castiron, and OkTest (236/236) are green.

PR #2756, opened 2026-09-17
test(x509): stabilize retry budget timing

huanglianggit:

### Problem
The X.509 retry-budget regression test used a 35 ms request delay with a 55 ms timeout. On Windows, scheduler and startup overhead can consume the 20 ms gap before the first simulated API response, causing an unintended one-attempt timeout.

### Root cause
The regression's control flow depended on a narrow wall-clock margin instead of leaving enough time for the first request and retry bookkeeping to complete.

### Fix
Increase the simulated request delay, total timeout, and retry-local preparation delay proportionally. The test still requires two local preparations, two API attempts, and a timeout during the second API attempt, so it continues to verify that retry-local preparation time is excluded from the X.509 network budget.

### Tests
- `./scripts/test tests/lib/x509-workload-request-boundaries.test.ts` (48 passed), run twice
- `pnpm lint`
- `pnpm exec tsc --noEmit`
- `pnpm build`

### Compatibility / risk
Test-only change. No generated SDK source, public API, runtime behavior, or protocol behavior changes.


HAYDEN-OAI (approved):
Reviewed at 59037a7. No substantive findings.

The timing changes preserve the intended timeout behavior: one 250 ms network attempt fits within the 400 ms budget, two exceed it, and the 500 ms retry-local preparation remains excluded by the existing budget accounting. No SDK runtime or API changes.

Static review only; tests were not run.

PR #2760, opened 2026-09-17
feat(api): add managed Responses WebSocket sessions

markstuart-oai:
## Summary

Adds managed Responses WebSocket sessions with custom request headers, independent lanes and final-response collection.

## Changes

- Preserve client headers and per-connection overrides during WebSocket handshakes
- Route independent response lanes and collect final responses
- Reconnect explicitly with application restoration and no automatic request replay

## Validation

Validated with local WebSocket regression suites, lint/type/package checks, and current-head hosted CI. A fresh authenticated smoke test passed custom headers and two successive Responses on one WebSocket.


src/lib/responses/responses-websocket-lane.ts
```diff
@@ -0,0 +1,284 @@
...
+    send: (event: ResponsesClientEvent) => void,
+    release: (bytes: number) => void,
+    maxEvents: number,
+    maxBytes: number,
+  ) {
+    Object.defineProperty(this, 'streamID', {
+      value: streamID,
+      enumerable: true,
+      configurable: true,
+      writable: true,
```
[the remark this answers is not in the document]
markstuart-oai: Verified in `d57b3a7c910b`. The lane now captures its registered routing ID privately and uses that value for outgoing commands. The public streamID property keeps its existing descriptor and mutability. Regression cases reassign or redefine that property on named and default lanes and verify commands and responses still use the registered lane, while the other lane remains independent. Verified that both implementation and regression cases are present in this public head.

jbeckwith-oai (approved):
Reviewed the updated public head d57b3a7c910b thoroughly, focusing on handshake header/auth precedence and redirect containment, lane routing/isolation and buffer accounting, cancellation/reconnect behavior, and terminal/final-response handling. The prior mutable routing-ID finding is fixed, all review threads are resolved, required CI is green, and the focused local suite passes (173 tests).

src/lib/responses/responses-websocket-lane.ts
```diff
@@ -0,0 +1,322 @@
...
+        }
+        if (terminal) {
+          this.#validateTerminalResponse(event);
+          // Socket deltas need not include the setup events required by SSE.
+          // The terminal response is authoritative; finalized items fill an omitted output.
+          const completedEntries = [...completedOutput];
+          completedEntries.sort(([left], [right]) => left - right);
+          const snapshot = cloneResponse(context, {
+            ...event.response,
+            output: event.response.output ?? completedEntries.map(([, item]) => item),
```
jbeckwith-oai: The new own-property checks protect `response.output_item.done`, but the terminal path still reads `event.response.output` and `output_text` through the prototype chain. If `Object.prototype.output` or `output_text` is polluted while a terminal response omits those fields, these expressions replace the lane's completed items/canonical text with inherited data; I reproduced that with a plain `JSON.parse` response object. Please select both fields with own-property checks (and use the same decision for the line 224 recount condition), then add a terminal-response regression alongside the new inherited output-item cases.

jbeckwith-oai (changes requested):
One in-scope inherited-field gap remains in final-response accumulation.

src/lib/responses/responses-websocket-lane.ts
markstuart-oai: Verified in `8edc1444fd04`. Terminal normalization now selects output and output_text only when they are own fields, and the canonical-text recount uses that same selection. Omitted or null fields still use accumulated output/canonical text; explicit empty output and explicit text remain authoritative. Stable and beta socket regressions cover inherited values larger than the configured byte budget, all three terminal types, unchanged raw events, mutable returned snapshots and an unaffected neighboring lane. The new regressions fail on the prior implementation and pass with this correction.

dpiet-oai (approved):

jbeckwith-oai (approved):
Re-reviewed exact head `6137f02c469c22643d5114b3a0164f551cff304f`, focusing on the terminal-response own-field correction and the custom-event snapshot/accounting hardening. Terminal `output`/`output_text` selection now ignores inherited values while preserving explicit empty/null semantics and accumulated canonical output; invalid terminal normalization fails only the affected lane. Custom events are cloned once, nested inherited serialization hooks are masked on the clone, and the retained JSON is exactly what byte accounting measures without mutating caller input. Both prior threads are resolved, and the complete exact-head public CI matrix is green.

jbeckwith-oai (approved):
Re-reviewed exact head `8ee3e86cbd082d30b4f5b26acd7008f730dd1ec8`. The final delta is correctly scoped: completed output items must be non-null, non-array wire objects before entering the typed response, malformed items fail only their lane, and future object-shaped item types remain supported without a discriminator allowlist. The existing output-index/own-property checks, canonical terminal finalization, lane isolation, raw-event observability, and byte accounting remain intact.

All public review threads are resolved. The focused stable/beta WebSocket error suite passed (58 tests), and exact-head build, lint, Node 22/24/26, ecosystem, CodeQL, and Castiron checks are green. The paired internal repository was not inspected.

PR #2763, opened 2026-09-17
fix(api): validate WebSocket results and preserve header defaults

markstuart-oai:
## Summary

Preserves Responses WebSocket header defaults and rejects malformed managed response output.

## Changes

- Ignore undefined connection headers and remove explicitly null headers
- Validate completed and terminal output while preserving future event variants and independent response lanes


jbeckwith-oai (approved):
Reviewed exact head `db1515c221e0bfc197758a0d0fa6645a1f797100`; no blocking issues found.

The output-item checks are scoped to the fields the finalizer consumes: terminal and completed-item paths require an own string discriminator, message content is an array of discriminated objects, and output-text parts require an own string `text`. Invalid data fails only that lane, while unknown future object-shaped variants remain forward-compatible.

The stable and beta Node handshakes now apply per-connection headers case-insensitively over the normalized client/auth defaults: `undefined` leaves defaults intact, `null` removes them, and explicit strings (including empty strings) override them. Redirect disabling and proxy/dispatcher behavior remain unchanged.

Both focused suites passed locally (94 tests). Public review threads are empty, and exact-head build, lint, Node 22/24/26, ecosystem, CodeQL, and Castiron checks are green. The paired internal repository was not inspected.

dpiet-oai (approved):

PR #2768, opened 2026-09-18
feat(api): add environment-variable vault credentials

jbeckwith-oai:
## Summary

Adds environment-variable credentials for Beta Agents vaults.

## Changes

- Create environment-variable credentials with a name, secret value, and destination permissions
- Rotate stored secret values without changing the credential name or destination permissions
- Read credential metadata and use distinct request and response networking types


dpiet-oai (approved):

markstuart-oai (approved):
The generated vault credential types reflect the new environment-variable variant without exposing `secret_value` in returned metadata. Create requires the write-only secret and networking policy, while rotate requires a replacement secret and preserves configuration; the discriminated unions match the transformed OpenAPI shape. I found no consequential issue in this SDK diff. Exact-head CI, including test matrix, Castiron consistency and CodeQL, passes; live proxy/authorization behavior is outside this generated-client review.

